PR #8, opened 2019-11-11
chore: moving kernel and unity-client files into their own dirs

moliva:
# What? 
...

# Why? 
...


eordano:
Overriden by https://github.com/decentraland/explorer/pull/37

PR #38, opened 2019-11-12
fix: planeshape UVs (authored by: @pravus)

eordano:
WHY?
The 'uv' propety of the PlaneShape is not working, because when we re-generate the geometry of the shape, we are not re-assigning the new mesh to the entities mesh filter.

WHAT?
Added meshFilters collection in MeshesInfo and tied its update to the onShapeUpdated entity event.

The corresponding test already existed (PlaneShapeUpdateWithUVs()) and will be fixed in https://app.zenhub.com/workspaces/world-5d8ba4a04da3400001fc292f/issues/decentraland/unity-client/1030


eordano:
@pravusjif @BrianAmadori can we move forward with this?

javiermaria (approved):

BrianAmadori (approved):

PR #40, opened 2019-11-12
chore: add debug context menu for waiting components (authored by: @BrianAmadori)

eordano:
## WHY
Sometimes components can get stuck. We have no way of knowing which components have that problem making the debugging cost of these problems higher than it should.

## WHAT
Added a new context menu item to `ParcelScene` component to give detailed information in the console about which components are being waited, and which entities are attached to them. 

image

image


pbosio (approved):

eordano:
Looks good! @pravusjif what do you think?

javiermaria (approved):

pravusjif (approved):

PR #53, opened 2019-11-12
(closed without merging)

BrianAmadori:
Was: `feat: asset bundle support`

Batch tool parameters quick guide:

`-alwaysBuild`
Build asset bundles regardless of they presence in the content server

`-keepBundles`
Keep downloaded assets in Assets/_Downloaded folder after finished building,
useful for taking advantage of local caching when building many scenes in a row

`-sceneCid <scene-id>`
Download scene with cid <scene-id>

`-parcelsXYWH <x> <y> <width> <height>`
Build all scenes in range of parcels in the rectangle defined.
The rectangle pivot is bottom-left. i.e. 10,10,5,5 will stretch from 10,10 to 15,15.


BrianAmadori:
closing because I split this PR.

PR #57, opened 2019-11-12
fix: triangles count metrics

### LINKED ISSUE #18: Triangle counting isn't being updated correctly
In the SDK, I see this issue, here's how to reproduce it:

- Run the Hummingbird scene:  https://github.com/decentraland-scenes/Hummingbirds
- Open the stats panel with "P" and check the triangle number
- Click on the tree to spawn a hummingbird, see that the counter rises
- Click on it again to spawn another hummingbird, the counter stays the same

In this scene we're reusing a single GLTF shape component on various entities, and it's only being counted once, when in reality it should be counted multiple times.

In the Unity Builder, if I add the same asset twice, I don't see the counter being updated for the second model I add.
You can try it out here
https://builder-unity.decentraland.org/

Another possibly related issue in the Builder is the triangle count often gets stuck, and I can keep adding or removing entities without the triangle value changing. I have to leave and enter the project for the value to be fixed.

pbosio:
Count amount of triangles even if mesh is already in memory.
Closes #18

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneMetricsController.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneMetricsController.cs
@@ -178,6 +178,7 @@ namespace DCL
                 if (mf != null && mf.sharedMesh != null)
                 {
                     model.bodies--;
+                    model.triangles -= mf.sharedMesh.triangles.Length / 3;
                     isDirty = true;
 
                     if (uniqueMeshes.Contains(mf.sharedMesh))
@@ -227,6 +228,7 @@ namespace DCL
                 if (mf != null && mf.sharedMesh != null)
                 {
                     model.bodies++;
+                    model.triangles += mf.sharedMesh.triangles.Length / 3;
                     isDirty = true;
 
                     if (!uniqueMeshes.Contains(mf.sharedMesh))
@@ -254,7 +256,6 @@ namespace DCL
         public void AddMesh(Mesh mesh)
         {
             model.meshes++;
-            model.triangles += mesh.triangles.Length / 3;
             isDirty = true;
 
             if (VERBOSE) { Debug.Log("Mesh name = " + mesh.name + " ... tri count = " + (mesh.triangles.Length / 3)); }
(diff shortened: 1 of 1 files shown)


BrianAmadori (changes requested):
Remember to update the tests

javiermaria (approved):

javiermaria (approved):

nearnshaw:
@BrianAmadori . @pbosio    
from what I understand, tests won't be able work until we fix all of the ones that Pravus was working on, right?
This fix is a blocker for shipping the Unity-based Builder, which is supposed to go out this week
Any chance we can leave the tests for later?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneMetricsController.cs
```diff
@@ -178,6 +178,7 @@ void RemoveGameObject(GameObject go)
                 if (mf != null && mf.sharedMesh != null)
                 {
                     model.bodies--;
+                    model.triangles -= mf.sharedMesh.triangles.Length / 3;
```
pravusjif: Why are we dividing by 3 the triangles instead of the vertices?

javiermaria: Following that, the number of vertices in a model is not `triangles / 3` unless the model consists of completely detached triangles - there will always be shared vertices between triangles.

pravusjif: Also, we should actually just use `mf.sharedMesh.triangles.Length` right? no division at all

pravusjif (approved):

javiermaria (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneMetricsController.cs
BrianAmadori: Yep, I think I initially though this was referring to vertices for some reason and added the `/ 3` but Javi is right. Its makes no sense.

BrianAmadori (approved):

PR #59, opened 2019-11-12
chore: sending unreachable position eagerly when teleporting

moliva:
# What? 
Sends unreachable position message to the current position topic when teleporting to avoid appearing in the same spot while the world is loading.


eordano (approved):
Although I hate the codename `MORDOR`, it looks good.

BrianAmadori (approved):

PR #60, opened 2019-11-13
feat: export wearables as Asset Bundles

eordano:
@BrianAmadori will need your help! I don't have a better strategy than this.


eordano:
@BrianAmadori do you think you can take it from here? what do you think of the approach?

BrianAmadori (approved):

pravusjif (approved):

eordano:
Ping @BrianAmadori

PR #72, opened 2019-11-14
fix: analytics config override always to avoid sending prod data

### LINKED ISSUE #71: Ensure that we never use a production setting for Segment
This is high priority:

- Diego has realized that sometimes using feature branches previews (`/branch/....`) we are combining that with the `?ENV=org` setting, which causes segment to report inaccurate data about our numbers

- We should make sure that whatever mechanism we use, we are not polluting the DB.

A possible approach could be to modify `queueTrackingEvent` to check that the `window.location` value is actually `explorer.decentraland.org` before sending prod data.

eordano:
Fixes #71

### DIFF
--- a/kernel/packages/shared/index.ts
+++ b/kernel/packages/shared/index.ts
@@ -38,21 +38,26 @@ import { initializeUrlPositionObserver } from './world/positionThings'
 import { setWorldContext } from './protocol/actions'
 import { profileToRendererFormat } from './passports/transformations/profileToRendererFormat'
 
+enum AnalyticsAccount {
+  PRD = '1plAT9a2wOOgbPCrTaU8rgGUMzgUTJtU',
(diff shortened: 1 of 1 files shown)


kernel/packages/shared/index.ts
```diff
@@ -43,7 +43,10 @@ function initializeAnalytics() {
   const TLD = getTLD()
   switch (TLD) {
     case 'org':
-      return initialize('1plAT9a2wOOgbPCrTaU8rgGUMzgUTJtU')
+      if (window.location.host === 'explorer.decentraland.org') {
```
moliva, line 51: This might need to be revisited in the future as I understand we will need to use other hosts as well depending on the region.

moliva (approved):

moliva:
Review #74 along with this to improve code legibility.

eordano:
Beautiful thanks!

PR #104, opened 2019-11-15
fix: ALL_WEARABLES was not working if fetching inventory errored

eordano:


kernel/packages/shared/passports/sagas.ts
```diff
@@ -233,7 +230,11 @@ export function* submitPassportToRenderer(action: PassportSuccessAction): any {
     while (!(yield select(baseCatalogsLoaded))) {
       yield take(CATALOG_LOADED)
     }
-    yield call(sendLoadProfile, action.payload.profile)
+    const profile = { ...action.payload.profile }
+    if (ALL_WEARABLES) {
+      profile.inventory = (yield select(getExclusiveCatalog)).map((_: Wearable) => _.id)
```
moliva, line 235: Beware that this will be reset if the profile does not contain an `ethAddress`.

moliva (approved):

PR #109, opened 2019-11-15
fix: minimap metadata format

moliva:
# What? 
Sending minimap scene information in chunks to avoid deserialization issues.


eordano (approved):

moliva:
Waiting for @AjimenezDCL to confirm if we should go with this approach or another serializable-friendly format.

AjimenezDCL:
@moliva I'm thinking about closing this PR so we tackle down the real issue (the size of the message).

AjimenezDCL (approved):

PR #120, opened 2019-11-21
feat: asset bundle loading support

BrianAmadori:
# What? 
This PR adds support for the loading of asset bundles from the new content server bucket.

The entry point for the code is at `LoadWrapper_GLTF`. So you can start reviewing from there.

The following systems were added in order to implement the full support:

#### `AssetPromiseKeeper_AB`
This APK is in charge of loading AssetBundles and their dependencies. The dependencies are determined by the loading of `.depmap` files (described below). 

Dependencies will be loaded recursively and sorted by asset type in order for them to be resolved properly by Unity. Textures are prioritary, then materials, then everything else.

#### `AssetPromiseKeeper_AB_GameObject`
This APK is in charge of loading an AssetBundle hierarchy. Then it proceeds by instancing the GameObject assets of this AB, and pooling it. 

After the GameObject is instantiated, an additional optimization pass will ensure the materials with same CRC are shared between different instances. Keeping the draw call batching optimizations that we used with GLTF loading.

#### `AssetLibrary_Poolable` & `AssetLibrary_RefCounted`
The `AssetLibrary` classes were generalized in these two types, as I found these will be the most useful ones. 

`AssetLibrary_Poolable` is used for Poolable `GameObject` bound AssetPromiseKeeper systems, such as the `GLTF` and `AB_GameObject` ones. 

`AssetLibrary_RefCounted` is intended to be used in all systems were `Instantiate` is not needed and the same asset can be reused between many rendereables, for instance this is the case for Textures, Materials, etc. 

Its currently being used by the `AssetPromiseKeeper_AB` system.

#### Misc changes
- Kernel now sends an additional field with the `AssetBundle` content server in its scene messages.
- Generalized test suites for `Poolable` and `RefCounted` AssetPromiseKeeper systems, so less code is duplicated (there's still room for improvement on this)
- Added `IAssetPromiseSettings` and `AssetPromiseSettings_Rendering` classes to share settings logic between `AB_GameObject` and `GLTF` promises.
- Added an optional field in `WSSController` to use a custom content server if we wish. This is useful for retrieving assets from a custom local http server.
- Change folder structure for all `AssetPromiseKeeper` test suites, now each asset type will have its own test folder.

#### Known room for improvements
- Several tests are still duplicated
- I'm not very happy with the heavy generic usage in the `AssetPromiseKeeper` classes, it makes it hard to write generalized tests. The reason behind the generics is to keep the interfaces friendly to the end user.
- There's too much coupling between `AssetPromiseKeeper` and `AssetPromise` classes.

#### Known issues
When in editor, if many asset bundle requests fail there's the chance of the editor getting crazy, with behaviours including: 

- Getting stuck in 100% CPU usage. 
- Asset loading requests getting stuck forever. The requests just won't finish or lag considerably. 

If after many tests with failing requests you are experimenting this, there's no fix, you'd have to restart the editor.

#### About `.depmap` files:
An asset will be stored as `content-server-url.org/content-hash` and the corresponding `.depmap` will be stored in `content-server-url.org/content-hash.depmap`. The .depmap has a simple json format containing the hashes of the dependencies, they are mainly textures, but they potentially can be animations, etc.

# Why? 
AssetBundles greatly improve the CPU bound cost of loading 3D models. However there's an increase of bandwidth-bound cost that can be mitigated by using gzip and chunk compression. 

As our bottleneck was heavy on CPU side, we are using uncompressed asset bundles by now.


kernel/packages/entryPoints/editor.ts
```diff
@@ -84,6 +84,7 @@ async function getSceneData(scene: IScene & { baseUrl: string }): Promise<ILand>
   return {
     name: 'Editor scene',
     baseUrl: scene.baseUrl,
+    baseUrlBundles: '',
```
moliva, line 87: Is this a stub or will we not support asset bundles for the builder? What do you think @pbosio ?

kernel/packages/decentraland-loader/lifecycle/controllers/download.ts
```diff
@@ -159,7 +161,8 @@ export class SceneDataDownloadManager {
 
     const data: ILand = {
       sceneId: sceneId,
-      baseUrl,
+      baseUrl: baseUrl,
```
moliva:
```suggestion
      baseUrl,
```

kernel/packages/decentraland-loader/lifecycle/controllers/download.ts
```diff
@@ -159,7 +161,8 @@ export class SceneDataDownloadManager {
...
+      baseUrlBundles: baseUrlBundles,
```
moliva:
```suggestion
      baseUrlBundles,
```

kernel/packages/decentraland-loader/lifecycle/manager.ts
```diff
@@ -56,6 +56,7 @@ export async function initParcelSceneWorker() {
 
   server.notify('Lifecycle.initialize', {
     contentServer: DEBUG ? resolveUrl(document.location.origin, '/local-ipfs') : getServerConfigurations().content,
+    contentServerBundles: DEBUG ? null : getServerConfigurations().contentAsBundle,
```
moliva: In some cases we are returning `null` and empty strings in others, should we simplify into one value to represent the absence of asset bundle server?

kernel/packages/entryPoints/editor.ts
BrianAmadori: Just to be safe, and as we talked today we don't gain much by it, I won't enable AB support in builder for now.

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_GLTF.cs
```diff
@@ -19,14 +30,61 @@ public class LoadWrapper_GLTF : LoadWrapper
         [ContextMenu("Debug Load Count")]
         public void DebugLoadCount()
         {
-            Debug.Log($"promise state = {gltfPromise.state} ... waiting promises = {AssetPromiseKeeper_GLTF.i.waitingPromisesCount}");
+            float loadTime = Mathf.Min(loadFinishTime, Time.realtimeSinceStartup) - loadStartTime;
+
+            if (gltfPromise != null)
+                Debug.Log($"promise state = {gltfPromise.state} ({loadTime} load time)... waiting promises = {AssetPromiseKeeper_GLTF.i.waitingPromisesCount}");
```
eordano, line 34: hide under verbose?
```suggestion
                if (VERBOSE)
                    Debug.Log($"promise state = {gltfPromise.state} ({loadTime} load time)... waiting promises = {AssetPromiseKeeper_GLTF.i.waitingPromisesCount}");
```

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_GLTF.cs
```diff
@@ -19,14 +30,61 @@ public class LoadWrapper_GLTF : LoadWrapper
...
+
+            if (abPromise != null)
+                Debug.Log($"promise state = {abPromise.state} ({loadTime} load time)... waiting promises = {AssetPromiseKeeper_AssetBundle.i.waitingPromisesCount}");
```
eordano:
```suggestion
                if (VERBOSE)
                    Debug.Log($"promise state = {abPromise.state} ({loadTime} load time)... waiting promises = {AssetPromiseKeeper_AssetBundle.i.waitingPromisesCount}");
```

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_GLTF.cs
BrianAmadori: There's no need for the check. Note that this is under an `[ContextMenu("Debug Load Count")]` attribute. This means it will only be executed through a contextual menu in the shape unity inspector. We'd always want this menu item to write to the console.

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_GLTF.cs
```diff
@@ -19,14 +30,61 @@ public class LoadWrapper_GLTF : LoadWrapper
...
+        {
+            if (abPromise != null)
+            {
+                AssetPromiseKeeper_AssetBundle.i.Forget(abPromise);
+
+                if (VERBOSE)
+                    Debug.Log("Forgetting not null promise...");
+            }
+
+            string bundlesBaseUrl = useCustomContentServerUrl ? customContentServerUrl : entity.scene.sceneData.baseUrlBundles;
```
AjimenezDCL, line 67: Can we find a way to feed this `baseUrlBundles` in any other way to avoid adding coupling between the component and the entity/scene.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundle/AssetLibrary_AssetBundle.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            Pool pool = PoolManager.i.AddPool(asset.id, asset.container);
+
+            pool.OnCleanup -= OnPoolRemoved;
+            pool.OnCleanup += OnPoolRemoved;
+        }
+
+        public override Asset_AssetBundle Get(object id)
+        {
+            if (Contains(id))
```
AjimenezDCL: Can we invert the `if` to reduce nesting?
```
if (!Contains(id)) return null;
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundle/AssetPromise_AssetBundle.cs
```diff
@@ -0,0 +1,137 @@
+﻿using DCL.Helpers;
+using System;
+using System.Collections;
+using UnityEngine;
+
+namespace DCL
+{
+    public class AssetPromise_AssetBundle : AssetPromise<Asset_AssetBundle>
```
AjimenezDCL: This `AssetPromise` shares a fairly ammount of code with `AssetPromise_GLTF`, maybe not in this PR but we could try to put the common stuff in a parent class.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundleModel/AssetPromise_AssetBundleModel.cs
```diff
@@ -0,0 +1,80 @@
...
+                container.transform.parent = asset.container.transform;
+                container.transform.ResetLocalTRS();
+
+                OnSuccess?.Invoke();
+            }
+        }
+
+        protected override void ApplySettings_LoadStart()
+        {
+            Transform assetTransform = asset.container.transform;
```
eordano: This is killing my client with a null pointer exception

eordano (dismissed):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundle/Asset_AssetBundle.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace DCL
+{
+    public class Asset_AssetBundle : Asset
+    {
+        public GameObject container;
+        public AssetBundle ownerAssetBundle;
+        public string assetBundleAssetName;
+
+        public Asset_AssetBundle()
```
AjimenezDCL: Can the `ownerAssetBundle` and `assetBundleAssetName` of an `Asset_AssetBundle` change?

If not, what about using dependency injection to provide them?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/DCLCharacterController.cs
```diff
@@ -218,7 +218,10 @@ public void SetPosition(string positionVector)
     public void SetEnabled(bool enabled)
     {
         camera.enabled = enabled;
+        //NOTE(Brian): Added this clause to avoid the endless logs related with no having AudioListeners enabled in editor.
```
AjimenezDCL, line 221: Praise the lord!

AjimenezDCL (dismissed):

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_GLTF.cs
BrianAmadori: I already delved into this a while ago. The entity/scene coupling is necessary for the `OnCleanupEvent` suscription and the provider. Also it happens on component level, so we have to make sure other components aren't needing it too. 

We can try and figure how to decouple it out, but in any case this design issue It's out of this PR scope.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundle/AssetPromise_AssetBundle.cs
BrianAmadori: I'm kind of rewriting this. When the rewriting is done the only code in common will be the settings stuff. I'm coming back to you then.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundle/Asset_AssetBundle.cs
BrianAmadori: Kind of hard to do because we are creating it like this:

https://github.com/decentraland/explorer/blob/be47ed71a68f73784bc724caaca24d525dd60a50/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetPromise.cs#L94

This forces us to fill the dependencies in `ApplySettings_LoadStart` or `OnLoad`

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundle/AssetBundleLoadHelper.cs
```diff
@@ -0,0 +1,327 @@
...
+using UnityGLTF.Cache;
+using WaitUntil = DCL.WaitUntil;
+
+public static class MaterialCachingHelper
+{
+    public static void UseCachedMaterials(GameObject obj)
+    {
+        foreach (var rend in obj.GetComponentsInChildren<Renderer>(true))
+        {
+            var matList = new List<Material>(1);
```
pravusjif: Would it be any good to have this cached at `MaterialCachingHelper` level? and just clearing it before using it in this method.

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_GLTF.cs
AjimenezDCL: Yes, I agree!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundle/Asset_AssetBundle.cs
AjimenezDCL: Oh I see... Maybe we could implement a Factory to solve it but in any case, it might be overkill and is not a big deal. We can investigate further and tackle it down in another PR.

pravusjif:
I'm a little concerned about all the Linq usage, but I guess if it's not used on every frame it shouldn't impact on performance.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundle/AssetPromise_AssetBundle.cs
```diff
@@ -0,0 +1,238 @@
...
+                    //                            }
+                    //                        }
+                    //#endif
+                    //if (asset.EndsWith("glb") || asset.EndsWith("gltf"))
+                    //{
+                    //    if (!bundleToMainAssets.ContainsKey(assetBundle))
+                    //        bundleToMainAssets.Add(assetBundle, new List<string>(1));
+
+                    //    bundleToMainAssets[assetBundle].Add(asset);
+                    //}
```
pravusjif: Let's remove this commented code

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundle/AssetPromise_AssetBundle.cs
```diff
@@ -0,0 +1,238 @@
...
+            //        yield return null;
+            //    }
+
+            //    if (assetBundleRequest.isHttpError || assetBundleRequest.isNetworkError)
+            //    {
+            //        failedRequests.Add(url);
+            //        downloadingBundle.Remove(url);
+            //        Debug.LogWarning("AssetBundle request fail! " + url);
+            //        yield break;
+            //    }
```
pravusjif: removeeeee

pravusjif (dismissed):
Looking good mostly, but i understand several parts are being refactored right now so I'll hold the approve until seeing the new version (and testing the renderer with that)

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
```diff
@@ -0,0 +1,194 @@
...
+
+        static float maxLoadBudgetTime = 0.032f;
+        static float currentLoadBudgetTime = 0;
+        public static bool limitTimeBudget = false;
+
+        Coroutine loadCoroutine;
+        static HashSet<string> failedRequestUrls = new HashSet<string>();
+
+
+        static Dictionary<string, int> loadOrderByExtension = new Dictionary<string, int>()
```
AjimenezDCL, line 23: Could I get some insight on why we need to prioiritize assets by extension?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
```diff
@@ -0,0 +1,194 @@
...
+                    yield break;
+                }
+
+                string[] assets = assetBundle.GetAllAssetNames();
+                List<string> assetsToLoad = new List<string>();
+
+                assetsToLoad = assets.OrderBy(
+                    (x) =>
+                    {
+                        string ext = x.Substring(x.Length - 3);
```
AjimenezDCL: Can we use `Path.GetExtension()` here?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/Asset_AB.cs
```diff
@@ -0,0 +1,68 @@
...
+        public AssetBundle ownerAssetBundle;
+        public string assetBundleAssetName;
+
+        public Dictionary<string, Object> assetsByName = new Dictionary<string, Object>();
+        public Dictionary<string, List<Object>> assetsByExtension = new Dictionary<string, List<Object>>();
+
+        public List<T> GetAssetsByExtensions<T>(params string[] extensions)
+            where T : Object
+        {
+            var goList = new List<T>(2);
```
AjimenezDCL: Why is the list initialized with a capacity of 2?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Common/AssetPromiseSettings_Rendering.cs
```diff
@@ -0,0 +1,63 @@
+using DCL.Helpers;
+using UnityEngine;
+
+namespace DCL
+{
+    public class AssetPromiseSettings_Rendering : IAssetPromiseSettings<Transform>
```
AjimenezDCL, line 6: As you mentioned in the description, this is not optimal. If I had to create a new `IAssetPromiseSetting` I would have no idea what to put in that Generic, it's not clear why is it needed.

Let's let it go through in this PR and redesign it in the future?

AjimenezDCL (commented):
I'm approving but I am concerned about how maintainable is this system (AssetManager) at all, although it's true that using it from outside is quite straightforward and simple. Unfortunately I have no better proposal.

Great job, gonna test it!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
BrianAmadori: Yes, we have to load textures before materials, and load materials before meshes. If we don't do that the dependencies will not be resolved in build. In editor the behaviour is different. If you dont load them in the proper order, they are resolved randomly. Sometimes they do, sometimes they wont. Sometimes they wont and when you alt+tab they'd appear out of thin air.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
BrianAmadori: Nice catch. This worked with incomplete path or filenames? I think I decided to use Substring because of that edge case.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Common/AssetPromiseSettings_Rendering.cs
BrianAmadori: The generic is to know which object we'd need to apply the setting. In this case is a Transform, in a Texture case i'd be the texture, etc.

BrianAmadori: In any case yes, I hope to keep polishing this design whenever its possible.

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_GLTF.cs
```diff
@@ -73,13 +72,21 @@ void LoadAssetBundle(string targetUrl, Action<LoadWrapper> OnSuccess, Action<Loa
                 return;
             }
 
-            abPromise = new AssetPromise_AssetBundleModel(bundlesBaseUrl, targetUrl);
+            entity.scene.contentProvider.TryGetContentsUrl_Raw(targetUrl, out string hash);
```
pravusjif, line 75: What happens if the returned `hash` is invalid somehow? for example if it's an empty string

BrianAmadori: if its invalid the method just fills the out variable with `targetUrl`. The request would fail afterwards, calling the `OnFail` callback.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/MaterialCachingHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+    public static class MaterialCachingHelper
+    {
+        public static void UseCachedMaterials(GameObject obj)
+        {
+            if (obj == null)
+                return;
+
+            foreach (var rend in obj.GetComponentsInChildren<Renderer>(true))
+            {
+                var matList = new List<Material>(1);
```
pravusjif: Would it be better if we have the matList at MaterialCachingHelper level and just clear it here before using it?

BrianAmadori: done

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/Asset_AB.cs
BrianAmadori: I figured there is more probable we return 2 elements than 1. Not a strong one, just reverted it.

eordano (approved):
Please merge conflicts!

pravusjif (approved):

AjimenezDCL (approved):

PR #121, opened 2019-11-21
feat: gltf to asset bundle batch tool

BrianAmadori:
Batch tool parameters quick guide:

`-alwaysBuild`
Build asset bundles regardless of they presence in the content server

`-keepBundles`
Keep downloaded assets in Assets/_Downloaded folder after finished building,
useful for taking advantage of local caching when building many scenes in a row

`-sceneCid <scene-id>`
Download scene with cid

`-parcelsXYWH <x> <y> <width> <height>`
Build all scenes in range of parcels in the rectangle defined.
The rectangle pivot is bottom-left. i.e. 10,10,5,5 will stretch from 10,10 to 15,15.


unity-client/Assets/Batch/AssetBundleMenuItems.cs
```diff
@@ -0,0 +1,70 @@
+using System.Runtime.CompilerServices;
+using UnityEditor;
```
AjimenezDCL, line 4: I see this file is not inside a folder `Editor` is neither having a preprocessing direcitve.  Is a `Batch` folder excluded from builds as well?

unity-client/Assets/Batch/Tests/ParseOptionShould.cs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    [Test]
+    public void FailWhenTooManyArgumentsAreGiven()
+    {
+        Assert.IsFalse(DCL.AssetBundleBuilderUtils.ParseOptionExplicit(args, "testOption", 5, out string[] test5));
+        Assert.IsFalse(DCL.AssetBundleBuilderUtils.ParseOptionExplicit(args, null, 5, out string[] test6));
+        Assert.IsTrue(test5 == null);
+        Assert.IsTrue(test6 == null);
+    }
```
AjimenezDCL: I think this is a good opportunity to use the `[TestCase]` attribute and merge them all in a single method.

If you are afraid of losing some semantic in the way due to the automatically generated name of the test you can do something like this:
```
        [UnityTest]
        [TestCase("a", "b", TestName = "Testing_A_B")]
        [TestCase("a", "b", "c")]
        public void ParametrizedTest(params string[] params) {
       }
```
It supports the keyword `params` and you can set the property `TestName` to override the name of the test in the TestRunner. That code would create two test one called `ParametrizedTest("a", "b", "c")` and another one called `Testing_A_B`.

unity-client/Assets/Batch/Tests/UtilsTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        abDependency.Unload(true);
+
+        if (hasMap)
+        {
+            Debug.Log("Dependency has been generated correctly!");
+        }
+        else
+        {
+            Debug.Log("Dependency has NOT been generated correctly!");
+        }
```
AjimenezDCL: What is the point of this output? Shouldn't the result of the test in the `.xml` autogenerated file for Tests be enough. Are we gonna run this in an environment where is hard to retrieve files and easier to check the standard output?

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/ContentServerUtils.cs
```diff
@@ -0,0 +1,100 @@
+using UnityEngine;
+
+namespace DCL
+{
+    public static class ContentServerUtils
```
AjimenezDCL: Are we gonna ever use this for something else than AssetBundles? If no, should we move it to a path within the AssetBundle related scripts?

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/ContentServerUtils.cs
```diff
@@ -0,0 +1,100 @@
...
+                }
+
+                public Content content;
+            }
+
+            public Data[] data;
+        }
+
+        [System.Serializable]
+        public class MappingPair
```
AjimenezDCL: I think we should move `MappingPair` to the standard `Utils` collection script. We are using it across the whole project in quite different contexts.

I do agree with taking it out of `ContentProvider` though.

AjimenezDCL (changes requested):

unity-client/Assets/Batch/AssetBundleBuilder.cs
```diff
@@ -0,0 +1,639 @@
...
+        private void RetrieveAndInjectTexture(Dictionary<string, MappingPair> hashToGltfPair, string gltfHash, MappingPair mappingPair)
+        {
+            string fileExt = Path.GetExtension(mappingPair.file);
+            string realOutputPath = finalDownloadedPath + mappingPair.hash + "/" + mappingPair.hash + fileExt;
+            Texture2D t2d = null;
+
+            if (File.Exists(realOutputPath))
+            {
+                string outputPath = finalDownloadedAssetDbPath + mappingPair.hash + "/" + mappingPair.hash + fileExt;
+                t2d = AssetDatabase.LoadAssetAtPath<Texture2D>(outputPath);
```
pravusjif, line 509: Can we assume this texture we are loading is being disposed automatically? or it doesn't matter much because this runs on the content server or similar to transform the assets?

pravusjif (approved):

unity-client/Assets/Batch/AssetBundleBuilder.cs
```diff
@@ -0,0 +1,639 @@
...
+
+            EditorApplication.CallbackFunction updateLoop = null;
+
+            updateLoop = () =>
+            {
+                try
+                {
+                    //NOTE(Brian): We have to check this because the ImportAsset for GLTFs is not synchronous, and must execute some delayed calls
+                    //             after the import asset finished. Therefore, we have to make sure those calls finished before continuing.
+                    if (!GLTFImporter.finishedImporting || Time.realtimeSinceStartup - timer > 60)
```
javiermaria: Is 60 seconds arbitrary or how did you get to that number? It is reasonable, but I am curious

unity-client/Assets/Batch/AssetBundleMenuItems.cs
BrianAmadori: Check the `.asmdef` file. The only enabled platform is `Editor`.

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/ContentServerUtils.cs
BrianAmadori: The `MappingPair` struct is directly related to the `ContentServer`. That's the reason for being here and not in `Utils`.

unity-client/Assets/Batch/AssetBundleBuilder.cs
BrianAmadori: I believe the texture we are loading here is the `AssetDatabase` asset. This means we shouldn't have to `Destroy` it as we'd deleting the texture in that case, like with prefabs.

unity-client/Assets/Batch/AssetBundleBuilder.cs
BrianAmadori: It's just a bailout number in the weird case the `GLTFImporter` got dangling callbacks and `finishedImporting` never reaches true. In all cases I tested it this was never reached, but I'd want to prevent an infinite loop in the case a bug breaks this logic.

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/ContentServerUtils.cs
BrianAmadori: The DCL content server is a very important service that's more or less used by all the client stuff. So, I figured i'd be more useful to have it in the project as a uncoupled system, rather than couple it with the converter logic. This is making sense?

unity-client/Assets/Batch/AssetBundleMenuItems.cs
AjimenezDCL: Oh, that's really convenient.

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/ContentServerUtils.cs
AjimenezDCL: I was refering to classes like `ScenesAPIData` or `MappingsAPIData`. If I recall correctly, when I wrote that comment the only reference I could found to them was within the AssetBundles ecosystem. Therefore my next comment talking about moving the `MappingPair` to something more common.

Anyway, makes completely sense what you are saying. If those classes are thightly related I understand them being together.

javiermaria (approved):

eordano:
Please fix conflicts & merge!

unity-client/Assets/Batch/Tests/UtilsTests.cs
BrianAmadori: I added this way before because I couldn't make this method run as a test. Removed it.

AjimenezDCL (approved):

unity-client/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -1021,11 +1034,11 @@ void ProcessCurves(Transform root, GameObject[] nodes, AnimationClip clip, GLTFA
                         // var targetCount = primitives[0].Targets.Count;
```
oOZakMcKrackeNOo: We should remove commented code.

oOZakMcKrackeNOo (approved):

pravusjif (approved):

PR #137, opened 2019-11-25
fix: entities marked as out of boundaries at the edge of the scene

pbosio:
Objects at the very edge of the scene or objects with the same size as the scene were marked as out of bounds.

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
@@ -330,19 +330,23 @@ namespace DCL.Controllers
             // We need to check using a threshold from the target point, in order to cover correctly the parcel "border/edge" positions
 
             // We check the east-threshold position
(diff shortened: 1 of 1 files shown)


pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -330,19 +330,23 @@ public virtual bool IsInsideSceneBoundaries(Vector3 worldPosition, float height
             // We need to check using a threshold from the target point, in order to cover correctly the parcel "border/edge" positions
 
             // We check the east-threshold position
-            targetCoordinate.Set(Mathf.FloorToInt((worldPosition.x + ParcelSettings.PARCEL_BOUNDARIES_THRESHOLD) / ParcelSettings.PARCEL_SIZE), noThresholdZCoordinate);
+            targetCoordinate.Set(Mathf.FloorToInt((worldPosition.x + ParcelSettings.PARCEL_BOUNDARIES_THRESHOLD) / ParcelSettings.PARCEL_SIZE),
```
BrianAmadori: This is getting harder and harder to read with each change. I think we should move the `Mathf.FloorToInt((worldPosition.blah + threshold)` statements to local variables.

BrianAmadori (changes requested):

BrianAmadori (approved):

PR #169, opened 2019-11-29
feat: input controller system

AjimenezDCL:
New Input Controller which allows us to decouple CharacterController, CameraController and how the inputs are received.

It also contains the very raw implementation of a multicamera system. Once the system get's the approval we "just" have to tweak cinemachine virtual cameras.


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputActionResolverFactory.cs
```diff
@@ -0,0 +1,58 @@
...
+using UnityEngine;
+
+//In the future this information could be retrieved from a ScriptableObject.
+//This will allow us to:
+// - Have a SO per platform and the factory can load the needed one using directives.
+// - Change dynamically the keymap by updating that SO and recreating the instances.
+
+internal static class InputActionResolverFactory
+{
+    internal static InputActionResolver_OneTime[] CreateInstance(InputAction_OneTime oneTimeAction)
```
BrianAmadori: Why are we returning an array here? In which cases we would need more than one resolver?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputAction_InTime.cs
```diff
@@ -0,0 +1,138 @@
...
+
+#if UNITY_EDITOR
+
+    private void OnEnable()
+    {
+        Application.quitting -= CleanUp;
+        Application.quitting += CleanUp;
+    }
+
+    private void CleanUp()
```
BrianAmadori: Maybe we can move this code to a base class?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputAction_Measurable.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        if (Input.GetKeyUp(keyCode))
+        {
+            measurableAction.RaiseOnValueChanged(minValue);
+        }
+    }
+}
+
+internal class MouseButtonInputActionResolver_Measurable : InputActionResolver_Measurable
```
BrianAmadori: All of the input specific stuff shouldn't be in the InputController layer? I imagined the base measurable was going to listen to an input BaseVariable instead of this.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputActionResolverFactory.cs
AjimenezDCL: We can have multiple resolvers mapped to the same Action. Like MainAction with LeftMouseButton and E in the keyboard. It's also a pre-step to allow multiple devices at the same time, like controllers + keyboard (if im not mistaken you can use controllers in WebGL)

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputAction_InTime.cs
AjimenezDCL: I created a base class and deleted it. Besides the cleanup (only for Editor) I cannot really get common functionality since I cannot make Resolvers to inheritance a base class

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputAction_Measurable.cs
AjimenezDCL: Could you sketch how would it look?

AjimenezDCL: I think now I understand what you meant. Basically remove the Resolvers and move that logic to the InputController right? After giving it some thoughts in the weekend I think it could lead to a really elegant solution. Gonna change it on Monday and let's see how it looks

Although I see redundant what you said about an Action listening to another BaseVariable. 
What I'm gonna do is:
- InputController update Actions (each Action is a SO).
- Actions raise events when updated: 
    - OneTimeAction: OnTriggered
    - InTimeAction: OnStarted, OnFinished
    - MeasurableAction: OnValueChanged
- Systems in the project will suscribe to those Actions, for example:
   - DCLCharacterController will listen to an OneTimeAction.OnTriggered called Jump 
   - DCLCharacterController will listen to an InTimeAction.OnStarted and OnFinished called Sprint

I dont see where those extra BaseVariable you mentioned would go.

BrianAmadori: I think when I was referring to `BaseVariable` i was actually referring to Actions. I picture the `InputController` updating `Actions` and handling the keybinding to the different devices. Then, the other systems suscribe to those actions in order listen to the input.

AjimenezDCL: So more or less what is currently implemented? I dont really like the `InputProcessor` class but without that `InputController` would contain a lot of boilerplate

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraController.cs
```diff
@@ -1,91 +1,156 @@
...
 
-    internal BaseVariable<CameraState>.Change cameraStateChangedDelegate;
-    internal Dictionary<CameraState, CameraSetup> cameraSetups;
-    internal CameraSetup currentSetup;
+    [Header("Virtual Cameras")]
+    [SerializeField] internal CameraStateToVirtualCamera[] cameraModes;
+    [SerializeField] internal CinemachineVirtualCameraBase freeLookCamera;
+
+    [Header("InputActions")]
+    [SerializeField] internal InputAction_OneTime cameraChangeAction;
```
BrianAmadori: Remember we agreed upon changing the names to:
`InputAction_OneTime` -> `InputAction_Trigger` 
`InputAction_InTime` -> `InputAction_Hold`

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraController.cs
```diff
@@ -1,91 +1,156 @@
...
+            characterForward.Set(null);
+        }
+        else
         {
-            if (currentState == CameraState.FirstPerson)
-                currentState.Set(CameraState.ThirdPerson);
-            else
-                currentState.Set(CameraState.FirstPerson);
+            var xzPlaneForward = Vector3.Scale(cameraTransform.forward, new Vector3(1, 0, 1));
+            switch (currentMode)
```
BrianAmadori, line 105: The mode specific code shouldn't be inside a `CameraSetup` instead on a switch?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Camera/Tests/CameraTests.cs
```diff
@@ -1,169 +1,169 @@
...
-            cameraSetup.Activate();
-
-            config.Set(new ThirdPersonCameraConfig() { offset = Vector3.one * 6 });
-
-            Assert.AreEqual(config.Get().offset, dummyTransform.position);
-        }
-    }
-}
\ No newline at end of file
+﻿//using System.Collections;
```
BrianAmadori: Remember to remove commented code.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    public void Update_OneTime()
+    {
+        for (var i = 0; i < oneTimeActions.Length; i++)
+        {
+            var action = oneTimeActions[i];
+            switch (action.GetDCLAction())
+            {
+                case DCLAction_OneTime.Jump:
```
BrianAmadori: Ideally i'd try to set those action mappings in a more data-driven fashion. So we can set them modifying an inspector struct array or similar. I have a similar issue with the `InputProcessor` methods. If isn't too difficult to design this part of the system using an strategy pattern I'd give a shot at it.

BrianAmadori (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraController.cs
AjimenezDCL: Oopsie, I completely forgot. Gonna change it

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraController.cs
AjimenezDCL: I removed `CameraSetup` becase almost everything done there was overtaken by Cinemachine. I could create those `CameraSetup` and add this code, but it might be overkill.

Dont worry about the document/diagrams. Once the review is done I will update it.

BrianAmadori: Ok, as we agreed we are going to refactor this in the polish phase if we found it necessary.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
BrianAmadori: As we agreed, we are going to standby on this change and keep it as technical debt. As we have the following issues:

- We need to map actions to KeyCode / MouseButton / etc. and the required struct can get a bit messy.
- We may be gold plating this as Unity is rolling a new Input system, and we didn't investigate enough to know if it can save us this binding code.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
```diff
@@ -26,28 +27,28 @@ public enum DCLAction_Measurable
 
 public class InputController : MonoBehaviour
 {
-    public InputAction_OneTime[] oneTimeActions;
-    public InputAction_InTime[] inTimeActions;
+    [FormerlySerializedAs("oneTimeActions")] public InputAction_Trigger[] triggerTimeActions;
```
BrianAmadori: When the prefab is re-serialized and the values are copied from the old field to the new one you can remove this attribute.

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
AjimenezDCL: Oh yes, I was planing to do it but got caught in something else and forgot. Mini commit to fix it

unity-client/Assets/Scripts/MainScripts/DCL/Components/Billboard/Billboard.cs
```diff
@@ -48,7 +49,7 @@ public void Update()
 
         Vector3 GetLookAtVector()
         {
-            Vector3 lookAtDir = (entityTransform.position - DCLCharacterController.i.cameraTransform.position);
+            Vector3 lookAtDir = (entityTransform.position - cameraPosition);
```
pravusjif, line 52: This is actually wrong (it was wrong before you updated it, and i fixed it in the tests PR) the calculation should be the other way around: `cameraPosition - entityTransform.position` to get the vector that looks **towards the player**. But, leave it like this and I'll keep the fix when resolving this conflict with the tests fix PR

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/DCLCharacterController.cs
```diff
@@ -267,40 +241,38 @@ void Update()
...
+                transform.forward = characterForward.Get().Value;
+                CommonScriptableObjects.playerUnityEulerAngles.Set(transform.eulerAngles);
+            }
         }
 
-        if (jumpButtonPressed && (Time.time - lastJumpButtonPressedTime < 0.15f)) // almost-grounded jump button press allowed time
+        if (jumpButtonPressed)
         {
-            if (isGrounded || (Time.time - lastUngroundedTime) < 0.1f) // just-left-ground jump allowed time
+            if (Time.time - lastJumpButtonPressedTime < 0.15f) // almost-grounded jump button press allowed time
```
pravusjif: This is not working correclty, try it out, press the jump button a moment before hitting the ground (keep pressing it) and the chara should start jumping as soon as it hits the ground, it worked before, you can compare with .zone: https://explorer.decentraland.zone/

pravusjif: We also lost the "just ungrounded jump" feature

pravusjif (changes requested):

pravusjif (approved):

PR #184, opened 2019-12-03
fix: asset promise keeper test fixes

BrianAmadori:
# Fixes 
- fix: AssetPromise_GLTF tests assembly wasn't being compiled at all, so tests were missing
- fix: A single forceNewInstance test was failing.
- fix: MemoryManager wasn't being used at all


unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Singleton.cs
```diff
@@ -14,5 +14,12 @@ public static T i
                 return instance;
             }
         }
+
+        public virtual void Initialize()
```
AjimenezDCL: Why cannot we make `Initialize` private and call it in the getter if we are creating the instance?

AjimenezDCL (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Singleton.cs
BrianAmadori: Because I want to call it explicitly in `SceneController` initialization. The `MemoryManager` class overrides it with some custom stuff. 

The alternative is to move the custom stuff to the constructor and then do something super confusing like `_ = MemoryManager.i;` to create the singleton instance.

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Singleton.cs
AjimenezDCL: Fair enough!

AjimenezDCL (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Singleton.cs
BrianAmadori: I think i'm going to move `Initialize()` to the `MemoryManager` directly instead of having it as virtual. Sounds a bit cleaner to me.

PR #199, opened 2019-12-05
fix: FPS panel not null

eordano:
makes logs a lot cleaner


BrianAmadori (commented):
Care to clarify a bit what is this changing exactly? Why are we merging this to `fix/unittest_pass_again` and not `master`?

eordano:
I'm seeing this every time I run the tests (once per *every* single unit test)

image

I'm merging this to "unit tests shall pass again" because it seems more in line with that epic (in the spirit of "feature branches")

pravusjif (approved):

PR #222, opened 2019-12-11
feat: interaction hover feedback

pravusjif:
* Added logic for interactive objects detection (OnClick and OnPointerDown components)
* Added place holder prefab that will be tuned up later by Romi
* Disabled hover feedback until the implementation is finished in the next PR (`PointerEventsController.enableInteractionHoverFeedback`)


BrianAmadori (commented):
`edited`

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InteractionHoverCanvasController.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        ConfigureIcon(button);
+
+        Hide();
+    }
+
+    void ConfigureIcon(WebInterface.ACTION_BUTTON button)
+    {
+        // When we allow for custom input key bindings this implementation will change
+        switch (button)
```
BrianAmadori: I think we should just use prefabs instead of setting the UI element properties like this. It'd be more manageable and Romi could take care of it without changing code, as we are already doing with the avatar editor stuff.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PointerEventsController.cs
```diff
@@ -31,6 +29,56 @@ public void Initialize(bool isTesting = false)
...
+                hoverInteractiveObjectsRoutine = SceneController.i.StartCoroutine(HoverInteractiveObjects());
+        }
+
+        IEnumerator HoverInteractiveObjects()
+        {
+            RaycastHit hitInfo;
+
+            while (true)
+            {
+                if (RenderingController.i.renderingEnabled)
```
BrianAmadori: We can invert this `if` to avoid indentation.

BrianAmadori (changes requested):
We should add at least a single test that checks the toast is being displayed as intended.

Also, all the assets related to this feature should be under a feature related folder. Putting everything inside the same root `Resources` folder will become unmanageable in the long run.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InteractionHoverCanvasController.cs
pravusjif: OK, but in the future we will probably come back to an approach similar to the current one, when we let the users bind the input to any key

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PointerEventsController.cs
pravusjif: Done

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InteractionHoverCanvasController.cs
pravusjif: Done, made 4 separate hover icon prefabs (one for each action, including the "unknown" input action as well)

pravusjif:
> We should add at least a single test that checks the toast is being displayed as intended.
> 
> Also, all the assets related to this feature should be under a feature related folder. Putting everything inside the same root `Resources` folder will become unmanageable in the long run.

Added a test

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InteractionHoverCanvas/InteractionHoverCanvasController.cs
```diff
@@ -0,0 +1,95 @@
...
+                break;
+            case WebInterface.ACTION_BUTTON.SECONDARY:
+                prefabPath = SECONDARY_ICON_PREFAB_PATH;
+                break;
+            default: // WebInterface.ACTION_BUTTON.UNKNOWN
+                prefabPath = UNKNOWN_ICON_PREFAB_PATH;
+                break;
+        }
+
+        hoverIcon = Object.Instantiate(Resources.Load(prefabPath), backgroundTransform) as GameObject;
```
BrianAmadori: Why are we using `Resources.Load` here? As this is a `MonoBehaviour`, we could reference the prefabs directly using this prefab serialization. This is better because we don't depend on the object names/path for this to work.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PointerEventsController.cs
```diff
@@ -40,15 +97,15 @@ public void Cleanup()
             InputController_Legacy.i.RemoveListener(WebInterface.ACTION_BUTTON.SECONDARY, OnButtonEvent);
```
BrianAmadori, line 97: We shouldn't kill the coroutine here too?

unity-client/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/OnPointerEventComponent.cs
```diff
@@ -1,21 +1,36 @@
 using DCL.Controllers;
 using DCL.Models;
 using DCL.Helpers;
+using DCL.Interface;
 using UnityEngine;
 
 namespace DCL.Components
 {
-    public class OnPointerEventComponent : UUIDComponent
+    public class OnPointerEventComponent : UUIDComponent<OnPointerEventComponent.Model>
```
BrianAmadori: The `Component` name in all those family of classes is redundant. I think the code is a bit more readable if we remove it.
```suggestion
    public class OnPointerEvent : UUIDComponent<OnPointerEvent.Model>
```

BrianAmadori: Just checked and it doesn't seem to produce any name collision. So its safe to rename it.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InteractionHoverCanvas/InteractionHoverCanvasController.cs
```diff
@@ -0,0 +1,95 @@
+﻿using UnityEngine;
```
BrianAmadori: I think this class and all the resources related to it should be under a subfolder in the `UUIDComponent` subfolder, as its only used by the `PointerEvent` base class.

BrianAmadori (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InteractionHoverCanvas/InteractionHoverCanvasController.cs
pravusjif: OK, let's keep that as a rule then, yesterday I almost ask you how you'd like this instantiation to be made and ended up using the Resources way because we use that everywhere i think, then let's always use referene instantiation in monobehaviours, I'm changing it now.

pravusjif: done

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PointerEventsController.cs
pravusjif: I don't think so, since there is no persistent data related to this, but I added a little cleanup code (reset the `lastHoveredObject` and `newHoveredObject` vars)

unity-client/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/OnPointerEventComponent.cs
pravusjif: Done

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InteractionHoverCanvas/InteractionHoverCanvasController.cs
pravusjif: Done

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PointerEventsController.cs
BrianAmadori: If we don't kill coroutines on Cleanup methods consecutive tests might fail in the cases of not reloading scenes. Remember we are going in that direction now with the test perf. fixes.

unity-client/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/OnPointerDownComponent.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace DCL.Components
 {
-    public class OnPointerDownComponent : OnPointerEventComponent
+    public class OnPointerDownComponent : OnPointerEvent
```
BrianAmadori: We shouldn't be renaming those classes too?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PointerEventsController.cs
pravusjif: The thing is, if the scene is not reloaded, the SceneController doesn't call its Awake, the PointerEventsController.Initialize is not called and this coroutine never starts again, if we want that... then I can add the StopCoroutine, no problem

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/OnPointerDownComponent.cs
pravusjif: Yes, you're right, i just renamed OnPointerDownComponent, OnPointerUpComponent and OnClickComponent

pbosio (approved):

PR #227, opened 2019-12-12
fix: camera feel polish improvements

BrianAmadori:
List of changes related to feel:

* Walk/run speed was _decreased_ to make animations look correctly. Now holding shift key makes character walk instead running. Running is the default movement action.  This also has the side effect of giving more time to the furthest scenes to load.

* Animation jitter fixes. Avatar rotates correctly according to its movement direction. 

* Avatar framing polish according to @RominaMarchetti @ShibuyaMartin @MadRex and myself.

* When strafing the avatar rotates to the left/right, and camera follows (without the mario64 rotation style).

* First draft of camera collision behaviour

* fix: Going to third person after rotating in first person could cause the camera to "remember" the previous rotation angle incorrectly, in a way that third person started looking at the avatar sideways when it shouldn't.

* fix: Avatar was clipping the floor in most cases.

* fix: Avatars lacked shadows.

Known issues:

* Camera jitters when avatar is on top of moving platforms.

* We still have many camera collision issues. Some of them caused by bad collider placement. This known to be a very difficult theme in all games that implement it and will be an ongoing effort.


eordano:
Please rebase/solve conflicts

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_GLTF.cs
```diff
@@ -50,10 +50,12 @@ public override void Load(string targetUrl, Action<LoadWrapper> OnSuccess, Actio
 
             alreadyLoaded = false;
 
-            if (useGltfFallback)
-                LoadAssetBundle(targetUrl, OnSuccess, (x) => LoadGltf(targetUrl, OnSuccess, OnFail));
-            else
-                LoadAssetBundle(targetUrl, OnSuccess, OnFail);
+            LoadGltf(targetUrl, OnSuccess, OnFail);
+
+            //if (useGltfFallback)
```
eordano: No commented code please

eordano (approved):
All looking good, one minor comment only: please delete commented code

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraController.cs
```diff
@@ -74,84 +50,48 @@ private void OnRenderingStateChanged(bool enabled)
...
-        freeLookCamera.gameObject.SetActive(active);
+        cachedModeToVirtualCamera[currentMode].OnSelect();
     }
 
     private void PrecisionChanged(Vector3 newValue, Vector3 oldValue)
     {
         transform.position += newValue - oldValue;
     }
 
+    public GameObject cursor;
```
pravusjif: Where is this being used?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraStateFPS.cs
```diff
@@ -0,0 +1,49 @@
...
+    bool hasLocalCameraUnselectValues = false;
+    public override void OnSelect()
+    {
+        base.OnSelect();
+    }
+
+    public override void OnUnselect()
+    {
+        base.OnUnselect();
+    }
```
pravusjif: This overrides look unnecessary as they only call the same base class method inside

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraStateTPS.cs
```diff
@@ -0,0 +1,91 @@
...
+            defaultVirtualCamera.m_YAxis.Value = 0.5f;
+        }
+
+        base.OnSelect();
+    }
+
+    public override void OnUnselect()
+    {
+        base.OnUnselect();
+    }
```
pravusjif: We could remove this override

pravusjif (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraController.cs
BrianAmadori: Initially I was hiding the cursor in TPS, but I deleted the code later because its better to hide the cursor when we implement the loose aim stuff. So, this code remained as vestigial. Removing it.

pravusjif (approved):
Tested the build and it looks good

PR #230, opened 2019-12-13
feat: run tests on circleci

eordano:


moliva:
I see a `Too long with no output (exceeded 10m0s): context deadline exceeded` possibly from CircleCI. We could change the value to something bigger as we know they take time, as in:
```
- run:  
    command: ...
    no_output_timeout: 30m
```
See: https://support.circleci.com/hc/en-us/articles/360007188574-Build-has-hit-timeout-limit

Reference: https://circleci.com/gh/decentraland/explorer/3223

.circleci/config.yml
```diff
@@ -366,6 +419,13 @@ workflows:
               only: /(.*)/
             tags:
               only: /^\d+\.\d+\.\d(.*)/ # npm tag + latest
+      - run-unity-tests:
+          <<: *all_branches_and_tags
+          requires:
+            - build-unity
+          filters:
+            branches:
+              only: /(master)|(release)|(staging)|(feat\/runtest)/
```
moliva: IMO, we should change this to run in every branch if possible.
```suggestion
              only: /(.*)/
```

eordano:
Running on all branches is not feasible economically (40 runs are 150 usd).

eordano:
image
Ready to merge :D

unity-client/Assets/Scripts/Tests/CharacterControllerTests/CharacterControllerTests.cs
```diff
@@ -133,6 +133,7 @@ public IEnumerator Character_UpdateSOPosition()
         }
 
         [UnityTest]
+        [NUnit.Framework.Explicit("fixme (@pravus)")]
```
BrianAmadori: Please add `[Category("Explicit")]` so we can check the explicit tests easily via the Test Runner editor window.

BrianAmadori (approved):

unity-client/Assets/Scripts/Tests/CharacterControllerTests/CharacterControllerTests.cs
eordano: ouch, this slipped me, thanks

.circleci/config.yml
```diff
@@ -366,6 +413,13 @@ workflows:
               only: /(.*)/
             tags:
               only: /^\d+\.\d+\.\d(.*)/ # npm tag + latest
+      - run-unity-tests:
```
moliva, line 416: Normally I would ask the publish renderer step to require this to pass, maybe we can add this in the short future after checking that tests are running smoothly.

moliva (approved):

.circleci/config.yml
eordano: adding it!

PR #238, opened 2019-12-19
feat: integrating new communications mechanism with lighthouse

moliva:
# What? 
Introduces the new comms interface to connect via Lighthouse.

It still uses the previous `v1` implementation as the default, maintaining compatibility between the two of them, but enables us to start testing with the new protocol.

Establish a new way to switch between different communication modes with the `COMMS` query parameter:
- `v1:local`: local comms using web sockets (previously `LOCAL_COMMS`)
- `v1:remote`: default mode as of today, broker communication server
- `v2:p2p`: uses new communication lighthouse in p2p mode
- `v2:server`: uses new communication lighthouse where all communications passes through the server (similar to v1 implementation)


kernel/packages/shared/comms/interface/types.ts
```diff
@@ -91,3 +91,21 @@ export type Pose = [number, number, number, number, number, number, number]
 export type PoseInformation = {
   v: Pose
 }
+
+export type Package<T> = {
+  type: 'profile' | 'chat' | 'position'
+  time: number
+  data: T
+}
```
eordano, line 99: As discussed yesterday with @pablitar lets add a TTL here

kernel/packages/shared/comms/index.ts
```diff
@@ -222,15 +235,20 @@ export function processChatMessage(context: Context, fromAlias: string, data: Ch
...
   processNewLogin(identity, context, fromAlias)
 
-  const msgTimestamp = data.getTime()
+  const msgTimestamp = data.time
 
   const peerTrackingInfo = ensurePeerTrackingInfo(context, fromAlias)
 
   if (msgTimestamp > peerTrackingInfo.lastProfileUpdate) {
-    const profileVersion = data.getProfileVersion()
+    const profileVersion = data.data.version
```
pablitar: Why do we have two things named "data" nested with each other?

kernel/packages/shared/comms/interface/types.ts
pablitar: This is actually a layer above the Peer Protocol.

The TTL, hops, and all network related messages will be handled inside the Peer Library. The idea is to try not to couple Kernel and Explorer to these "implementation details", as much as that is feasible.

I'll be working in the gossip implementation in this branch: https://github.com/decentraland/katalyst/tree/feat/gossip-network

I'll try to have something to show rather sooner than later.

kernel/packages/shared/comms/v2/LighthouseWorldInstanceConnection.ts
```diff
@@ -0,0 +1,139 @@
...
+      }
+    })
+    const leaving = currentRooms.map(current => {
+      if (!rooms.some(room => current.id === room)) {
+        return this.peer.leaveRoom(current.id)
+      } else {
+        return Promise.resolve()
+      }
+    })
+    return Promise.all([...joining, ...leaving]).then(NOOP)
```
pablitar, line 137: Why is this `then` necessary?

pablitar (approved):
Nice work!!

kernel/packages/shared/comms/v2/LighthouseWorldInstanceConnection.ts
moliva: Mostly because of the return type of the method.

kernel/packages/shared/comms/index.ts
moliva: It's because of how it was historically called the package that is now the wrapper of the message. Changing this to avoid mixing the two.

eordano (approved):

PR #259, opened 2019-12-23
fix: messaging buses idle performance

BrianAmadori:
Messaging buses were taking 13 ms of CPU time, even when no messages where being processed.

Modified the code so the CPU overhead drops back to zero in the case of exploring scenes with no systems in place.

Main offenders were:
- Multiple calls of dictionary `ContainsKey` per frame.
- Multiple calls of dictionary getters.
- Multiple calls of `!= string` per frame.
- ProcessMessages were using Time.realtimeSinceStartup even when no messages where being enqueued.
- When no messages were processed, instead of skipping a frame the internal loop continued to execute until the useless overhead surpassed the time budget.


eordano:
I'd try to either:

* propose simpler fixes (less lines of code or less nested coupled complexity)
* add tests

Otherwise, making sure this is correct is very hard

BrianAmadori:
@eordano this PR does micro-optimizations more than anything else. I can't think of a good way to test it that doesn't take too much time. About the fixes complexity, I actually reduced the code complexity in the messages processing side, so maybe i'd be better if you point out concrete examples.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingController.cs
```diff
@@ -42,32 +42,46 @@ public enum QueueState
         public Dictionary<string, MessagingBus> messagingBuses = new Dictionary<string, MessagingBus>();
         public IMessageHandler messageHandler;
         public string debugTag;
+        public bool sleeping;
```
eordano: What’s the definition of sleeping? Under which circumstances is this triggered, what kinds of things are safe/unsafe to do while sleeping?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
```diff
@@ -294,21 +348,24 @@ bool ProcessBus(MessagingController controller, string busId, ref float prevTime
...
+                bool queueResult = bus.ProcessQueue(timeBudget, out yieldReturn);
+
+                bus.owner?.RefreshSleepingState();
+
+                if (queueResult)
+                    return true;
+
+                prevTimeBudget -= Time.realtimeSinceStartup - startTime;
+
+                if (prevTimeBudget <= 0)
```
eordano, line 358: What is the process by which this value becomes non negative again?

eordano (commented):
Here are two examples of accidental complexity I found. One is new, the other one is inherited

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingController.cs
eordano: Coupled complexity: one needs to know the definition of sleeping to code on this or related classes, and it’s far from evident that “sleeping” means “idle queue”

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
pravusjif: At the beginning of the `ProcessBus ()` ( inside the `if (bus.isRunning && bus.pendingMessagesCount > 0)` ) the `float startTime` is initialized with the game elapsed time at that moment, later when `prevTimeBudget` is re-calculated we know if there is time budget available for processing these messages, so if the prevTimeBudget is 0 or less, we have to escape, to avoid processing messages in that case.

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingController.cs
BrianAmadori: `sleeping` just means "will not be processed". The queue will sleep when its out of messages. I prefer flag names to state the behaviour it toggles and not infer the condition under it which changes value, because that can (and will) change and become more complex, and in that case the name would get outdated and make the code harder to read.

This said, maybe I can change the name to `enabled` and invert the conditions. Maybe `enabled` is more conventional and makes it easier to infer its purpose?

BrianAmadori: On top of this `sleeping` connotates a negation, and this can lead to double negatives.

BrianAmadori: renaming done.

pbosio (approved):

PR #261, opened 2019-12-24
fix: rely on "dirty" attribute

eordano:
How to test this:

* set `DEBUG_SYNC_SYSTEM` to true
* make sure all the logging entries are always good misses or good hits

Start reviewing here: https://github.com/decentraland/explorer/pull/261/files#diff-009b9ffa1533b81274550200ceeba6beL143


kernel/packages/decentraland-ecs/src/decentraland/Implementation.ts
```diff
@@ -15,9 +15,34 @@ import {
...
 // This number is defined in the protocol ECS.SetEntityParent.3
 const ROOT_ENTITY_ID = '0'
 
+const dirtyLegend: { [key: number]: string } = {
+  0: "`dirty` is not a flag, but don't send anything because nothing changed",
+  1: '`dirty` is not actually a flag in components. Use sameJSON comparison',
+  10: "Damn, it's the same JSON but it's dirty? we're in trouble",
+  11: "It's dirty, and we know it",
+  100: "It's not dirty, and we know it.",
+  101: "In deep shit. It's dirty, and we don't know it."
```
eordano:
```suggestion
  101: "BAD: JSON cache is different, but `dirty` was false."
```

kernel/packages/decentraland-ecs/src/decentraland/Implementation.ts
```diff
@@ -15,9 +15,34 @@ import {
...
+
 // This number is defined in the protocol ECS.SetEntityParent.3
 const ROOT_ENTITY_ID = '0'
 
+const dirtyLegend: { [key: number]: string } = {
+  0: "`dirty` is not a flag, but don't send anything because nothing changed",
+  1: '`dirty` is not actually a flag in components. Use sameJSON comparison',
+  10: "Damn, it's the same JSON but it's dirty? we're in trouble",
+  11: "It's dirty, and we know it",
+  100: "It's not dirty, and we know it.",
```
eordano:
```suggestion
  101: "GOOD: JSON cache is different, and `dirty` was true."
```

kernel/packages/decentraland-ecs/src/decentraland/Implementation.ts
```diff
@@ -15,9 +15,34 @@ import {
...
+const SAFE_INEFFICIENT_COMPARE = false
+const DEBUG_SYNC_SYSTEM = false
+
 // This number is defined in the protocol ECS.SetEntityParent.3
 const ROOT_ENTITY_ID = '0'
 
+const dirtyLegend: { [key: number]: string } = {
+  0: "`dirty` is not a flag, but don't send anything because nothing changed",
+  1: '`dirty` is not actually a flag in components. Use sameJSON comparison',
+  10: "Damn, it's the same JSON but it's dirty? we're in trouble",
```
eordano:
```suggestion
  10: "Maybe BAD: JSON cache is a hit, but `dirty` was true."
```

kernel/packages/decentraland-ecs/src/decentraland/Implementation.ts
```diff
@@ -15,9 +15,34 @@ import {
...
+  11: "It's dirty, and we know it",
```
eordano:
```suggestion
  11: "GOOD: JSON cache is a hit, and `dirty` was false."
```

kernel/packages/decentraland-ecs/src/decentraland/Implementation.ts
```diff
@@ -139,16 +160,10 @@ export class DecentralandSynchronizationSystem implements ISystem {
         const component = entity.components[componentName]
         const classId = getComponentClassId(component)
 
-        if (classId !== null && !isDisposableComponent(component)) {
-          const componentJson: string = JSON.stringify(component)
```
eordano, line 143: Unless I'm reading this wrong, we're f*ing it up here: every frame, every "non disposable component" -- that changed OR DID NOT CHANGED, will get `JSON.stringified`.

This includes all `Transforms` if I'm not mistaken

moliva: Indeed. My two questions regarding this would be:
- Why are we doing this extra check and not the actual `dirty` flag already in the components? Would we be missing any change there?
- How is the `JSON#stringify` affecting our performance?
Maybe we should run a comparison between the two and check the cases were one seems dirty and the other is not to make sure we are not missing any important case.

eordano: Exactly! That's why I added this function that logs which cases do we stumble upon. In all cases I've seen, dirty was a reliable present attribute.

kernel/packages/decentraland-ecs/src/decentraland/Implementation.ts
```diff
@@ -162,6 +178,59 @@ export class DecentralandSynchronizationSystem implements ISystem {
...
+  private efficientCompare(entity: IEntity, entityId: string, component: any, componentName: string, classId: any) {
+    if (classId !== null && !isDisposableComponent(component) && component.dirty) {
+      this.sendUpdateEntityComponent(entity, component, componentName, classId, JSON.stringify(component), entityId)
+    }
+  }
+  private inefficientCompare(entity: IEntity, entityId: string, component: any, componentName: string, classId: any) {
+    if (classId !== null && !isDisposableComponent(component)) {
+      const componentJson: string = JSON.stringify(component)
+
+      const isDirty = component.dirty === undefined
```
eordano:
```suggestion
      const isDirty = component.dirty
```

pravusjif (approved):

pravusjif (approved):

moliva (approved):

PR #325, opened 2020-01-08
feat: `/goto magic` and `/goto random`

eordano:


pravusjif (approved):
It works, personally I'd prefer having the "/goto random" as a true random between the 90k parcels, and have another "/goto randompoi" or similar that takes the user to one of the POIs as it does right now.

AjimenezDCL (approved):

eordano:
Adding `/goto magic` for pois and `/goto random` for a true random parcel! Please re-test it!

PR #333, opened 2020-01-12
feat: tweaks of performance to the bus priorities

eordano:
[This PR started as only moving code around: behavior should be pretty similar]

I was having a hard time visualizing bus queue priorities: I think this slightly improves the visibility of that, what do you think @BrianAmadori @pravusjif? I tried to use less lines so that the priority between queues is clearly distinguished. I made some changes that I hope should improve start up time.

### DIFF
--- a/unity-client/Assets/Scenes/InitialScene.unity
+++ b/unity-client/Assets/Scenes/InitialScene.unity
@@ -334,8 +334,8 @@ MonoBehaviour:
   customContentServerUrl: http://localhost:1338/
   baseUrlMode: 0
   baseUrlCustom: 
-  environment: 2
-  startInCoords: {x: 64, y: -64}
+  environment: 4
+  startInCoords: {x: -6, y: -20}
   forceLocalComms: 1
   allWearables: 0
   testWearables: 0
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
@@ -220,81 +220,86 @@ namespace DCL
 
         IEnumerator ProcessMessages()
         {
-            IEnumerator yieldReturn = null;
-            float prevTimeBudget = INIT_MSG_BUS_BUDGET_MAX;
-            float start = Time.unscaledTime;
 
-            // This loop makes sure that queues are emptied out in the correct order.
-            // Every time we're done with a queue, we call `continue` to start again processing messages in the right priority.
-            // `ProcessEventsFromBus`, the function called for each bus, will only process one event/task.
-            // That's a very important aspect of its interface. it would be great if we could enforce that somehow.
-            // This is important because there's a maximum budget of time alloted for this.
-            // It would be good to check if there's an alternative implementation of this,
-            // maybe we could break earlier so we don't waste time checking on empty queues?
-            while (Time.realtimeSinceStartup - start >= GLOBAL_MAX_MSG_BUDGET)
+            while (true)
             {
-                // TODO: If in the previous iteration we found a corutine to run, shouldn't we call it?
-                // What is the output `yieldReturn` parameter being used for?
-                // if (yieldReturn != null)
-                //     yield return yieldReturn;
+                IEnumerator yieldReturn = null;
+                float start = Time.unscaledTime;
+                float prevTimeBudget = INIT_MSG_BUS_BUDGET_MAX;
 
-                // High priority buses: UI events, global initialization events, and UI initialization events
-                if (ProcessEventsFromBus(uiSceneController, uiSceneController?.uiBus, ref prevTimeBudget, ref yieldReturn))
-                    continue;
-                if (ProcessEventsFromBus(globalController, globalController?.initBus, ref prevTimeBudget, ref yieldReturn))
-                    continue;
-                if (ProcessEventsFromBus(uiSceneController, uiSceneController?.initBus, ref prevTimeBudget, ref yieldReturn))
-                    continue;
-
-                // Next in priority: events for the current scene
-                if (ProcessEventsFromBus(currentSceneController, currentSceneController?.initBus, ref prevTimeBudget, ref yieldReturn))
-                    continue;
-                if (ProcessEventsFromBus(currentSceneController, currentSceneController?.uiBus, ref prevTimeBudget, ref yieldReturn))
-                    continue;
-                if (ProcessEventsFromBus(currentSceneController, currentSceneController?.systemBus, ref prevTimeBudget, ref yieldReturn))
-                    continue;
-
-                // Then: other initialization events from scenes
-                bool shouldRestart = false;
-                for (int i = 0; i < sortedControllersCount; i++)
+                // This loop makes sure that queues are emptied out in the correct order.
+                // Every time we're done with a queue, we call `continue` to start again processing messages in the right priority.
+                // `ProcessEventsFromBus`, the function called for each bus, will only process one event/task.
+                // That's a very important aspect of its interface. it would be great if we could enforce that somehow.
+                // This is important because there's a maximum budget of time alloted for this.
+                // It would be good to check if there's an alternative implementation of this,
+                // maybe we could break earlier so we don't waste time checking on empty queues?
+                while (Time.realtimeSinceStartup - start <= GLOBAL_MAX_MSG_BUDGET)
                 {
-                    if (ProcessEventsFromBus(sortedControllers[i], sortedControllers[i].initBus, ref prevTimeBudget, ref yieldReturn))
-                    {
-                        shouldRestart = true;
-                        break;
-                    }
-                }
-                if (shouldRestart)
-                    continue;
 
-                // Low priority: UI events for scenes
-                for (int i = 0; i < sortedControllersCount; i++)
-                {
-                    if (ProcessEventsFromBus(sortedControllers[i], sortedControllers[i].uiBus, ref prevTimeBudget, ref yieldReturn))
-                    {
-                        shouldRestart = true;
-                        break;
-                    }
-                }
-                if (shouldRestart)
-                    continue;
+                    // High priority buses: UI events, global initialization events, and UI initialization events
+                    if (ProcessEventsFromBus(uiSceneController, uiSceneController?.uiBus, ref prevTimeBudget, ref yieldReturn))
+                        continue;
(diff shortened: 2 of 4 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
```diff
@@ -205,133 +205,98 @@ public void UpdateThrottling()
...
-            float start;
-
-            while (true)
+            if (controller != null && controller.enabled)
             {
-                prevTimeBudget = INIT_MSG_BUS_BUDGET_MAX;
-                start = Time.unscaledTime;
-
-                bool processedBus = false;
+                if (ProcessBus(bus, ref prevTimeBudget, out yieldReturn))
```
eordano: What is `yieldReturn` used for?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
```diff
@@ -340,9 +310,6 @@ bool ProcessBus(MessagingBus bus, ref float prevTimeBudget, out IEnumerator yiel
             if (bus.isRunning && bus.pendingMessagesCount > 0)
             {
                 float startTime = Time.realtimeSinceStartup;
-
-                yieldReturn = null;
```
eordano, line 344: Can you advise on whether this is correct?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
```diff
@@ -205,133 +205,103 @@ public void UpdateThrottling()
...
-
-                bool processedBus = false;
-
-                // When breaking this second loop, we skip a frame
-                while (true)
+                IEnumerator yieldReturn = null;
+                float start = Time.unscaledTime;
+                float prevTimeBudget = INIT_MSG_BUS_BUDGET_MAX;
+
+                // This loop makes sure that queues are emptied out in the correct order.
```
eordano: Please read these comments carefully!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
```diff
@@ -205,130 +210,110 @@ public void UpdateThrottling()
...
             while (true)
             {
-                prevTimeBudget = INIT_MSG_BUS_BUDGET_MAX;
-                start = Time.unscaledTime;
-
-                bool processedBus = false;
-
-                // When breaking this second loop, we skip a frame
-                while (true)
+                start = Time.realtimeSinceStartup;
```
eordano, line 245:
```suggestion
                start = Time.realtimeSinceStartup;
                prevTimeBudget = INIT_MSG_BUS_BUDGET_MAX;
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
eordano: If I messed up with the description please correct me. I'm pretty sure I have a bad mental model.

unity-client/Assets/Scenes/InitialScene.unity
```diff
@@ -334,8 +334,8 @@ MonoBehaviour:
   customContentServerUrl: http://localhost:1338/
   baseUrlMode: 0
   baseUrlCustom: 
-  environment: 2
-  startInCoords: {x: 64, y: -64}
+  environment: 4
```
eordano, line 337: This switches to `.org` by default

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
BrianAmadori: Some components may have updates that takes more than one frame (mostly UI ones). In that case they provide an `IEnumerator` that should be yielded in order for those components to be updated correctly, that `IEnumerator` is filled out in that `yieldReturn` variable.

I'm reading changes from many commits ago and i'm noticing this `yieldReturn` is never yielded, this is a case of "Why this was working all this time along?" because not yielding it should break components.

For this to work correctly, we should check for nullity and yield.
```
if (yieldReturn != null) 
    yield yieldReturn;
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
```diff
@@ -205,130 +210,113 @@ public void UpdateThrottling()
...
-                while (true)
+                // This loop makes sure that queues are emptied out in the correct order.
+                // Every time we're done with a queue, we call `continue` to start again processing messages in the right priority
+                // (note that the next run might be in the next frame due to the time budget).
+                // `ProcessEventsFromBus`, the function called for each bus, will only process one event/task.
+                // That's a very important aspect of its interface. it would be great if we could enforce that somehow with the type system.
+                // This is important because there's a maximum budget of time alloted for this.
+                // It would be good to check if there's an alternative implementation of this,
+                // maybe we could break earlier so we don't waste time checking on empty queues?
+                while (Time.realtimeSinceStartup - start <= GLOBAL_MAX_MSG_BUDGET)
```
BrianAmadori, line 253: I'm not sure if this is more readable than having `while (true)` and checking at the end, because we have more than one condition. Moreover, we are missing the `processedBus` check. This is important because this fixed the case of having a big chunk of time doing nothing when no buses have messages.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
```diff
@@ -205,130 +210,113 @@ public void UpdateThrottling()
...
-                            break;
-                    }
-
-                    //-------------------------------------------------------------------------------------------
-                    // Rest of the scenes INIT
                     bool shouldRestart = false;
 
-                    for (int i = 0; i < sortedControllersCount; i++)
+                    // Highest priority: If rendering is not activated, go straight to the init events
+                    if (!IsRenderingActivated())
```
BrianAmadori: We have evidence that this is actually saving CPU time? Because we are already prioritizing init messages without this code and the load time bottleneck is not on this side but more on download times. This is not needed imho.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
```diff
@@ -205,130 +210,113 @@ public void UpdateThrottling()
...
-                    //-------------------------------------------------------------------------------------------
-                    // Global Controller INIT
-                    if (globalController != null && globalController.enabled)
+                    if (yieldReturn != null)
                     {
-                        processedBus = true;
-                        if (ProcessBus(globalController.initBus, ref prevTimeBudget, out yieldReturn))
-                            break;
+                        // Commenting this line improves performance by a lot. It seems to be locking until it returns.
+                        // yield return yieldReturn;
```
BrianAmadori: I'm reading changes from many commits ago and i'm noticing this yieldReturn is never yielded, this is a case of "Why this was working all this time along?" because not yielding it should break components.

For this to work correctly, we should check for nullity and yield.
```
if (yieldReturn != null) 
    yield yieldReturn;
```
I think this should be done in the outer loop.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
BrianAmadori: This should be cool because the `yieldReturn` is filled out by `bus.ProcessQueue(timeBudget, out yieldReturn);
` line.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
```diff
@@ -205,130 +210,113 @@ public void UpdateThrottling()
             }
         }
 
+        /**
+         * Calls to ProcessBus to run events from this queue. Returns true if ProcessBus returns true
+         */
+        private bool ProcessEventsFromBus(MessagingController controller, MessagingBus bus, ref float prevTimeBudget, ref IEnumerator yieldReturn)
```
BrianAmadori: For consistency sake we should use `out IEnumerator yieldReturn`. The only difference with `ref` is that we expect a new value to be filled out instead of modifying an existing one.

BrianAmadori: This abstraction looks like its saving a lot of code and that's generally good, but be very careful with it as it this is a critical section of the code. Naive calls like the ones to this method may introduce unwanted bottlenecks. I had to make a lot of micro-optimizations to this code in my last performance commit, and this will introduce repeated `if` calls in cases that were grouped. Please profile.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
```diff
@@ -205,130 +210,113 @@ public void UpdateThrottling()
...
-
-                // When breaking this second loop, we skip a frame
-                while (true)
+                // This loop makes sure that queues are emptied out in the correct order.
+                // Every time we're done with a queue, we call `continue` to start again processing messages in the right priority
+                // (note that the next run might be in the next frame due to the time budget).
+                // `ProcessEventsFromBus`, the function called for each bus, will only process one event/task.
+                // That's a very important aspect of its interface. it would be great if we could enforce that somehow with the type system.
+                // This is important because there's a maximum budget of time alloted for this.
+                // It would be good to check if there's an alternative implementation of this,
```
BrianAmadori: Please let's keep code quality discussions in this medium instead of code comments. Please check the `processedBus` check in the former version of this file for empty queue optimization.

BrianAmadori (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
eordano: I replaced the need for processedBus with "continue" and a yield null at the end

PR #335, opened 2020-01-12
feat: don't necessarily lock Corutines on WebRequests

eordano:
From the documentation in Unity,

> ### Do not block on WWW or WebRequest downloads
> Do not use code which blocks on a WWW or WebRequest download, like this:
> 
> ```
> while(!www.isDone) {}
> ```
> 
> Blocking on WWW or WebRequest downloads does not work on Unity WebGL. Because WebGL is single threaded, and because the XMLHttpRequest class in JavaScript is asynchronous, your download never finishes unless you return control to the browser; instead, your content deadlocks. Instead, use a Coroutine and a yield statement to wait for the download to finish.

It's not clear whether this improves or not, but it works at least as well as before

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs
@@ -134,7 +134,11 @@ public class NFTShapeLoaderController : MonoBehaviour
         string jsonURL = $"{NFTDataFetchingSettings.DAR_API_URL}/{darURLRegistry}/asset/{darURLAsset}";
 
         UnityWebRequest www = UnityWebRequest.Get(jsonURL);
-        yield return www.SendWebRequest();
+        www.SendWebRequest();
+        while (!www.isDone)
+        {
+            yield return null;
+        }
 
         if (!www.WebRequestSucceded())
         {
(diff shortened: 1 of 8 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs
```diff
@@ -134,7 +134,11 @@ IEnumerator FetchNFTImage()
         string jsonURL = $"{NFTDataFetchingSettings.DAR_API_URL}/{darURLRegistry}/asset/{darURLAsset}";
 
         UnityWebRequest www = UnityWebRequest.Get(jsonURL);
-        yield return www.SendWebRequest();
+        www.SendWebRequest();
```
BrianAmadori, line 137: Those changes aren't addressing the Unity documentation recommendations. 

Note that the doc prohibits you from doing:
```
while(!www.isDone) {}
```
That is effectively a blocking while, and not the same as doing:
```
while (!www.isDone) { yield return null; }
```
That while is yielded, and will not block anything by definition.

These changes can be merged, but imho they won't change anything of substance on the loading behaviour.

eordano: Cool! This is not for the loading, i was seeing busy loops blocking the renserer

BrianAmadori: Ok. This shouldn't fix those :P.

PR #339, opened 2020-01-13
(abandoned)

eordano:
wip: report passport updates to the renderer


eordano:
#361 makes this obsolete

PR #346, opened 2020-01-14
fix: more messaging manager micro-optimizations

BrianAmadori:
This should shave 6ms+ from `ProcessMessages()` CPU overhead.


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingBus.cs
```diff
@@ -87,8 +87,8 @@ public class QueuedSceneMessage_Scene : QueuedSceneMessage
         public IMessageHandler handler;
 
         public LinkedList<QueuedSceneMessage> pendingMessages = new LinkedList<QueuedSceneMessage>();
-        public bool hasPendingMessages => pendingMessages != null && pendingMessages.Count > 0;
-        public int pendingMessagesCount => pendingMessages != null ? pendingMessages.Count : 0;
+        public bool hasPendingMessages => pendingMessagesCount > 0;
+        public int pendingMessagesCount;
```
eordano, line 92: Isn't this a dangerous change? In the sense that somebody might add something to the `pendingMessages` queue without noticing that they have to update this count?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingBus.cs
```diff
@@ -108,10 +108,7 @@ public class QueuedSceneMessage_Scene : QueuedSceneMessage
         System.Text.StringBuilder stringBuilder = new System.Text.StringBuilder();
         public int unreliableMessagesReplaced = 0;
 
-        public bool isRunning
-        {
-            get; private set;
-        }
+        public bool isRunning;
```
eordano: Can we find a better name for this variable? Or document its behavior?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
```diff
@@ -9,12 +9,16 @@ public class MessagingControllersManager : Singleton<MessagingControllersManager
     {
         public static bool VERBOSE = false;
 
-        private const float GLOBAL_MAX_MSG_BUDGET = 0.016f;
+        private const float MAX_MSG_BUDGET = 0.010f;
+        private const float MAX_MSG_BUDGET_FOR_FAR_SCENES = 0.003f;
```
eordano: I think we run into a risk here -- what if this budget is never enough, and the scene lags behind, while we still had the time in frames to avoid that clogging?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
```diff
@@ -205,175 +218,123 @@ public void UpdateThrottling()
...
             while (true)
             {
-                prevTimeBudget = INIT_MSG_BUS_BUDGET_MAX;
+                yield return null;
+
+                timeBudgetCounter = RenderingController.i.renderingEnabled ? MAX_MSG_BUDGET : float.MaxValue;
                 start = Time.unscaledTime;
 
-                bool processedBus = false;
+                bool uiSceneControllerActive = uiSceneController != null && uiSceneController.enabled;
```
eordano: Same as before -- I'm worried that these optimizations would lead to inconsistent state. Can we find a way to cache this in a safer way?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
BrianAmadori: The 3 ms budget is only for the systems of the far scenes. The systems buses use mostly unrealiable messages, so the queue would not explode. I'm going to rename this const to be more clear.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingBus.cs
BrianAmadori: This is a performance concession. This optimization is needed because `LinkedList.Count` has shown to be slow in the critical loop. (around 1ms per frame). 

When I reach the cleanup stage of this PR i'm going to review this so the maintenance risks are minimal.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingBus.cs
BrianAmadori: Let me check what i can do!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
BrianAmadori: What's the risk are you seeing here?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingBus.cs
eordano: Can we use LinkedList.Any?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
eordano: My bad. I didn't realize this is in the main loop. My apologies

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingBus.cs
BrianAmadori: I refuse to use Linq in critical code

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingControllersManager.cs
BrianAmadori: Ended removing it

eordano:
Works on my machine ©

Please fix the tests

eordano (approved):

unity-client/Assets/Scripts/MainScripts/Debugging/WSS/WSSController.cs
```diff
@@ -348,34 +348,34 @@ void Update()
...
-                                HUDController.i.ConfigureNotificationHUD(msg.payload);
+                                HUDController.i?.ConfigureNotificationHUD(msg.payload);
                                 break;
                             case "ConfigureAvatarEditorHUD":
-                                HUDController.i.ConfigureAvatarEditorHUD(msg.payload);
+                                HUDController.i?.ConfigureAvatarEditorHUD(msg.payload);
                                 break;
                             case "UpdateMinimapSceneInformation":
-                                MinimapMetadataController.i.UpdateMinimapSceneInformation(msg.payload);
+                                MinimapMetadataController.i?.UpdateMinimapSceneInformation(msg.payload);
```
AjimenezDCL, line 376: Why is this null check needed?

AjimenezDCL (approved):

unity-client/Assets/Scripts/MainScripts/Debugging/WSS/WSSController.cs
BrianAmadori: I needed to be able to remove the HUD stuff and have the scene still working for performance testing purposes. I added the scene `InitialSceneWithoutUI` with all the UI stuff removed.

PR #348, opened 2020-01-14
feat: more performant scene sort

eordano:


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -553,7 +573,7 @@ private string DecodeAndEnqueue(string payload)
 
         private IEnumerator DeferredDecoding()
         {
-            float lastTimeDecoded = Time.unscaledTime;
+            float lastTimeDecoded = Time.realtimeSinceStartup;
```
BrianAmadori, line 576: `Time.unscaledTime` is better here as its more performant in wasm side. 

When you are sure you are at the beginning of a frame `Time.unscaledTime` is better. When you have to measure time inside the same frame `Time.realtimeSinceStartup` should be used.

BrianAmadori (commented):
Small comment, still have to check the rest of the PR.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
eordano: Ok. I am always defaulting to `realtimeSinceStartup` because I'm not confident enough to know what are the "signals that we're at the beginning of a frame". Because of that I saw it as unreliable :( In general, I try to avoid any assumption required for correctness -- and if there are any invariants assumed, try to enforce them or document them.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -129,9 +130,18 @@ void Awake()
 
         bool positionDirty = true;
 
-        private void SetPositionDirty(DCLCharacterPosition obj)
+        private const int LOCAL_MORDOR = -123456789;
```
pravusjif: Why can't we use `DCL.Configuration.EnvironmentSettings.MORDOR`?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -169,31 +179,51 @@ private void Update()
         {
             InputController_Legacy.i.Update();
 
-            PrioritizeMessageControllerList();
+            AdjustStateToNewPosition();
```
pravusjif: I think we could have a clearer name here as "AdjustScenesSortedByDistance" or "AdjustDistanceSortedScenesList" or similar, what do you think? do you have any other clearer name for what this method does?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -169,31 +179,51 @@ private void Update()
...
+                    {
+                        currentSceneId = currentScene.sceneData.id;
+                    }
+                }
                 OnSortScenes?.Invoke();
             }
         }
 
-        public event System.Action OnSortScenes;
+        public event Action OnSortScenes;
```
pravusjif: since we are modifying this file, can we move these "scattered variables" to the top with the rest? thanks!

pravusjif (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
eordano: Yeah, I used "state" because it also updates the position cache... maybe this is doing too much, taking a look

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
eordano: Because it's an integer. changing it...

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
eordano: Fixed!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
eordano: Fixed!

pravusjif (approved):

PR #350, opened 2020-01-15
fix: performance improvement for ObservableComponents

eordano:


kernel/packages/decentraland-ecs/src/decentraland/Implementation.ts
```diff
@@ -92,7 +97,7 @@ export class DecentralandSynchronizationSystem implements ISystem {
         const component = entity.components[componentName]
         const classId = getComponentClassId(component)
 
-        if (classId !== null) {
+        if (classId !== null && componentNameRE.test(getComponentName(component))) {
```
BrianAmadori, line 101: Regex tests tend to be fairly slow. Are you sure this is a performance improvement? How often this path is reached?

eordano: Regexps are fairly optimized in javascript. This avoids doing any sort of dirty comparison for components that are not the engine (any custom component is not checked). I'll do an approach with `.startsWith` to see how it looks like

eordano: image

BrianAmadori: Nice to know 👍

kernel/packages/decentraland-ecs/src/decentraland/Implementation.ts
```diff
@@ -203,15 +221,21 @@ export class DecentralandSynchronizationSystem implements ISystem {
    * created component is fired immediatly after.
    */
   private disposableComponentCreated(event: DisposableComponentCreated) {
-    this.dcl.componentCreated(event.componentId, event.componentName, event.classId)
+    if (componentNameRE.test(event.componentName)) {
```
BrianAmadori, line 221: Maybe we can use early exit in those functions instead to reduce indentation.

BrianAmadori (approved):

moliva (approved):
LGTM.

PR #357, opened 2020-01-15
chore: add isPointerBlocker to shapes

pravusjif:
* Added new isPointerBlocker property to shapes
* Improved Configuration layers
* Updated blocking pointer events test scene (0.20)

--------------------------------------------------------------
Since this is a new property, it's not used anywhere in the deployed world, so there's only the unit test and the local test scene (public/test-scenes/0.20.objects-blocking-click-for-objects-behind-them) in which the low-wall has the `isPointerBlocker` in false, so you can click the chest behind it.


kernel/packages/decentraland-ecs/src/decentraland/Components.ts
```diff
@@ -154,6 +154,12 @@ export class Shape extends ObservableComponent {
   @ObservableComponent.field
   withCollisions: boolean = true
 
+  /**
+   * Set to true to turn on the PointerEvents blocking for the entity.
+   */
+  @ObservableComponent.field
+  isPointerBlocker: boolean = true
```
eordano, line 161: hmm, by default, be careful about this. all "default"s should be "false". May I suggest `transparentForPointerCollisions` or something like that?

nearnshaw: hmm `transparentForPointerCollisions` is too long
the nice thing about `isPointerBlocker` is that it's the same name we use for a similar property in UI components
I'd vote that consistency w that > consistency of all default as false

eordano: I know, but this is code that is already deployed.

So, the default is `true` for EVERYTHING that is already deployed (because it was done with a previous version of the SDK).

See my point?

eordano: A workaround is to differentiate between "undefined" and "false" in Unity.

pravusjif: but the `isPointerBlocker` property is new, it hasn't been deployed anywhere yet, Tibi, I'm missing your point on that.

Regarding the naming of this new property I also think `isPointerBlocker` is better as it is consistent with the naming of the `isPointerBlocker` in the UIShapes (they don't inherit from the BaseShape, they are separate, in case you referred to this when you mentioned the already-deployed code) and as we have the `withCollisions` and `visible` in true by default I don't see the problem in having the new `isPointerBlocker` in true by default as well

pbosio (approved):

kernel/packages/decentraland-ecs/src/decentraland/Components.ts
BrianAmadori: It'd be nice to have all the defaults as `false` but in this case sounds unrealistic to me. Agree with @pravusjif .

eordano: All the code deployed thus far, that has a Shape, will send to unity a Component payload with this field missing, regardless of the value we set here as a default.

Because it's a new property, unity when decoding this will see it as "null" (or whatever the Json parser decides)

My point is: be careful with handling this null/missing default, specially because it's gonna be "falsy".

pbosio: AFAIK when Unity parse a JSON if a field is missing in that JSON it won't modify the value of the field, so it will keep the value set as default in the C# class definition

eordano:
Awesome, thanks!

kernel/packages/decentraland-ecs/src/decentraland/Components.ts
pravusjif: Yes, keep in mind we have the same model structures in unity's side, if there is a missing field it will use the default we have on that side (and those should always match the defaults in kernel). So there will be no problem, and if there were i would've detected it on my manual tests with already-existent shapes when developing the feature

unity-client/Assets/Scripts/MainScripts/DCL/Components/BaseShape.cs
```diff
@@ -85,10 +86,22 @@ public static void ConfigureVisibility(GameObject meshGameObject, bool shouldBeV
...
                 }
             }
         }
+
+        protected int CalculateCollidersLayer(Model model)
+        {
+            // We can't enable this layer changer logic until we redeploy all the builder and street scenes with the corrected 'withCollision' default in true...
+            /* if (!model.withCollisions && model.isPointerBlocker)
+                return PhysicsLayers.onPointerEventLayer;
+            else */
```
BrianAmadori, line 100: We should remove these commented lines.

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/BaseShape.cs
pravusjif: please check the comment above those lines, we should ship it that way until we re-deploy all the builder scenes with the correct default valures of the GLTFShape (the withCollisions is the one we need updated)

BrianAmadori: Oh so its part of the comment? It didn't seem like it.

PR #364, opened 2020-01-16
fix: SceneController deferred decoding performance spike

BrianAmadori:
Budget was too high and allowed a maximum of 100 ms spikes. I confirmed 60ms spikes in some situations of walking around in the world.

Method logic was simplified too.


eordano:
😍

eordano (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -580,27 +577,24 @@ private string DecodeAndEnqueue(string payload)
...
-                else
-                {
-                    yield return null;
-                    lastTimeDecoded = Time.unscaledTime;
+                    if (Time.realtimeSinceStartup - lastTimeDecoded < maxTimeForDecode)
+                        continue;
                 }
+
+                yield return null;
+                lastTimeDecoded = Time.unscaledTime;
```
eordano: wouldn't it be better to put this on top of the while loop and call it `start`?

pbosio (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -579,28 +576,25 @@ private string DecodeAndEnqueue(string payload)
...
-                else
-                {
-                    yield return null;
-                    lastTimeDecoded = Time.unscaledTime;
+                    if (Time.realtimeSinceStartup - start < maxTimeForDecode)
+                        continue;
                 }
+
+                yield return null;
+                start = Time.unscaledTime;
```
pravusjif, line 597: Not sure why we use unscaledTime here instead of realtimeSinceStartup.
Also, wouldn't it be better to put this (or `start = Time.realtimeSinceStartup;`) at the beginning of the `while()` as Tibi suggested?

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
BrianAmadori: We use `unscaledTime` because is a bit faster. The strategy is to use `unscaledTime` as soon as the frame starts and then use `realtimeSinceStartup` later. Since `unscaledTime` is only refreshed once per frame and is useless for measuring in-frame time later.

PR #386, opened 2020-01-20
fix: messaging controllers lazy sorting issue

BrianAmadori:
- fix: GLTF budget was not calculated properly
- fix: replaced all the `GetCurrentScene` calls by the more performant `SceneController.curentSceneId` 
- fix: buses are sorted after `AddController` so the renderer won't have to wait for the player to move to process the messages
- removed the no more used `ThrottlingController` and its logic remnants.
- scene sorting method micro-optimizations according to deep profiling measurements.


eordano (approved):

eordano:
Looks a lot better imho

pbosio (approved):

pravusjif (approved):

PR #407, opened 2020-01-22
fix: parcel blocker instancing perf spike

BrianAmadori:
For big scenes (e.g. plazas) we are instancing too many green blockers. This should cut the number of blockers by a fair amount.

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
@@ -1,4 +1,4 @@
-using DCL.Components;
+﻿using DCL.Components;
 using DCL.Configuration;
 using DCL.Helpers;
 using DCL.Models;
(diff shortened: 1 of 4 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace DCL.Controllers
 {
-    public class ParcelScene : MonoBehaviour, ICleanable
+    public partial class ParcelScene : MonoBehaviour, ICleanable
```
pravusjif: Why are we using the partial keyword?

pravusjif (approved):

pbosio (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
BrianAmadori: Oopsie. I moved a class to a new file and the VS auto-refactor added this. Going to remove it.

PR #408, opened 2020-01-22
fix: limit systems fps with distance

BrianAmadori:
As we deactivated the decode throttling in a previous commit, the decoding perf. hit due to enormous messaging throughtput that takes place in some parts of the world is easily noticeable. 

With this we also will reduce greatly the lossy messages being replaced. As they aren't processeed but aren't free either.

Better tested in ZONE: 63,-67
https://explorer.decentraland.zone/branch/fix/limit-systems-fps-with-distance/index.html?position=63,-67


pravusjif (approved):
Looks good to me (and compared to zone I see some improvement) but this should be checked by someone with more Kernel insight

moliva (approved):

kernel/packages/shared/world/parcelSceneManager.ts
```diff
@@ -86,7 +86,8 @@ export async function enableParcelSceneLoading(options: EnableParcelSceneLoading
     globalSignalSceneLoad(sceneId)
     // create the worker if don't exist
     if (!getSceneWorkerBySceneID(sceneId)) {
-      const parcelScene = new options.parcelSceneClass(ILandToLoadableParcelScene(parcelSceneToStart))
+      let parcelScene = new options.parcelSceneClass(ILandToLoadableParcelScene(parcelSceneToStart))
```
moliva: nitpick: this could still be a `const` as the reference remains unchanged

PR #417, opened 2020-01-23
fix: remove MonoBehaviour inheritance from LoadWrapper classes

BrianAmadori:
This is the first step in reducing the overhead of `AddComponent` calls.

- All the `LoadWrapper` instances are stored in the `LoadableShape.attachedLoaders` dictionary and retrieved using `GetLoaderForEntity`. Having a static dictionary in the base class with `gameObject` as key, means than any given entity can only have one `LoadWrapper` attached.

- As all `LoadWrapper` classes are not `MonoBehaviour` anymore, the `OnDestroy` method was removed. This should not be an issue because we are already ensuring `Unload` gets called before destroying any entity through the `DetachShape` calls.

- I'm still on the fence of removing the need for having to create a new `meshRootGameObject` instance. As all the GLTF/AB we are loading already have a container object.


unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadableShape.cs
```diff
@@ -109,7 +119,19 @@ protected virtual void AttachShape(DecentralandEntity entity)
...
+
+                if (!attachedLoaders.ContainsKey(entity.meshRootGameObject))
+                {
+                    loadableShape = new LoadWrapperType();
+                    attachedLoaders.Add(entity.meshRootGameObject, loadableShape);
+                }
+                else
+                {
+                    loadableShape = GetLoaderForEntity(entity) as LoadWrapperType;
+                }
```
pravusjif: Can't we do something like the following here?
```
LoadWrapperType loadableShape = GetLoaderForEntity(entity) as LoadWrapperType;

if (loadableShape == null)
{
    loadableShape = new LoadWrapperType();
    attachedLoaders.Add(entity.meshRootGameObject, loadableShape);
}
```

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShape.cs
```diff
@@ -33,12 +34,23 @@ protected override void AttachShape(DecentralandEntity entity)
...
+
+                if (!attachedLoaders.ContainsKey(entity.meshRootGameObject))
+                {
+                    loadableShape = new LoadWrapper_NFT();
+                    attachedLoaders.Add(entity.meshRootGameObject, loadableShape);
+                }
+                else
+                {
+                    loadableShape = GetLoaderForEntity(entity) as LoadWrapper_NFT;
+                }
```
pravusjif: same question as in LoadableShape.cs here

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_NFT.cs
```diff
@@ -23,6 +23,14 @@ public override void Load(string src, System.Action<LoadWrapper> OnSuccess, Syst
         {
             if (string.IsNullOrEmpty(src)) return;
 
+            if (loaderController == null)
+                loaderController = entity.meshRootGameObject.GetComponent<NFTShapeLoaderController>();
+
+            loaderController.collider.enabled = withCollisions;
+            loaderController.backgroundColor = backgroundColor;
```
pravusjif, line 36: I'm confused about this 2 values, shouldn't this be configured from the shape class itself using the model parsed from the kernel?

pravusjif (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadableShape.cs
AjimenezDCL: Haven't check the PR yet, but in any case you can avoid the null check of @pravusjif suggestion with:
```
if (GetLoaderForEntity(entity) is LoadWrapperType loadableShape)
{
    loadableShape = new LoadWrapperType();
    attachedLoaders.Add(entity.meshRootGameObject, loadableShape);
}
```

EDIT: This is nonsense, ignore it

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_NFT.cs
```diff
@@ -23,6 +23,14 @@ public override void Load(string src, System.Action<LoadWrapper> OnSuccess, Syst
         {
             if (string.IsNullOrEmpty(src)) return;
 
+            if (loaderController == null)
+                loaderController = entity.meshRootGameObject.GetComponent<NFTShapeLoaderController>();
+
+            loaderController.collider.enabled = withCollisions;
```
AjimenezDCL, line 35: Do we need a null check here? Could the `meshRootGameObject` not have the component?

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_OBJ.cs
```diff
@@ -1,52 +1,47 @@
...
+            if (string.IsNullOrEmpty(src))
+                return;
+
+            if (objLoaderComponent == null)
+                objLoaderComponent = entity.meshRootGameObject.GetOrCreateComponent<DynamicOBJLoaderController>();
+
+            objLoaderComponent.OnFinishedLoadingAsset += CallOnComponentUpdated;
+
+            alreadyLoaded = false;
+            objLoaderComponent.OnFinishedLoadingAsset += () => OnSuccess(this);
```
AjimenezDCL: I see this callback being subscribed to `objLoaderComponent.OnFinishedLoadingAsset` but not being unsubscribed. Are we controlling that use-case somewhere else?

Otherwise we could have a delegate as a field in the class, add `OnSuccess(this)` to the delegate and the delegate to `objLoaderComponent.OnFinishedLoadingAsset`.

AjimenezDCL (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadableShape.cs
pravusjif: what kind of magic is that? XD.
Also, that code block shoud run when the GetLoaderForEntity(entity) **is not** LoadWrapperType, would `if(!(GetLoaderForEntity(entity) is LoadWrapperType) loadableShape)` work? I don't know if this convinces me as it is kind of confusing XD

AjimenezDCL: Oh... you are right.

I thought the condition was `if (loadableShape != null)`. My fault!

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_NFT.cs
BrianAmadori: Yes, the shape class pass the values from the model to these two values. Had to make the params pass around this way for scoping purposes.

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_OBJ.cs
BrianAmadori: Let me check. The `OBJ` stuff is pretty outdated and should be removed, but in the meantime i'll try to apply a workaround.

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadableShape.cs
BrianAmadori: Cool suggestions guys. The `loadableShape` must be stored, so I think the inline `is` will not work. We should use that new C# idiom more often though. Let me check what I can do with this expression.

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_NFT.cs
BrianAmadori: Maybe. Let me check.

BrianAmadori: This component is added here:
https://github.com/decentraland/explorer/blob/fac3bdacd71d925433594d9ad4a43a0ea7707f14/unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShape.cs#L26
Just before calling this code. That prefab contains the `NFTShapeLoaderController` component. Not sure of a better way of doing this.

AjimenezDCL: Maybe we can somehow abort the execution safely if the component is not found. To avoid crashes.

BrianAmadori: ok, added an early exit just in case.

pravusjif (approved):

PR #420, opened 2020-01-24
V2 comms protobuf

pablitar:
# What? 
Using protobuf for message payloads in v2 comms

# Why? 
It greatly reduces bandwidth usage


moliva (approved):
Nice! 😬

PR #422, opened 2020-01-24
fix: remove MonoBehaviour inheritance from PoolableObject

BrianAmadori:


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
```diff
@@ -98,8 +98,8 @@ protected IEnumerator LoadAssetBundleWithDeps(string baseUrl, string hash, Actio
                 }
             }
 
-            if (concurrentRequests >= 25)
-                yield return new WaitUntil(() => concurrentRequests < 25);
+            if (concurrentRequests >= 20)
```
eordano: can we extract this as a variable?

eordano (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PoolManager/PoolManager.cs
```diff
@@ -192,5 +229,19 @@ public void ReleaseAllFromPool(object id)
...
+            foreach (var kvp in poolables)
+            {
+                if (kvp.Value.gameObject == null)
+                {
+                    kvp.Value.node.List.Remove(kvp.Value);
+                    kvp.Value.node = null;
+                }
+            }
+
+            poolables = poolables.Where((kvp) => kvp.Value.gameObject != null).ToDictionary(p => p.Key, p => p.Value);
```
AjimenezDCL: I think we can optimize this method.

```
var toRemove = List<GameObject>();
foreach (var kvp in poolables)
{
        if (kvp.Value.gameObject == null)
        {
                kvp.Value.node.List.Remove(kvp.Value);
                kvp.Value.node = null;
        }
        else
        {
                toRemove.Add(kvp.Key);
        }
}

foreach(key in toRemove) poolables.Remove(key);
```

Instead of iterate three times (and create big additional structures through LinQ) we iterate poolables once + `toRemove.Count`

AjimenezDCL (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PoolManager/PoolManager.cs
BrianAmadori: This is me being lazy, thanks for the snippet 💃

pravusjif (changes requested):
The code looks OK, but testing the build, I see that after running around for a while loading and unloading scenes, there comes a point in which a nullreference ecxception appears in the browser console and the scenes stop loading, so it's critical.
I tested running around here https://explorer.decentraland.zone/branch/fix/remove-poolableobject-monobehaviour/index.html?position=-2%2C14
going to the north for around 30 secs, this doesn't happen in .zone so it sounds like a bug in this branch

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PoolManager/PoolManager.cs
BrianAmadori: Ok, I remembered the reason I didn't implemented that way in the first place. It was because `GameObject` references get null when a `GameObject` is destroyed. So i thought the dictionary would end with a bunch of hashcode ints referring to no more existent `GameObjects`, and thus they would have to be removed by value. 

This means that the `toRemove` list would get filled up with `null` references and removing by key shouldn't work.

Turns out that the approach *does* work. But I'm not sure of the sorcery that's happening under the hood to allow it. Maybe it has to do something with `==` operator being overriden for `UnityEngine` objects.

AjimenezDCL: What the hell... o_o

Then I leave it to your judgment, whatever it feels safer.

eordano:
Can confirm the bug reported by @pravusjif

it looks like this (going for more than one minute now, blockers never stop)
 
image

BrianAmadori:
I couldn't open the link due to "peer connection couldn't be established" and in local couldn't repro. Just merged from master because this branch was pretty outdated to see if that's the issue.

eordano:
lets merge this asap (confirmed fixed).


Can we raise the `20` back to `25`? It's a pretty noticeable slowdown when walking around builder scenes that have many objects around.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PoolManager/PoolManager.cs
AjimenezDCL: I hope I'am not too late. What about this?

```
Dictionary<GameObject, PoolableObject> newPoolables = new Dictionary<GameObject, PoolableObject>();

foreach (var kvp in poolables)
{
        if (kvp.Value.gameObject == null)
        {
                kvp.Value.node.List.Remove(kvp.Value);
                kvp.Value.node = null;
	        newPoolables.Add(kvp.Key, kvp.Value);
        }
}

poolables = newPoolables;
```

Faster and no unity's sorcery involved

pravusjif (approved):
the previous critical bug now seems fixed.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PoolManager/PoolManager.cs
BrianAmadori: Already resolved it this way without alloc:

https://github.com/decentraland/explorer/blob/889791af772fb9f7babad6d65ee6e1946d32adc4/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PoolManager/PoolManager.cs#L250

Unity sorcery is involved, but it works.

AjimenezDCL (approved):

PR #423, opened 2020-01-24
fix: reported current scene id

pravusjif:
Since the scene sorting is not accuarate based on grid coords (it sorts based on **parcels**, not **scenes**), we go back to traversing the scenes collection (at most with our current LOS for loaded scenes it would be like 25~ items, not much at all)

------------------------------------
You can test this fixes the door opening in the Space Office scene at **ZONE 56.-59**, using the keycard at the back of the building


BrianAmadori (approved):

BrianAmadori:
Actually I think it would traverse 2 or 3 scenes at most, because we are iterating by the ordered list. The current scene should be in the first few items most of the times.

PR #439, opened 2020-01-27
fix: avatar and model visibility fixes

BrianAmadori:
fix: avatar incomplete models were being shown when it shouldn't due to `SetVisibility` incorrectly overriding the renderer enabled states.

fix: some models showing in 0,0,0 of the world for a frame when it shouldn't 

fix: rewrote `BeHiddenUntilWholeAvatarIsReady` as wasn't asserting the entire avatar, just a wearable.

fix: `AssetPromise_GLTF` wasn't setting visible = false correctly leading to logic errors.

fix: Asset bundle promises weren't setting Invisibility correctly while the asset is being loaded.

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
@@ -20,8 +20,6 @@ namespace DCL
         Action OnSuccessCallback;
         Action OnFailCallback;
 
-        private bool visibility = true;
-        
         internal BodyShapeController bodyShapeController;
         internal Dictionary<string, WearableController> wearablesController = new Dictionary<string, WearableController>();
         internal FacialFeatureController eyesController;
@@ -30,7 +28,7 @@ namespace DCL
 
         internal bool isLoading = false;
 
-        public void ApplyModel (AvatarModel model, Action onSuccess, Action onFail)
+        public void ApplyModel(AvatarModel model, Action onSuccess, Action onFail)
         {
             this.model = model;
             this.OnSuccessCallback = onSuccess;
@@ -82,7 +80,7 @@ namespace DCL
             }
         }
 
-        private IEnumerator LoadAvatar ()
+        private IEnumerator LoadAvatar()
         {
             if (string.IsNullOrEmpty(model.bodyShape))
             {
@@ -101,7 +99,7 @@ namespace DCL
             {
                 HideAll();
 
-                bodyShapeController = new BodyShapeController( ResolveWearable(model.bodyShape));
+                bodyShapeController = new BodyShapeController(ResolveWearable(model.bodyShape));
                 SetupDefaultFacialFeatures(bodyShapeController.bodyShapeType);
                 bodyShapeController.Load(transform, OnWearableLoadingSuccess, OnWearableLoadingFail);
             }
@@ -152,7 +150,9 @@ namespace DCL
                 bodyShapeController.SetupMouth(mouthMaterial, mainTexture, model.skinColor);
             }));
 
-            yield return new WaitUntil( () => eyesReady && eyebrowsReady && mouthReady);
+            yield return new WaitUntil(() => eyesReady && eyebrowsReady && mouthReady);
+
+            isLoading = false;
 
             SetupAnimator();
 
@@ -163,7 +163,6 @@ namespace DCL
             yield return null;
             ResolveVisibility();
 
-            isLoading = false;
             OnSuccessCallback?.Invoke();
         }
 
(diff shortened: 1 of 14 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
```diff
@@ -46,13 +48,31 @@ public void ApplyModel (AvatarModel model, Action onSuccess, Action onFail)
             }
 
             isLoading = true;
-            StartCoroutine(LoadAvatar());
+            loadCoroutine = CoroutineStarter.Start(LoadAvatar());
```
AjimenezDCL, line 51: Not a big deal but being `AvatarRenderer` a `Monobehaviour` cannot we avoid using `CoroutineStarter`?

AjimenezDCL: Nevermind, it would break the logic in `SetVisibility` because it would stop the coroutines. All good!

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
```diff
@@ -310,6 +337,7 @@ private void HideAll()
 
         protected virtual void OnDestroy()
         {
+            CoroutineStarter.Stop(loadCoroutine);
```
AjimenezDCL: Shouldn't we `Stop` the facial features coroutines as well?

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
```diff
@@ -9,7 +9,7 @@ public class AvatarShape : BaseComponent
         public AvatarName avatarName;
         public AvatarRenderer avatarRenderer;
         public AvatarMovementController avatarMovementController;
-        [SerializeField] private GameObject minimapRepresentation;
+        public GameObject minimapRepresentation;
```
AjimenezDCL: Small change, but can we use `[SerializeField] internal GameObject minimapRepresentation` instead? I believe we already have an `AssemblyInfo` setting _internals as visible_ for the test assembly.

AjimenezDCL (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/DCLCharacterController.cs
```diff
@@ -410,7 +410,7 @@ void ReportMovement()
         float height = 0.875f;
 
         var reportPosition = characterPosition.worldPosition + (Vector3.up * height);
-        var compositeRotation = Quaternion.LookRotation(cameraForward.Get());
+        var compositeRotation = cameraForward != Vector3.zero ? Quaternion.LookRotation(cameraForward.Get()) : Quaternion.identity;
```
pravusjif: Do we gain anything from this change? most of the time the camera forward won't be Vector3.zero, and probably never XD, since it will alway's have at least minor rotations right?

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/DCLCharacterController.cs
pravusjif: or you meant to check the camera at its initial state? (in which case we could also check the "renderingEnabled" bool i guess, and use the Quaternion.identity there)

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
BrianAmadori: Exactly :)

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/DCLCharacterController.cs
BrianAmadori: This was just to prevent an annoying Unity warning displayed constantly when `Quaternion` is generated from a `Vector3.zero` that was keeping me to see the proper logs in the console. Maybe is better to leave it in an `UNITY_EDITOR` flag.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
BrianAmadori: Good catch!

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
BrianAmadori: Yeah, I'll try to apply this change.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/DCLCharacterController.cs
BrianAmadori: Ok, just removed this change and fixed the warnings setting `0,0,1` as the default value of 
`cameraForward.asset`. This way the console keeps clean and we don't risk failing compilation due to the flag.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
BrianAmadori: Done

PR #444, opened 2020-01-29
feat: limit chat messages

eordano:


kernel/packages/ui/avatar/chatWindow.ts
```diff
@@ -198,5 +199,9 @@ async function sendMsg(messageToSend: string) {
 
 function addMessage(messageEntry: MessageEntry): void {
   internalState.messages.push(messageEntry)
+  const length = internalState.messages.length
+  if (length > MAX_CHAT_MESSAGES) {
+    internalState.messages = internalState.messages.slice(length - MAX_CHAT_MESSAGES)
+  }
   createLogMessage(messagesLogStackContainer, messageEntry)
```
moliva: We should be removing the message components from the UI as well, might need to keep track of these.

eordano: oh right this is not React 😓  haha

kernel/packages/unity-interface/dcl.ts
```diff
@@ -371,6 +371,40 @@ export const unityInterface = {
       gameInstance.SendMessage('SceneController', 'UpdateMinimapSceneInformation', JSON.stringify(chunk))
     }
   },
+  TriggerAirdropDisplay() {
```
BrianAmadori: I think this should be out of this PR.

BrianAmadori (dismissed):

BrianAmadori:
The logic works OK, however you can see how the messages processing starts moving messages up one by one in the world. That's can get annoying if a lot of messages are coming in.

PR #447, opened 2020-01-30
feat: tutorial unity-side

pbosio:


unity-client/Assets/Tutorial/Scripts/TutorialStages/GenesisPlazaStageController.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        base.OnStageStart();
+
+        // NOTE: we should probably remove this toast when proper system for wellcome toast is created
+        ShowWellcomeToast();
+
+        TutorialController.i?.SetRunningStageFinished();
+    }
+
+    private void ShowWellcomeToast()
```
BrianAmadori:
```suggestion
    private void ShowWelcomeToast()
```

unity-client/Assets/Tutorial/Scripts/TutorialStages/GenesisPlazaStageController.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        string notificationText = $"Wellcome, {UserProfile.GetOwnUserProfile().userName}!";
```
BrianAmadori:
```suggestion
        string notificationText = $"Welcome, {UserProfile.GetOwnUserProfile().userName}!";
```

unity-client/Assets/Tutorial/Scripts/TutorialStages/GenesisPlazaStageController.cs
```diff
@@ -0,0 +1,35 @@
...
+        Vector2Int currentCoords = CommonScriptableObjects.playerCoords.Get();
+        string parcelName = MinimapMetadata.GetMetadata().GetTile(currentCoords.x, currentCoords.y)?.name;
+        if (!string.IsNullOrEmpty(parcelName))
+        {
+            notificationText += $" You are in {parcelName} {currentCoords.x}, {currentCoords.y}";
```
BrianAmadori: Maybe this would get more clean (and easier to localize later) if we extract those literals to variables.

unity-client/Assets/Tutorial/Scripts/TutorialStages/GenesisPlazaStageController.cs
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        NotificationModel model = new NotificationModel()
+        {
+            message = notificationText,
+            scene = "",
+            type = NotificationModel.NotificationType.GENERIC_WITHOUT_BUTTON,
+            timer = 5
```
BrianAmadori: This `5` is better as a constant?

unity-client/Assets/Tutorial/Scripts/TutorialTooltip.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            return;
+        }
+
+        isShowing = false;
+        if (animator && hideAnimationClip)
+        {
+            animator.clip = hideAnimationClip;
+            animator.Play();
+            StartCoroutine(HideAfterAnimation(hideAnimationClip.length));
```
BrianAmadori: We should be using animation events instead of this approach. Way cleaner and more performant.

unity-client/Assets/Tutorial/Scripts/TutorialTooltip.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System.Collections;
+using UnityEngine;
+using UnityEngine.EventSystems;
+
+public class TutorialTooltip : MonoBehaviour, IPointerClickHandler
+{
+    public float secondsOnScreen = 12f;
+
+    [SerializeField] private AnimationClip showAnimationClip = null;
```
BrianAmadori, line 8: I think its better if we use a simple mecanim animation override here. If in the future we want to polish the animation flow it'll get easier to maintain. (e.g. if we want some idle animation or similar). 

Also, this'd allow us to remove most of the animation handling code.

unity-client/Assets/Tutorial/Scripts/TutorialStages/InitialStageController.cs
```diff
@@ -0,0 +1,99 @@
...
+
+    public override void OnStageFinished()
+    {
+        base.OnStageFinished();
+        DCLCharacterController.OnPositionSet -= OnTeleport;
+        DCLCharacterController.OnCharacterMoved -= OnCharacterMove;
+        HUDController.i?.minimapHud.SetVisibility(true);
+    }
+
+    private IEnumerator StageSecuence()
```
BrianAmadori:
```suggestion
    private IEnumerator StageSequence()
```

unity-client/Assets/Tutorial/Scripts/TutorialController.cs
```diff
@@ -0,0 +1,209 @@
...
+
+#if UNITY_EDITOR
+    [Header("Debugging")]
+    public int debugFlagStartingValue = 0;
+    public bool debugRunTutorialOnStart = false;
+    [Space()]
+#endif
+
+    [Header("Stage Controller References")]
+    [SerializeField] TutorialStageController initalStage = null;
```
BrianAmadori: Its more maintainable to use an array here instead of having each state as a individual reference.

unity-client/Assets/Tutorial/Scripts/TutorialController.cs
```diff
@@ -0,0 +1,209 @@
...
+    public void SetRunningStageFinished()
+    {
+        tutorialFlagMask |= (int)runningStage.flag;
+        runningStage.OnStageFinished();
+        UserProfile.GetOwnUserProfile().SetTutorialFlag(tutorialFlagMask);
+        // TODO: send update to user profile
+        StartTutorialStageIfNeeded(tutorialFlagMask);
+    }
+
+    private void StartTutorialStageIfNeeded(int tutorialFlagMask)
```
BrianAmadori: Couldn't we just make this an `IEnumerator` and `yield` all the stages out? This could save the `StartCoroutine` boilerplate inside each `StageController`.

unity-client/Assets/Tutorial/Scripts/TutorialController.cs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    private int GetTutorialFlagFromUserProfile()
+    {
+        return UserProfile.GetOwnUserProfile().tutorialFlag;
+    }
+
+    private void OnRenderingStateChanged(bool renderingEnabled)
+    {
+        if (renderingEnabled && isTutorialEnabled)
```
BrianAmadori: We can invert this `if` statement and early exit to avoid nesting.

BrianAmadori (changes requested):

unity-client/Assets/Tutorial/Scripts/TutorialStages/GenesisPlazaStageController.cs
pravusjif: Let's do this when we prepare everything for localization in the future

unity-client/Assets/Tutorial/Scripts/TutorialStages/GenesisPlazaStageController.cs
pravusjif: Done

unity-client/Assets/Tutorial/Scripts/TutorialStages/GenesisPlazaStageController.cs
pravusjif: Done

unity-client/Assets/Tutorial/Scripts/TutorialStages/GenesisPlazaStageController.cs
pravusjif: Done

unity-client/Assets/Tutorial/Scripts/TutorialTooltip.cs
pravusjif: Done

unity-client/Assets/Tutorial/Scripts/TutorialTooltip.cs
pravusjif: We are not using mecanim, we are using legacy Animation component. You suggest we make these animations non-legacy (by toggling the bool in unity debug mode) and use an animator with 2 states instead?

unity-client/Assets/Tutorial/Scripts/TutorialStages/InitialStageController.cs
pravusjif: Done

unity-client/Assets/Tutorial/Scripts/TutorialController.cs
pravusjif: Done

unity-client/Assets/Tutorial/Scripts/TutorialTooltip.cs
BrianAmadori: exactly

kernel/packages/config/index.ts
```diff
@@ -117,6 +117,7 @@ export const DISABLE_AUTH = location.search.indexOf('DISABLE_AUTH') !== -1 || DE
 export const ENGINE_DEBUG_PANEL = location.search.indexOf('ENGINE_DEBUG_PANEL') !== -1
 export const SCENE_DEBUG_PANEL = location.search.indexOf('SCENE_DEBUG_PANEL') !== -1 && !ENGINE_DEBUG_PANEL
 export const SHOW_FPS_COUNTER = location.search.indexOf('SHOW_FPS_COUNTER') !== -1 || DEBUG
+export const TUTORIAL_ENABLED = WORLD_EXPLORER && true
```
AjimenezDCL: Is the `&& true` needed?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/AvatarEditorHUDController.cs
```diff
@@ -23,6 +23,8 @@ public class AvatarEditorHUDController : IDisposable, IHUD
 
     protected AvatarEditorHUDView view;
 
+    public Action<bool> OnVisibilityChanged;
```
AjimenezDCL, line 26: I dont think this is needed.

The only subscriber is `InitialStageController` and the same script is the one changing the visibility of the `AvatarEditorHUD`.

SetVisibility could be called by Kernel through the HUDController, but in that case, I think this event should be in the HUDController so we don't couple the autonomous HUDs to any external system (that is not HUDController).

unity-client/Assets/Tutorial/Scripts/TutorialController.cs
pravusjif: In that case, to avoid having "hardcoded TutorialFlags" on the different array elements based on index in `CreateStagesChainOfResponsibility()` (we'd setup each stage using the array elements in the following way: 
```
private void CreateStagesChainOfResponsibility()
    {
        // 1st stage - initialStage
        firstStage = new GenericSceneStageHandler(TutorialFlags.InitialScene, tutorialStagesArray[0],
            () =>
            {
                HUDController.i?.minimapHud.SetVisibility(false);
            });

        // 2nd stage - LoginRewardAndUserEmail
        var nextStage = firstStage.SetNext(new GenericSceneStageHandler(TutorialFlags.LoginRewardAndUserEmail, tutorialStagesArray[1], null));

        // 3rd stage - ChatAndAvatarExpressions
        nextStage = nextStage.SetNext(new GenericSceneStageHandler(TutorialFlags.ChatAndAvatarExpressions, tutorialStagesArray[2],
            () =>
            {
                HUDController.i?.avatarHud.SetVisibility(false);
                SetChatVisible(false);
                // TODO: hide avatar expressions
                //HUDController.i?.expressionsHud.SetVisibility(false);
            }));
    }
```
) but if some day someone changes the array elements order, it would become a problem) I'd have to create a new mini-class inside TutorialController, with both a reference for the gameobject (as we have now instead of an array, but for each element) and a reference to the `TutorialFlags` corresponding to that stage.

Is that OK with you?

unity-client/Assets/Tutorial/Scripts/TutorialController.cs
pravusjif: Not sure about this, with the previously implemented "chain of command" I should refactor that to get this working with nested coroutines, is it worth the time?

AjimenezDCL (dismissed):

unity-client/Assets/Tutorial/Scripts/TutorialController.cs
BrianAmadori: It wouldn't be more simple to just have each stage define its own flag? (i.e. just initialize it in the stage constructor). If for some reason this isn't viable that approach sounds fine.

unity-client/Assets/Tutorial/Scripts/TutorialController.cs
BrianAmadori: I think we'd reach a far more simpler design, easier to maintain. But if you think it can't be finished by today i'd leave it as it is.

kernel/packages/config/index.ts
BrianAmadori: We should change this to do a `location.search.indexOf('NO_TUTORIAL') === -1`.

Otherwise it'd give headaches to debug explorer in the future.

unity-client/Assets/Scripts/MainScripts/DCL/UserProfile/UserProfile.cs
```diff
@@ -122,9 +117,9 @@ public string[] GetInventoryItemsIds()
         return inventory.Keys.ToArray();
     }
 
-    public void SetTutorialFlag(int tutorialFlag)
+    public void SetTutorialFlag(int newTutorialFlagsMask)
```
BrianAmadori, line 120: Why a bitmask? Why not just a counter?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/AvatarEditorHUDController.cs
BrianAmadori: This is in place because we have to detect somehow when the user closes the avatar hud in order to continue the tutorial, so the check and callback should be in Unity side. Do you have a suggestion to improve this approach with this in mind? (asking honestly because I can't figure out anything from the top of my head).

BrianAmadori (approved):

PR #450, opened 2020-01-30
fix: retina tools not finding build issue

BrianAmadori:


eordano:
Why are we not commiting WebGLRetinaSettings.json?

BrianAmadori:
Closing this because for some reason the canvas doesn't render anything when the plugin applies the patch. We'll have to choose between fixing the plugin (and then applying the patch using another hacky string replacement approach from CI when we go OSS) or to remove it altogether.

PR #452, opened 2020-01-30
testing: tutorial scene

pbosio:
just to check tutorial scene and tooltips


BrianAmadori:
Closed because we are about to merge the unity PR.

PR #454, opened 2020-01-30
config: max visible peers increased

pablitar:
# What? 
Increases maxVisiblePeers size to 50

# Why? 
Because that's the max amount of peers we will have in a layer

### DIFF
--- a/kernel/packages/config/index.ts
+++ b/kernel/packages/config/index.ts
@@ -126,7 +126,7 @@ export namespace commConfigurations {
 
   export const peerTtlMs = 60000
 
-  export const maxVisiblePeers = 25
+  export const maxVisiblePeers = 50
 
   export const iceServers = [
     {


pablitar:
@eordano @BrianAmadori Do you know any reason why this number is 25? Would there be a problem in increasing it to 50?

eordano:
It's likely to have been there from a long time ago. LGTM, if you think we can handle this, although I'd be more thorough with the testing (is latency acceptable? won't we run into problems for certain kinds of experiences?)

eordano (approved):

BrianAmadori:
Before increasing this please take into the account the rendering performance implications too. Lag isn't the only issue. Avatar skinned meshes can't be batched and animation processing isn't cheap either.

pablitar:
Instead of changing the default, maybe we should make this configurable through URL parameters, and we can start testing the impact.

What do you think?

BrianAmadori:
+1. sounds like a good idea.

BrianAmadori:
I'm closing this PR because its cluttering. We can re-open it when we implement the setting to override the peers amount.

PR #458, opened 2020-01-31
feat: tutorial scene integration

pbosio:
You can test the tutorial with 

https://explorer.decentraland.zone/branch/feat/kernel-tutorial-scene/index.html?ENV=org&RESET_TUTORIAL

Notes:
* the URL param `RESET_TUTORIAL` resets the user tutorial step, if you finish the tutorial (the daily reward tip is the last one), then you can re-access the url (**without** the `RESET_TUTORIAL` param) and see that the tutorial is not triggered.
* If you access with a user that claimed a name, the claim name panel (after closing the first avatar editor) won't show up, only with no-claimed users (that can be triggered accessing from an incognito window with ENV=zone instead of org)


moliva (approved):

kernel/packages/decentraland-loader/lifecycle/controllers/download.ts
```diff
@@ -29,7 +30,9 @@ export class SceneDataDownloadManager {
   emptyScenesPromise?: Promise<Record<string, ContentMapping[]>>
   emptySceneNames: string[] = []
 
-  constructor(public options: { contentServer: string; contentServerBundles: string }) {}
+  constructor(public options: { contentServer: string; contentServerBundles: string; tutorialBaseURL: string }) {
+    // stub
```
AjimenezDCL, line 34: What about this? Are we missing something here? If there's something to be implemented here can we change it with a proper // TODO?

kernel/packages/decentraland-loader/lifecycle/controllers/download.ts
```diff
@@ -201,14 +204,45 @@ export class SceneDataDownloadManager {
   }
 
   async getParcelDataBySceneId(sceneId: string): Promise<ILand | null> {
+    if (isTutorial()) {
```
AjimenezDCL, line 207: All these checks for tutorial make me think we could implement some inheritance based architecture here. Probably is too much effort and too little gain, but maybe is it worth taking a look into it in the future.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AirdroppingHUD/AirdroppingHUDController.cs
```diff
@@ -39,6 +39,8 @@ public enum State
     internal int currentItemShown = -1;
     internal int totalItems => model?.items?.Length ?? 0;
 
+    public static event System.Action OnAirdropFinished = null;
```
AjimenezDCL, line 42: I really dislike having this event and coupling other classes to AirdroppingHUD. I think a better approach would have been ignoring the huds and such and through inheritance create what was needed for the tutorial, and once the tutorial was finished, fallback to the normal flow.

Anyway, I won't get too picky since this does the job, but I would recommend a refactor in the future.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/HUDController.cs
```diff
@@ -169,6 +169,30 @@ public void ConfigureAirdroppingHUD(string configurationJson)
         airdroppingHUDController?.SetVisibility(configuration.active && configuration.visible);
     }
 
+    const int NOTIFICATION_DURATION = 5;
+
+    public void ShowWelcomeNotification()
```
AjimenezDCL, line 174: This goes against the whole convention we have been using for the HUDs and specifically for the notifications.

Certainly Kernel should be capable of accessing the needed information to provide a model through the existent `ShowNotificationFromJson` method.

AjimenezDCL (changes requested):

kernel/packages/decentraland-loader/lifecycle/controllers/download.ts
BrianAmadori: Nailed it. I already have this redesign in a separate branch, but I don't want to expose this for now because we are going to merge it by 2022.

AjimenezDCL: Keeping the good stuff for yourself :(.

Nice to hear!

kernel/packages/decentraland-loader/lifecycle/controllers/download.ts
pravusjif: I have no idea why that new param is in that constructor XD it precedes my intervention of this PR, I'll update the comment at least

moliva: There is no `TODO` here, but an empty block of code will generate a warning, that's why we have a stub comment.
The constructor is necessary though to set the options field with the required `tutorialBaseURL`.

pravusjif: ok so we will leave it as it is @AjimenezDCL

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/HUDController.cs
pravusjif: I created an issue with all the info to be tackled in the future:
https://app.zenhub.com/workspaces/explorer-5dc988c462b0b700019b0b42/issues/decentraland/explorer/549

AjimenezDCL: Nice, approving

AjimenezDCL (approved):

PR #460, opened 2020-01-31
fix: kernel should first await connection then join layer

pablitar:
# What? 
Changing the order of connection steps when connecting to lighthouse

# Why? 
Because it doesn't make sense joining a layer if you cannot establish peer connections


eordano:
Thanks, this was getting annoyinh

eordano (approved):

PR #461, opened 2020-02-01
feat: empty parcel simple import from builder

eordano:
This allows the following workflow:

0) `make watch` and open http://localhost:8080
A) Build a scene in the builder
B) Download it and unzip it in the folder `kernel/static/loader/empty-parcels`
C) Run `make empty-parcels`
D) Refresh the client to see how it looks like.


kernel/packages/decentraland-loader/lifecycle/controllers/download.ts
```diff
@@ -28,12 +25,22 @@ export class SceneDataDownloadManager {
...
+  emptyScenesPromise?: Promise<Record<string, ContentMapping[]>>
+  emptySceneNames: string[] = []
 
-  constructor(public options: { contentServer: string; contentServerBundles: string }) {
-    // stub
-  }
+  constructor(public options: { contentServer: string; contentServerBundles: string }) {}
 
   async resolveSceneSceneId(pos: string): Promise<string | null> {
+    if (!this.emptyScenesPromise) {
```
BrianAmadori, line 35: This method is supposed to return the scene giving it an `sceneId`. If the first request isn't an empty parcel the logic would break. Without mentioning that it also breaks the contract of the method. There isn't a better way of doing this?

BrianAmadori (changes requested):

kernel/packages/decentraland-loader/lifecycle/controllers/download.ts
eordano: This is lazy initialization. This if will be hit only one time, why do you think it breaks the contract?

pravusjif (approved):

kernel/packages/decentraland-loader/lifecycle/controllers/download.ts
BrianAmadori: This method is not for initialization, is for retrieving a sceneData using a sceneId by param. If the emptyScenesPromise is not resolved and some scene is fetched, incorrect data will be returned (param will be ignored). Can't this code be moved outside this function? to another method/constructor?

BrianAmadori (approved):

PR #471, opened 2020-02-03
fix:  #419 prevent survey in the first loading screen

### LINKED ISSUE #419: NPS survey "recommend to a friend" popup can appear in loading screen
The dialog is showing in loading screen. When the loading takes too much time it can be observed.

It should only appear when:
- Rendering is activated (loading screen not shown)
- Some time has passed in the world exploration state

On top of fixing that, let's configure Delighted so that the survey only appears from the 2nd play session onwards (or from the 2nd day, both are OK).

2fd:
Fix #419

# What? 
Prevent display delighted's survey


kernel/packages/unity-interface/dcl.ts
```diff
@@ -194,13 +195,14 @@ export function setLoadingScreenVisible(shouldShow: boolean) {
   document.getElementById('load-messages-wrapper')!.style.display = shouldShow ? 'block' : 'none'
   document.getElementById('progress-bar')!.style.display = shouldShow ? 'block' : 'none'
   if (!shouldShow) {
+    isTheFirstLoading = false
```
BrianAmadori, line 204: When the user is in tutorial flow, this wouldn't show the survey popup as soon as the 2nd teleport is triggered? Chat and other features will not be available yet. Are we sure this is the intended behaviour?

moliva: I agree with Brian, maybe we need another criterion to show this survey or a combination of them, e.g. finished tutorial + certain elapsed in world or teleport

2fd: I agree with you guys, this isn't the optimal solution but was the best effort/benefit. I have other priorities, but you are free to implement these changes

BrianAmadori (approved):

kernel/packages/unity-interface/dcl.ts
BrianAmadori: Ok, just approving it. We may improve this in the future.

moliva (approved):

PR #472, opened 2020-02-03
fix: blocking promises performance spike

BrianAmadori:


BrianAmadori:
Closing this PR because its a bit frankensteiny. Splitted it into smaller ones.

PR #479, opened 2020-02-04
feat: Sending peer position (as parcel) to lighthouse for optimization

pablitar:
# What? 
The explorer starts sending the parcel positions of peers

# Why? 
Because we can use that information to optimize network connection


moliva (approved):
Amazing! 🌟

PR #484, opened 2020-02-05
feat: katalyst realm picking

moliva:
# What? 
- Pick katalyst realm from DAO based on number of users and latency
- Consume all necessary services from the selected realm
- Remove centralized configurations


pablitar (approved):
This is great!!!

PR #485, opened 2020-02-05
feat: add new params to AssetBundleBuilder

BrianAmadori:
New params can be used as follows:

`-baseUrl <url>` 

Changes the baseUrl from `http://content.decentraland.org` to `<url>`. The url must *not* end with a dash.

`-output <absolute-path>`

Asset bundles will be dumped in `<absolute-path>/AssetBundles`


BrianAmadori:
This will be merged only when @marcosnc confirms it works correctly when hooked up to the Asset Bundle conversion service.

AjimenezDCL (approved):

PR #492, opened 2020-02-07
Release 2020-02-07

moliva:
# What? 
...

# Why? 
...


kernel/packages/shared/crypto/Authenticator.ts
```diff
@@ -0,0 +1,201 @@
+import { hash, sign, recover } from 'eth-crypto'
```
eordano, line 1: Should we use the lib here?

eordano (approved):

kernel/packages/shared/crypto/Authenticator.ts
moliva: Indeed. Adding it to next release.

BrianAmadori (approved):

pablitar (approved):
Testing shows it should be OK. Crossing my fingers!

PR #500, opened 2020-02-08
Signing expiration date to iso

pablitar:
# What? 
Instead of formatting date using default string formatting, it uses ISO8601

# Why? 
This avoids potential locale problems when generating the message to be signed


marcosnc (approved):
Thank you for the fast response!

PR #507, opened 2020-02-10
chore: Updating peer version + Changed scene comms to its own message type

pablitar:
# What? 
Updating peer library version
Changed scene comms type to be optimistic

# Why? 
To include some fixes to network connectivity an room joining:
See: https://github.com/decentraland/katalyst/commit/1981f48450d26a8f4ffeda8682945ddcd4583e94
and https://github.com/decentraland/katalyst/commit/a6f9193286557ada51a699c5bb1620dd2f2cdaad

Also changed scene comms type to be optimistic, in order for the queries to consume less resources

### DIFF
--- a/kernel/package-lock.json
+++ b/kernel/package-lock.json
@@ -2685,9 +2685,9 @@
       }
     },
     "decentraland-katalyst-peer": {
-      "version": "0.1.23",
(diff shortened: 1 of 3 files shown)


moliva (approved):
LGTM!

PR #513, opened 2020-02-10
chore: changing wording and trigger for id taken error

moliva:
# What? 
...

# Why? 
...


kernel/packages/shared/index.ts
```diff
@@ -258,7 +261,11 @@ export async function initShared(): Promise<Session | undefined> {
 
       break
     } catch (e) {
-      if (e.message && e.message.startsWith('error establishing comms')) {
+      if (e.message && e.message.includes('is taken')) {
```
BrianAmadori: Checking errors like this is very error prone. We should use type checks instead of this.

eordano: Agreed! Do we have a list of these errors?

moliva: Addressed! We do not have a list of them, but rather create the difference when trying to treat them separately.

eordano:
Heads up @xpalaum this reverts the usability of "log in on multiple instances". We should be aware of this and notify Sam.

eordano (approved):

PR #515, opened 2020-02-10
fix: invalid loading timeout

BrianAmadori:
Renderer holds sending the `ActivateRenderingACK` message until:
* Current scene finishes downloading.
* Avatar preview is ready.

This should fix:
* Pressing `done` before avatar preview is ready in avatar editor crashes explorer
* Being stuck inside meshes when loading dissapears. This happening mostly on tutorial scenes/big plazas.


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/LoadingLock.cs
```diff
@@ -0,0 +1,40 @@
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace DCL
+{
+    public class LoadingLock
```
AjimenezDCL: I think this whole class could be static and we can avoid coupling `RenderingController` to other systems.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -32,6 +32,7 @@ enum State
 
         public event System.Action<DecentralandEntity> OnEntityAdded;
         public event System.Action<DecentralandEntity> OnEntityRemoved;
+        public event System.Action<State> OnStateChange;
```
AjimenezDCL: Is this needed? I cannot find uses of this event.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -50,11 +51,14 @@ enum State
         [System.NonSerialized]
         public bool unloadWithDistance = true;
 
+        [System.NonSerialized]
+        public bool lockRendering = true;
```
AjimenezDCL: Few concerns about this. 
First of all, is it needed? it's only used to avoid unlocking without locking first, but since the `LoadingLock` is doing a `Contains` check, I think we can safely remove it.

If I'm missing something and it's really needed, must it be public? I dont see uses of it.

AjimenezDCL (changes requested):

eordano (commented):
Can we use a different name for `lock` and `rendering`?

I'm having trouble understanding what's the meaning of "Lock Rendering".

Maybe something like:

```
class ResourceLoadingCounter {
  Set<string> resourceIdLoading = new HashSet<>();
  int loadingCount { private set, get };
  
  public bool isLoading => loadingCount == 0;

  public void MarkLoading(String id) { /** maybe throw if it's already on the set ? */ }
  public void MarkFinished(String id);
}
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
BrianAmadori: Yeah, I'll just remove it.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/LoadingLock.cs
BrianAmadori: I'm trying to avoid making more singletons or static classes, and for some reason I think this class can be useful to locking other things, that's the reason I'm not making it static. Moreover, in this case the `LoadingLock` is directly related to `RenderingController`, because `RenderingController` depends on it in order to dispatch the ACK message. 

I have no doubt that we have to decouple `RenderingController` of all the systems it's accessing directly, but I think this is not the case.

AjimenezDCL: Sounds about right. 

What do you think of removing the `Loading` from the name (maybe adjusting the rest a bit to make it clear) and put it in Utils?

BrianAmadori:
@eordano its not a loading ref counter. It just an generic semaphore class you can use to lock out the `ACK` response until it gets to 0. 

As seen by looking at the changes, its currently incremented/decremented with following activities:

- AvatarEditor starts loading the preview avatar model (decrementes when finishes)
- ParcelScene of current scene starts loading (decrementes when finishes) 

I can remove all `rendering` references, but i'm not sure of a better name for this class.

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/LoadingLock.cs
BrianAmadori: Sounds good, going to try it.

AjimenezDCL (approved):

PR #519, opened 2020-02-11
feat: fake airdrop is back

eordano:


BrianAmadori (approved):
Sounds good for the time being but let's not forget to remove it at launch.

PR #524, opened 2020-02-11
fix: concurrent requests limit for asset bundles was not working at all.

BrianAmadori:


eordano (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
```diff
@@ -205,5 +200,31 @@ protected override void OnLoad(Action OnSuccess, Action OnFail)
...
+        {
+            if (concurrentRequests >= MAX_CONCURRENT_REQUESTS)
+            {
+                yield return new WaitUntil(() => concurrentRequests < MAX_CONCURRENT_REQUESTS);
+            }
+        }
+
+        void RegisterConcurrentRequest()
+        {
+            if (requestRegistered)
```
AjimenezDCL, line 214: Although I like the idea of this safe check, it makes me worry about the flow. Shouldn't it be in the `LoadAssetBundleWithDeps` method? Maybe I'm missing something, but if a request is registered already we shouldn't have to wait for a slot, neither  call `LoadAssetBundle`.

AjimenezDCL (approved):

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
BrianAmadori: Maybe a better approach should be to handle the concurrency limit from `AssetPromiseKeeper`. But its a bit hard to implement right now. Maybe we can improve upon this in the future.

PR #525, opened 2020-02-11
fix: scene sorting was being performed ten times in the same frame in some cases

BrianAmadori:
This issue was introducing 100 ms hiccups.

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
@@ -138,6 +138,7 @@ namespace DCL
         }
 
         private bool sceneSortDirty = false;
+        private int lastSortFrame = 0;
         public event Action OnSortScenes;
 
         private static readonly int MORDOR_X = (int)EnvironmentSettings.MORDOR.x;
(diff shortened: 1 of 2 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -469,7 +474,8 @@ public void LoadParcelScenes(string decentralandSceneJSON)
 
             MessagingControllersManager.i.ForceEnqueueToGlobal(MessagingBusId.INIT, queuedMessage);
 
-            TrySortScenesByDistance(forceSort: true);
+            sceneSortDirty = true;
+            TrySortScenesByDistance();
```
eordano: I feel like we can keep the `true` here, what do you think?

```suggestion
            TrySortScenesByDistance(true);
```

eordano (approved):
Looks good to me. Minor consideration

AjimenezDCL (approved):
I think we could tackle down the issue in an easier way.

Surely `SceneController.LoadParcelScenes` is the one causing the hiccups. Do we really need that method to call `TrySortScenesByDistance`? I think we can remove the call (and therefore remove the `forceSort` param) and only call `TrySortScenesByDistance` in the update using the existent `sceneSortDirty` logic. The biggest issue could be 1 frame delay between detecting the player's new position.

I'm approving nonetheless,

pravusjif (approved):

pravusjif:
> I think we could tackle down the issue in an easier way.
> 
> Surely `SceneController.LoadParcelScenes` is the one causing the hiccups. Do we really need that method to call `TrySortScenesByDistance`? I think we can remove the call (and therefore remove the `forceSort` param) and only call `TrySortScenesByDistance` in the update using the existent `sceneSortDirty` logic. The biggest issue could be 1 frame delay between detecting the player's new position.
> 
> I'm approving nonetheless,

this may be worth checking out

PR #526, opened 2020-02-12
chore: pack analytics every 15 seconds

eordano:


eordano:
CC: @dnul is this OK?

CC: @2fd is this OK?

dnul:
Only problem I see, is that it messes with the timestamp of the events and hence it  can alter the ordering of them. 

Also during loading, lifecycle_events are sent alltogether so we could miss some of them if it drops below 15 secs from loading.

eordano:
Oh, @dnul but `queueTrackingEvent` fills in the `time` field with `new Date().getTime()` (see line 44 of the modified file)

dnul (approved):

PR #527, opened 2020-02-12
fix: MY EYES! -- always use .org for hard-coded content

eordano:
Which is currently only some wearables

Fixes no-eyes problem:
image
To:
image


kernel/packages/shared/passports/sagas.ts
```diff
@@ -178,7 +178,7 @@ function takeLatestById<T extends Action>(
 function overrideBaseUrl(wearable: Wearable) {
   return {
     ...wearable,
-    baseUrl: getContentUrl() + '/contents/',
+    baseUrl: getWearablesSafeURL() + '/contents/',
     baseUrlBundles: getServerConfigurations().contentAsBundle + '/contents/'
```
eordano:
```suggestion
    baseUrlBundles: getServerConfigurations().contentAsBundle + '/'
```

PR #532, opened 2020-02-12
fix: object lifecycle fixes

BrianAmadori:
## Code cleaning:

* `ColliderInfo` now stores references to entity/scene instead of ids because is more convenient.
* Removed `ColliderInfo` instance from `HitInfo` as it made a bit difficult to track bugs. `ColliderInfo` ocurrences should be always gathered through the `ColliderManager`.
* Reduced indenting in a bunch of methods
* Removed Rigidbody usage in all OnPointerX methods as it is not really needed. Rigidbody was only used as means of getting the root entity object of colliders. We have `ColliderManager` for that.
* Removed unused/unneeded methods in `UUIDComponent` setup.
* Now `OnPointerEventColliders` colliders are tracked in `ColliderManager` where before they weren't for some reason.
* Fixed all code in `PointerEventsController` that expected `OnPointerEvent` components to be in the same hierarchy as the entity gameObject.
* Removed duplicated test `OnPointerHoverFeedbackIsDisplayedCorrectly`

## Fixes: 

* fix: removing the same entity multiple times wasn't being ignored as it should, leading to potential errors or unnecessary performance drops.

* fix: pooled entities with uuid objects weren't being cleaned as it should, leading to incorrect interact toasts in unwanted entities

* fix: pooled entities didn't have their transform resetted correctly, leading to erratic transform values in arbitrary entities (i.e. randomic gigantic objects, misplaced objects etc)

* fix: ocassional nullReference error in OnPointerEvent.OnDestroy

---

Those fixes were performed testing specifically the farm scene. The observable results should be:

* Inverted billboards over farm plots should display correctly
* When clicking over a farm plot, sometimes the popup didn't appear. Now it should work correctly.
* Sometimes a null reference exception ocurred when interacting with objects. Now it should work correctly.
* Clicking over the fridge wasn't working at all. Now it works.


pravusjif (approved):

AjimenezDCL (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/CollidersManager/CollidersManager.cs
```diff
@@ -65,9 +65,9 @@ void AddOrUpdateEntityCollider(DecentralandEntity entity, Collider collider)
                 collidersList.Add(collider);
 
             ColliderInfo info = new ColliderInfo();
-            info.entityId = entity.entityId;
-            info.meshName = collider.transform.parent.name;
-            info.sceneId = entity.scene.sceneData.id;
+            info.entity = entity;
+            info.meshName = collider.transform.parent != null ? collider.transform.parent.name : "";
```
AjimenezDCL, line 69: I'm not in the computer here to check it but I think you can use a null coalescing operator.

`info.meshName = collider.transform.parent?.name ?? "";`

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PhysicsCast/PhysicsCast.cs
```diff
@@ -35,11 +37,16 @@ private void HitFirst(RaycastQuery query)
         {
             WebInterface.RaycastHitEntity hitEntity;
 
-            RaycastResultInfo raycastInfo = raycastHandler.Raycast(GetUnityRayFromQuery(query), query.ray.distance, ~layerMaskTarget, query.sceneId);
+            ParcelScene scene;
+            SceneController.i.TryGetScene(query.sceneId, out scene);
```
AjimenezDCL, line 41: You can declare scene in the same out param.

AjimenezDCL (approved):

pravusjif (approved):
the pointers test scene at https://explorer.decentraland.zone/branch/fix/object-lifecycle-fixes/index.html?ENV=zone&position=54%2C-71&realm=zeus-amber seems to work fine, let's wait for @nearnshaw test to see if anything comes up, but after that we are ready to merge

PR #533, opened 2020-02-12
fix: sending scene ready too soon

BrianAmadori:
Scene ready wasn't taking into account entity components. This could lead to some scenes sending the `SceneReady` message when they still have some pending components.

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Utils.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Utils.cs
@@ -167,6 +167,8 @@ namespace DCL.Helpers
             {
                 using (var webRequest = request)
                 {
+                    webRequest.timeout = 30;
+
                     yield return webRequest.SendWebRequest();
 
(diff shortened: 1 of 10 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Utils.cs
```diff
@@ -167,6 +167,8 @@ public static bool WebRequestSucceded(this UnityWebRequest request)
             {
                 using (var webRequest = request)
                 {
+                    webRequest.timeout = 30;
```
eordano, line 170: Ouch!

I don't think this is enough to download some stuff.

BrianAmadori: Right.

Anyway I think i'm going to close this PR because there's an easier fix for this. Going to decide when I get back to this.

BrianAmadori:
I'm closing this for now because it doesn't add significant improvements and we have other priorities. Will review if its worth reopening later.

PR #534, opened 2020-02-13
fix: avatar skewing

AjimenezDCL:
A side effect of using non-uniform scaling with parenting is the skewing of the childs. Adding a node with a  (1,1,1) scale between the non-uniform parent and the content should solve the issue.

EDIT: The above is invalid as a general solution. It will still fail in nested entities with arbitrary rotations/scales. The only way to solve this is by taking the CharacterRepresentation out of the parenting.

### DIFF
--- a/unity-client/Assets/Resources/Prefabs/CharacterController.prefab
+++ b/unity-client/Assets/Resources/Prefabs/CharacterController.prefab
@@ -25,11 +25,11 @@ Transform:
   m_PrefabInstance: {fileID: 0}
   m_PrefabAsset: {fileID: 0}
   m_GameObject: {fileID: 499268581}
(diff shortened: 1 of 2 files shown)


pravusjif:
Where a re we rotating the character now? if we don't use its `transform.forward` anymore

AjimenezDCL:
> Where a re we rotating the character now? if we don't use its `transform.forward` anymore

`rotationTransform.forward = characterForward.Get().Value; ` there

pravusjif (approved):

BrianAmadori:
I don't see how adding an intermediate parent should fix the skewing, as the skewed scale is still carried over. How is this working @AjimenezDCL ?

AjimenezDCL:
You are right, Brian. Is not a general solution. Gonna change the approach

AjimenezDCL:
Already fixed

PR #535, opened 2020-02-13
feat: add current realm data to SDK

moliva:
# What? 
Added realm data to SDK in EnvironmentAPI

# Why? 
Allow scene creators to take into account where are users currently interacting and match appropriately.


pablitar (approved):
Great!

PR #538, opened 2020-02-14
refactor: improve UI refresh performance

pravusjif:
Implemented a hierarchy tracking in a linked list in UIShapes to try another alternative at refreshing the elements


pravusjif:
I'm closing this PR ='( as I see no improvement in performance by tracking the ui elements in a local uiTree (LinkedList).

There is another recursive traversal that generates CPU and GC spikes: UISizeFitter, but that would require a different tree (LinkedList) tracking, and not sure it is worth it, as doing the same with the ReferenceContainers wasn't.

We can keep the branch in case we want to check this out again in the future.

PR #547, opened 2020-02-14
feat: Adding a command to switch to a layer with people in the same

pablitar:
# What? 
Adds the command `/changerealm crowd` to switch to a realm that has people in the current location

# Why? 
To allow for people to change realms to any that has people


kernel/packages/shared/dao/index.ts
```diff
@@ -56,7 +61,7 @@ function ping(url: string): Promise<{ success: boolean; elapsed?: number; result
       }
     }
 
-    http.open('GET', url, false)
+    http.open('GET', url, true)
```
eordano, line 64: <3

eordano (approved):

PR #548, opened 2020-02-14
chore: add 30 sec delay

eordano:


BrianAmadori:
This is kind of going against our target of loading time < 30 secs, and the fix for green blockers galore is already in the tutorial PR. This is worth it?

eordano:
> This is kind of going against our target of loading time < 30 secs, and the fix for green blockers galore is already in the tutorial PR. This is worth it?

every time someone sees these blockers a kitty dies. if one user stops seeing it because of this, merge.

PR #550, opened 2020-02-15
style: design improvements on HUDs

RominaMarchetti:
### **What?**
This branch replaces the old main UI atlas.
Sprites were replaced on Mini map, Avatar HUD, Chat, UI visibility. 
Architectural adjustments and minor color changes applied.
Canvas sorting adjusted for all the HUDS.
Text shadows fixed.
...

### **Why?**
Sprites looked low quallity and the ex atlas path was not the correct way for working. We had trouble with canvas organization (wrong overlapping). 

...


BrianAmadori (approved):
Chat looks way better!

PR #557, opened 2020-02-16
fix: tutorial avatar editor cursor lock

pravusjif:
* Set the avatar editor hud background image as racyast target (to avoid catching the mouse on the mouse catcher in certain areas) 

* Added mouse catcher unlock at the beginning of tutorial initial step

This way the camera won't move while using the avatar editor at the tutorial, you can test this at
https://explorer.decentraland.zone/branch/fix/TutorialAvatarEditorCursorLock/index.html?RESET_TUTORIAL


eordano (approved):

unity-client/Assets/Tutorial/Scripts/TutorialSteps/TutorialStep_Initial.cs
```diff
@@ -32,6 +38,8 @@ public override void OnStepStart()
 
             if (HUDController.i != null && HUDController.i.avatarEditorHud != null)
             {
+                mouseCatcher.UnlockCursor();
```
BrianAmadori: We should use `Utils.UnlockCursor` here instead of using the mouseCatcher this way.

BrianAmadori (approved):

PR #582, opened 2020-02-18
Fix: runloops URL and headless chrome (run as root)

mgagliardo:
* Fixing run chrome as root (needs flag `--no-sandbox`)
* Fixing invalid URL


moliva (approved):
Amazing!

PR #583, opened 2020-02-18
fix: tutorial

pbosio:
fix: timer running when rendering was disabled
fix: tutorial step restarting when player teleports
fix: player spawning on tutorial scene even if that step was already finished
fix: avatar hud visible in tutorial scene step
fix: expose time between tooltips variable in tutorial step prefab
fix: wait until tooltip hides to continue with the tutorial step flow


kernel/packages/config/index.ts
```diff
@@ -138,6 +138,10 @@ export function tutorialEnabled() {
   )
 }
 
+export function tutorialSceneEnabled() {
+  return tutorialEnabled() && (RESET_TUTORIAL || getUserProfile().profile.tutorialStep === tutorialStepId.INITIAL_SCENE)
+}
```
pravusjif, line 143: in `packages/config/index.ts` we have an `export function tutorialEnabled()` very similar to this one, can't we use the already existing one?

pravusjif (approved):

unity-client/Assets/Tutorial/Scripts/TutorialStep.cs
```diff
@@ -28,11 +37,13 @@ public virtual IEnumerator OnStepExecute()
...
-        WaitForSeconds waitForIdleTime = new WaitForSeconds(TutorialController.DEFAULT_STAGE_IDLE_TIME);
-
         public virtual IEnumerator WaitIdleTime()
         {
-            yield return waitForIdleTime;
+            if (waitForIdleTime != null)
+            {
+                yield return waitForIdleTime;
+            }
+            yield return waitForRendererEnabled;
```
AjimenezDCL, line 46: Shouldn't this be the other way around? so we dont miss any step if the loading is taking long

AjimenezDCL (approved):

unity-client/Assets/Tutorial/Scripts/TutorialStep.cs
pbosio: No, cause this way, no matter if renderer is disabled when time starts running, it will wait until the renderer is enabled again. If we do it the other way around, then if renderer is disabled after checking if renderer is enable and we are yielding the WaitForSeconds we will miss it

kernel/packages/config/index.ts
pbosio: No, cause is nice to have a way to discriminate if we running tutorials vs we are running tutorial scene (the scene is just a part of the tutorial).
Checking if kernel is in tutorial is used in other places of the code (like in the welcome notification when user spawn in a scene) and checking if it's should be running the tutorial scene is used for the code that force that scene and not others are loaded

unity-client/Assets/Tutorial/Scripts/TutorialStep.cs
AjimenezDCL: What I mean is first waiting for renderer to be enabled and then do the wait for the idle. 

With the current setup the time will count even if we are not rendering anything, right?

PR #585, opened 2020-02-18
fix: Canvas scaler match change from 0 to 1

RominaMarchetti:
# What?
_Canvas scaler match_ changed from 0 to 1

# Why? 
Because it was colliding with the chat window. This way won't happen again.


BrianAmadori (approved):

pbosio (changes requested):
Emotes' icon scales correctly but it's misplaced (it's behind chat window)

pravusjif (approved):
After addressing pato's request it should be fine

RominaMarchetti:
@pbosio @pravusjif  ready for review again.

pbosio (approved):

PR #586, opened 2020-02-18
fix: user account check

moliva:
# What? 
- Checks that stored session matches current user or invalidate otherwise
- Register to account change, network change and wallet disconnection events to reload the application
- Adds check on enable to support special Dapper case when not logged in


kernel/packages/shared/ethereum/provider.ts
```diff
@@ -38,9 +45,17 @@ export async function awaitWeb3Approval(): Promise<void> {
...
-              result = { successful: true, provider: window.ethereum }
+              if (accounts && accounts.length > 0) {
+                result = { successful: true, provider: window.ethereum }
+              } else {
+                // whether accounts is undefined or empty array => provider not enabled
+                result = {
+                  successful: false,
+                  provider: createProvider()
+                }
+              }
```
BrianAmadori: This code is repeated below. Can't we extract it to a method?

pablitar (approved):
Great work!

BrianAmadori (approved):

PR #593, opened 2020-02-19
fix: moving animation for own character issues

AjimenezDCL:


BrianAmadori:
The timing is a bit off. The speed should slightly higher to avoid the moonwalk effect.

PR #596, opened 2020-02-19
feat: Picking candidates by amount of users

pablitar:
# What? 
When picking up a candidate for realm, if it is time to pick an empty layer, the explorer now calculates the amount of users that the server has, and prioritizes those with less users

# Why? 
To avoid overloading a particular catalyst


moliva (approved):
Awesome! 🎉

kernel/packages/shared/dao/index.ts
```diff
@@ -120,11 +120,30 @@ async function fetchCatalystStatuses(nodes: { domain: string }[]) {
...
   const sorted = candidates
     .filter(it => it.layer.usersCount < it.layer.maxUsers)
     .sort((c1, c2) => {
-      const diff = c2.score - c1.score
-      return diff === 0 ? c1.elapsed - c2.elapsed : diff
+      const elapsedDiff = c1.elapsed - c2.elapsed
+      const usersDiff = usersByDomain[c1.domain] - usersByDomain[c2.domain]
+      const scoreDiff = c2.score - c1.score
+
+      return Math.abs(elapsedDiff) > 1500
```
BrianAmadori, line 140: This expression is a bit painful to read. I'd spread it a bit. So it looks like:
```
if ( bigLatencyDifference ) { 
   return elapsedDiff
} else if ( hasScoreDifference ) {
   return scoreDiff
} else if ( hasUsersDifference ) {
   return usersDiff
} else { 
   return elapsedDiff
}
```

Also, some sort of unified cumulative score wouldn't be a better strategy here? 
Those values are comparable for sorting (i.e. we aren't sorting with pears and oranges)?

eordano (approved):

PR #600, opened 2020-02-19
fix: tutorial not triggering correctly for no crypto users

pbosio:
Ensure that tutorial is triggered correctly for no crypto users and old users profile


BrianAmadori (approved):

moliva (approved):
👍

PR #601, opened 2020-02-19
fix: being john malkovich issue

BrianAmadori:
`AvatarRenderer.OnSuccessCallback` / `OnFailCallback` events weren't being triggered when suscribed from outside because the class overrided them.

Also removed the trigger approach as a defensive measure because OnTriggerEnter/Exit events aren't 100% reliable.


AjimenezDCL (approved):
The colliders that made this possible (in camera and in the avatar) should be removed as well

pravusjif (approved):
It works fine

PR #604, opened 2020-02-19
feat: New score function

pablitar:
# What? 
Changed score function to try to concentrate people in one layer up until it reaches 75% fullness

This is the curve of the new score function for 50 people:

image

Keep in mind that when a layer is empty or full, those are considered special cases

# Why? 
We get cases in which there are around 30 people in the world, but they are distributed in 4 or 5 layers, and that feels lonely


moliva (approved):
Alright! 💪

PR #605, opened 2020-02-19
fix: Don't crash when a catalyst server is down

pablitar:
# What? 
Before this fix, the explorer would crash when a Catalyst server didn't respond (connection refused, or network failure). The only case that was handled was when it responded with error code

# Why? 
Crashes are bad

### DIFF
--- a/kernel/packages/shared/dao/index.ts
+++ b/kernel/packages/shared/dao/index.ts
@@ -46,16 +46,23 @@ function ping(url: string): Promise<{ success: boolean; elapsed?: number; result
         started = new Date()
       }
       if (http.readyState === XMLHttpRequest.DONE) {
-        const ended = new Date().getTime()
-        if (http.status >= 400) {
+        try {
+          const ended = new Date().getTime()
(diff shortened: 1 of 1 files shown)


moliva (approved):

kernel/packages/shared/dao/index.ts
```diff
@@ -86,11 +93,8 @@ export async function fecthCatalystRealms(): Promise<Candidate[]> {
 
 async function fetchCatalystStatuses(nodes: { domain: string }[]) {
   const results = await Promise.all(nodes.map(node => ping(`${node.domain}/comms/status?includeLayers=true`)))
-  const successfulResults = results.filter($ => $.success)
-  if (successfulResults.length === 0) {
-    throw new Error('no node responded')
-  }
-  return zip(nodes, successfulResults).reduce(
+
```
pablitar, line 96: This filter was unnecessary, because it filters during the reduce. Also it caused an error because the nodes list and the successfulResults list were of different sizes when zipping them.

PR #610, opened 2020-02-19
fix: avatar animation on moving platforms

pravusjif:
Updated AvatarAnimatorLegacy to user the character controller velocity instead of the object's movement delta


unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarAnimatorLegacy.cs
```diff
@@ -87,7 +88,10 @@ void UpdateInterface()
 
         flattenedVelocity.y = 0;
 
-        blackboard.movementSpeed = flattenedVelocity.magnitude;
+        if (isOwnPlayer && DCLCharacterController.i.velocity.x == 0f && DCLCharacterController.i.velocity.z == 0f)
+            blackboard.movementSpeed = 0f;
+        else
+            blackboard.movementSpeed = flattenedVelocity.magnitude;
```
BrianAmadori: When you move, the `movementSpeed` will be the sum of the platform speed and the player speed, giving undefined behaviour. A way easier and more accurate fix shouldn't be to calculate this speed just using `localPosition` and leaving this code intact?

pravusjif: I'll test that

BrianAmadori (approved):

pbosio (approved):

PR #611, opened 2020-02-19
fix: tutorial show MOTD and ask for user's email popup

pbosio:


unity-client/Assets/Tutorial/Scripts/TutorialSteps/TutorialStep_ChatAndAvatarExpressions.cs
```diff
@@ -44,6 +69,45 @@ public override IEnumerator OnStepExecute()
 
             yield return dailyRewardTooltip.ShowAndHideRoutine();
             yield return WaitIdleTime();
+
+            bool hasWelcomeHudPop = false;
+            if (HUDController.i != null && HUDController.i.welcomeHudController != null)
+            {
+                isWelcomeHudVisible = true;
+                hasWelcomeHudPop = true;
+                Utils.UnlockCursor();
```
BrianAmadori: This is not necessary, should be already handled by the welcome popup.

pravusjif (approved):
I got through all the tutorial and then got the MOTD and the email prompt, great.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WelcomeHUD/WelcomeHUDController.cs
```diff
@@ -10,19 +10,18 @@ public class Model : HUDConfiguration
         public bool hasWallet; //TODO(Brian): Use WEB3 param in userProfile
```
BrianAmadori, line 10: I implemented this before the profile change. Can we use the profile web3 flag here and remove this? We should remove it from kernel too.

unity-client/Assets/Tutorial/Scripts/TutorialSteps/TutorialStep_ChatAndAvatarExpressions.cs
pbosio: Added lock and unlock cursor superpowers to WelcomeHUDController

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WelcomeHUD/WelcomeHUDController.cs
pbosio: unity receive the profile after WelcomeHUDController is initialized, so we currently can use that flag.
maybe we should address this in a future PR

BrianAmadori (approved):

PR #621, opened 2020-02-20
Pixelated UI fix

RominaMarchetti:


BrianAmadori (commented):
Did you test this thoroughly? I think you should try to enable mipmapping and trilinear filtering too, because the browser is scaling the canvas down and this is what gives us the pixelated issues.

PR #622, opened 2020-02-20
feat: add backup servers

moliva:
# What? 
Add backup servers

# Why? 
Add backup servers when realms are really crowded, while still fetching and updating profiles from the ones in the DAO to maintain state consistency.


pablitar (approved):
Great work!

PR #626, opened 2020-02-20
fix: all parenting related moving platform issues

BrianAmadori:
This should address:

* The avatar skewing issue
* Parent deactivating the character controller issue
* Extra code to handle moving platforms in DetachShape
* All the weird things that can happen if we don't unparent the character when we should


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -367,6 +367,7 @@ public void RemoveEntity(string id, bool removeImmediatelyFromEntitiesList = tru
             SceneController.i.OnMessageDecodeEnds?.Invoke("RemoveEntity");
             if (entities.ContainsKey(tmpRemoveEntityMessage.id))
             {
+                Debug.Log("removing entity " + tmpRemoveEntityMessage.id);
```
AjimenezDCL: Remove comments

AjimenezDCL (approved):

pbosio (changes requested):
After jumping Avatar falls in "slow motion"

pbosio (approved):

PR #641, opened 2020-02-21
feat: goto next treasure buttons

pbosio:
Add button below mini map to go to next treasure hunt scene.
Add button to go to next treasure hunt scene when an empty chest is opened.
Add popup to redirect user to next treasure hunt scene when airdrop's transaction is accepted/rejected.


BrianAmadori:
Closing because we are cancelling the daily reward concept.

PR #646, opened 2020-02-26
fix: peer reconnection

moliva:
# What? 
Connect to another realm when reconnection fails on peer library

### DIFF
--- a/kernel/packages/shared/comms/index.ts
+++ b/kernel/packages/shared/comms/index.ts
@@ -57,7 +57,12 @@ import { Realm, LayerUserInfo } from 'shared/dao/types'
 import { Store } from 'redux'
 import { RootState } from 'shared/store/rootTypes'
 import { store } from 'shared/store/store'
-import { setCatalystRealmCommsStatus, setCatalystRealm, markCatalystRealmFull } from 'shared/dao/actions'
+import {
+  setCatalystRealmCommsStatus,
+  setCatalystRealm,
+  markCatalystRealmFull,
+  markCatalystRealmConnectionError
+} from 'shared/dao/actions'
 import { observeRealmChange, pickCatalystRealm, changeToCrowdedRealm } from 'shared/dao'
 import { getProfile } from 'shared/passports/selectors'
 import { Profile } from 'shared/passports/types'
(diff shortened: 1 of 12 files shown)


kernel/packages/shared/comms/index.ts
```diff
@@ -650,6 +662,31 @@ export async function connect(userId: string) {
...
+
+  const candidates = getAllCatalystCandidates(store.getState())
+
+  const otherRealm = pickCatalystRealm(candidates)
+
+  const notificationMessage = realm
+    ? `Lost connection to ${realmString(realm)}, joining realm ${realmString(otherRealm)} instead`
+    : `Joining realm ${realmString(otherRealm)} instead`
+
+  notifyStatusTroughChat(notificationMessage)
```
BrianAmadori:
```suggestion
  notifyStatusThroughChat(notificationMessage)
```

kernel/packages/shared/dao/reducer.ts
```diff
@@ -89,6 +90,17 @@ export function daoReducer(state?: DaoState, action?: AnyAction): DaoState {
           }
         })
       }
+    case MARK_CATALYST_REALM_CONNECTION_ERROR:
+      return {
+        ...state,
+        candidates: state.candidates.map(it => {
+          if (it.catalystName === action.payload.catalystName) {
+            return { ...it, layer: { ...it.layer, elapsed: Number.MAX_SAFE_INTEGER } }
```
pablitar: Shouldn't we use a status for this? I know we already used the users count to mark a layer as full, but that made much more sense since the response signaled that state.

I suppose this also makes sense, since you are marking the response time of a catalyst as "infinite", which would be equivalent to an error state. But somehow I feel this is a little hacky.

In any case, it should work, so I'll still approve the PR.

pablitar (approved):
Awesome!

kernel/packages/shared/dao/reducer.ts
moliva: Thanks for the feedback. I added server connection status to candidates. Please review 😬

kernel/packages/shared/comms/index.ts
```diff
@@ -691,6 +703,31 @@ export async function connect(userId: string) {
...
+  const store: Store<RootState> = window.globalStore
+  const realm = getRealm(store.getState())
+
+  if (realm) {
+    store.dispatch(markCatalystRealmConnectionError(realm))
+  }
+
+  const candidates = getAllCatalystCandidates(store.getState())
+
+  const otherRealm = pickCatalystRealm(candidates)
```
pablitar, line 720: Should we update `pickCatalystRealm` to consider the candidate status?

moliva: Done!

pablitar (approved):
Approved again! :D

eordano (approved):

PR #648, opened 2020-02-26
chore: enable scene boundaries check

pravusjif:
* Refactored the SceneBoundariesChecker to be a single-instanced object again (instantiated and contained in the SceneController to be unique and replaced by the corresponding Debug Mode version in runtime)

* Optimized to run after a configurable time and taking into account the messaging processing CPU budget

-------------------------------------------------------
This can be tested in production in **ORG -131,-91**, where the grey building has a ramp to the right, to get to the 2nd floor, that ramp is outside the scene boundaries, in explorer.decentraland.org it can be seen, and in this branch build it cannot.


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneBoundariesController/SceneBoundariesChecker.cs
```diff
@@ -2,21 +2,76 @@
...
+        }
+
+        // TODO: Improve MessagingControllersManager.i.timeBudgetCounter usage once we have the centralized budget controller for our immortal coroutines
+        IEnumerator CheckEntities()
+        {
+            while (true)
+            {
+                if (timeBetweenChecks > 0 && Time.realtimeSinceStartup - lastChecktTime < timeBetweenChecks) yield return null;
+
+                foreach (var entity in entitiesToCheck)
```
BrianAmadori: It'd be nice to early test if `entitiesToCheck.Count > 0` before processing this foreach. Also, we should use the `using`/`MoveNext` construct as its a bit more memory efficient.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneBoundariesController/SceneBoundariesChecker.cs
```diff
@@ -2,21 +2,76 @@
...
 
-        public SceneBoundariesChecker(ParcelScene ownerScene)
+        HashSet<DecentralandEntity> entitiesToCheck = new HashSet<DecentralandEntity>();
+        HashSet<DecentralandEntity> checkedEntities = new HashSet<DecentralandEntity>();
+        Coroutine entitiesCheckRoutine = null;
+        float lastChecktTime;
+
+        public SceneBoundariesChecker()
+        {
+            entitiesCheckRoutine = SceneController.i.StartCoroutine(CheckEntities());
```
```suggestion
            entitiesCheckRoutine = CoroutineStarter.Start(CheckEntities());
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneBoundariesController/SceneBoundariesChecker.cs
```diff
@@ -2,21 +2,76 @@
...
+                lastChecktTime = Time.realtimeSinceStartup;
+
+                yield return null;
+            }
+        }
+
+        public void Stop()
+        {
+            if (entitiesCheckRoutine != null)
+                SceneController.i.StopCoroutine(entitiesCheckRoutine);
```
```suggestion
                CoroutineStarter.Stop(entitiesCheckRoutine);
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneBoundariesController/SceneBoundariesChecker.cs
```diff
@@ -53,7 +108,7 @@ void EvaluateMeshBounds(DecentralandEntity entity)
             Bounds meshBounds = Utils.BuildMergedBounds(entity.meshesInfo.renderers);
```
BrianAmadori: Optimization idea here: We could store the meshBounds here and if only position has changed just add the difference, only recalculating it if rotation/scaling changes.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneBoundariesController/SceneBoundariesChecker.cs
```diff
@@ -2,21 +2,76 @@
...
     {
-        protected ParcelScene scene;
+        [System.NonSerialized]
+        public float timeBetweenChecks = 1f;
 
-        public SceneBoundariesChecker(ParcelScene ownerScene)
+        HashSet<DecentralandEntity> entitiesToCheck = new HashSet<DecentralandEntity>();
+        HashSet<DecentralandEntity> checkedEntities = new HashSet<DecentralandEntity>();
+        Coroutine entitiesCheckRoutine = null;
+        float lastChecktTime;
```
```suggestion
        float lastCheckTime;
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -348,8 +337,8 @@ public DecentralandEntity CreateEntity(string id)
 
             newEntity.OnCleanupEvent += po.OnCleanup;
 
-            if (useBoundariesChecker || SceneController.i.isDebugMode)
-                newEntity.OnShapeUpdated += boundariesChecker.EvaluateEntityPosition;
+            if (SceneController.i.useBoundariesChecker)
+                newEntity.OnShapeUpdated += SceneController.i.boundariesChecker.AddEntity;
```
BrianAmadori: If an entity is removed we should remove it from this queue too or we might get weird errors.

BrianAmadori (changes requested):

pbosio (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Models/DecentralandEntity.cs
```diff
@@ -35,6 +35,27 @@ public GameObject meshRootGameObject
...
 
+            Vector3 lastPosition;
+            Vector3 lastScale;
+            Quaternion lastRotation;
+            Bounds _mergedBounds;
+            public Bounds mergedBounds
+            {
+                get
+                {
+                    if (meshRootGameObject.transform.position != lastPosition ||
```
BrianAmadori: We should only recalculate bounds if the rotation or scale changes. If position changes we just add the position delta to the bounds. This is taking advantage of the fact that most of the scenes just update entities position. Updating rotation and scale is more rare.

If we just check everything, the bounds will be recalculated almost always and will be the same as before.

unity-client/Assets/Scripts/MainScripts/DCL/Models/DecentralandEntity.cs
```diff
@@ -35,6 +35,27 @@ public GameObject meshRootGameObject
             public MeshFilter[] meshFilters;
             public List<Collider> colliders = new List<Collider>();
 
+            Vector3 lastPosition;
+            Vector3 lastScale;
+            Quaternion lastRotation;
+            Bounds _mergedBounds;
```
BrianAmadori:
```suggestion
            Bounds mergedBoundsValue;
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -348,8 +337,8 @@ public DecentralandEntity CreateEntity(string id)
 
             newEntity.OnCleanupEvent += po.OnCleanup;
 
-            if (useBoundariesChecker || SceneController.i.isDebugMode)
-                newEntity.OnShapeUpdated += boundariesChecker.EvaluateEntityPosition;
+            if (SceneController.i.useBoundariesChecker)
```
BrianAmadori, line 340: Maybe its a bit more accurate to null-check `SceneController.i.boundariesChecker` instead of this bool. If the bool changes purpose or the class isn't initialized correctly the nullity can slip away.

pbosio: > Maybe its a bit more accurate to null-check `SceneController.i.boundariesChecker` instead of this bool. If the bool changes purpose or the class isn't initialized correctly the nullity can slip away.

This suggestion may arise some issues with builder. I think that builder's will need to set SceneController.i.useBoundariesChecker to false (for example to avoid adding entities to the list of entities to check, cause builder do the boundaries check itself) but still need that the instance of boundariesChecker exist

BrianAmadori: That seems like an antipattern, why would the builder need such a confusing setup?

BrianAmadori: Ok, after discussing this personally we decided to move this suggestion to the future.

BrianAmadori (approved):

PR #650, opened 2020-02-27
fix: disabled MOTD and daily reward tooltip

pbosio:
Treasure hunt event has finished. So we need to disabled MOTD.
Daily rewards are not implemented. So we need to disabled daily reward tooltip in tutorial.


unity-client/Assets/Tutorial/Scripts/TutorialSteps/TutorialStep_ChatAndAvatarExpressions.cs
```diff
@@ -67,23 +67,20 @@ public override IEnumerator OnStepExecute()
             yield return avatarHUDTooltip.ShowAndHideRoutine();
             yield return WaitIdleTime();
 
-            yield return dailyRewardTooltip.ShowAndHideRoutine();
-            yield return WaitIdleTime();
+            //NOTE: dailyRewardTooltip disabled until daily reward is enabled
+            //yield return dailyRewardTooltip.ShowAndHideRoutine();
```
BrianAmadori: Can't we use a constant here instead of commenting the code?

BrianAmadori (approved):

pravusjif (approved):

PR #652, opened 2020-02-27
feat: Change realms automatically when you are alone

pablitar:
# What? 
This change adds a URL parameter that you can use to enable an experimental feature of changing the realm automatically when you are alone to a more crowded one.
* When you don't have any visible peer, every 40 seconds it checks to see if there is a more crowded realm
* If there is, then it tries to change to it
* If you are viewing people, it doesn't change realm

Currently in order to enable the feature you have to add the parameter `AUTO_CHANGE_REALM` to the URL. This is intended as an experimental feature to see if it improves the experience, and to check if the conditions are OK or they could be improved.

This PR also fixes a bug when changing realms, that would make you still see the people that were on the previous realm until they timed out. The same for them seeing you.


moliva (approved):
👏

moliva (approved):
LGTM!

PR #694, opened 2020-03-04
fix: faster pool cleanup + blocker issues

BrianAmadori:
Fixes:
- `CleanupPoolReferences` was causing big hiccups, managed to remove it while keeping state consistency.
- `PoolManager.HasPoolable` method was too slow and a source of hiccups.
- The spawn of surrounding blockers for big scenes was not working -- all were being spawned. Fixed it and added tests.

image


pbosio (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/BlockerHandler.cs
```diff
@@ -55,53 +53,57 @@ public void SetupBlockers(Vector2Int[] parcels, float height, Transform parent)
...
-                blockers.Add((pos.x, pos.y), blocker);
+                    blockers.Add(blockerPoolable);
+                }
             }
         }
 
         public void CleanBlockers()
         {
-            using (var it = blockers.GetEnumerator())
+            for (int i = 0; i < blockers.Count; i++)
```
pravusjif, line 101: We could cache this .Count before the loop as that accessor it is a bit slow

pravusjif (approved):

PR #702, opened 2020-03-05
fix: builder - dragging entity when trying to rotate camera

pbosio:
# What? 
Disable dragging when an entity is just pressed

# Why? 
In some cases is hard to rotate the camera without, unintentionally, moving some object


pravusjif (approved):

unity-client/Assets/Builder/Scripts/DCLBuilderObjectDragger.cs
```diff
@@ -57,38 +57,44 @@ private void OnSelectedObjectListChanged(Transform selectionParent, List<DCLBuil
...
+            }
+            isDragging = false;
         }
 
         private void OnMouseDrag(int buttonId, Vector3 mousePosition, float axisX, float axisY)
         {
-            if (buttonId == 0 && selectedEntities != null)
+            if (buttonId != 0 || selectedEntities == null)
+            {
+                return;
```
BrianAmadori, line 93: This means we aren't processing `OnMouseX` events for right clicks right? Maybe this is more clear if we use enums or at least a comment.

unity-client/Assets/Builder/Scripts/DCLBuilderObjectSelector.cs
```diff
@@ -152,14 +145,14 @@ private void OnMouseUp(int buttonId, Vector3 mousePosition)
             {
                 return;
             }
-            if (entityEnqueueForDeselectInfo.pressedEntity != null)
+            if (lastPressedEntityInfo.pressedEntity != null)
             {
-                if ((Time.unscaledTime - entityEnqueueForDeselectInfo.pressedTime) < MAX_SECS_FOR_CLICK)
+                if ((Time.unscaledTime - lastPressedEntityInfo.pressedTime) < MAX_SECS_FOR_CLICK)
```
BrianAmadori: Please add a comment for this as well, as its not very clear of the intention of having a time threshold.

BrianAmadori (approved):

PR #730, opened 2020-03-12
fix: rendering controller dependencies

### LINKED ISSUE #181: Fix RenderingController dependencies.
Right now, `RenderingController` script is littered by references to other scripts. This is making system decoupling very problematic. Right now we have many systems that can't be isolated in their own assemblies because of this arch. problem.

Having a smelly architecture that forces us to couple systems will lead to code spaguetti if not sanitized.

pbosio:
Closes #181


pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Rendering/RenderingController.cs
```diff
@@ -89,11 +89,21 @@ private void ActivateRendering_Internal()
         AssetPromiseKeeper_AB.i.useTimeBudget = true;
         AssetPromiseKeeper_AB_GameObject.i.useTimeBudget = true;
 
-        OnRenderingStateChanged?.Invoke(renderingEnabled);
+        CommonScriptableObjects.rendererState.Set(true);
 
         MemoryManager.i.CleanupPoolsIfNeeded(true);
         ParcelScene.parcelScenesCleaner.ForceCleanup();
```
BrianAmadori: What about those dependencies? We can't make them use the `rendererState` variable?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraController.cs
```diff
@@ -26,8 +26,7 @@ public class CameraController : MonoBehaviour
 
     private void Start()
     {
-        if (RenderingController.i != null)
-            RenderingController.i.OnRenderingStateChanged += OnRenderingStateChanged;
+        CommonScriptableObjects.rendererState.OnChange += OnRenderingStateChanged;
```
BrianAmadori, line 29: We shouldn't call `OnRenderingStateChanged` for a first time here? To account for if rendering state was changed before `Start()` call. This also applies to all components/classes that are suscribing to `OnChange` upon initialization.

BrianAmadori (approved):

PR #732, opened 2020-03-13
fix: blocking promises race condition lock issue

BrianAmadori:
When we enqueue new promises while blocked promises are being resolved, those new promises get stuck forever. This shouldn't happen.

Added test with mock classes so we make sure this doesn't happen with any of our asset managers.

The commit revert is included, so make sure to look at the individual commits to review the actual fix.

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Common/AssetPromiseKeeper.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Common/AssetPromiseKeeper.cs
@@ -4,6 +4,10 @@ using UnityEngine;
 
 namespace DCL
 {
+    public class AssetPromiseKeeper
+    {
+        public static float PROCESS_PROMISES_TIME_BUDGET = 0.006f;
+    }
     /// <summary>
     /// The AssetPromiseKeeper is the user entry point interface.
     /// It manages stuff like requesting something that's already being loaded, etc.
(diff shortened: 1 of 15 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Common/AssetPromiseKeeper.cs
```diff
@@ -245,6 +290,11 @@ private void LoadBlockedPromises(object loadedPromiseId)
                 promise.library = library;
                 promise.OnPreFinishEvent += CleanPromise;
```
pravusjif: has everything been pushed? i think in your computer, locally, this `CleanPromise()` was being called there without subscribing it to the `promise.OnPreFinishEvent`, is this OK here?

pravusjif (changes requested):
I think this branch is not up to date with your local changes, check the previous comment i made.
The bug (scenes that never finish loading) is still happening where we were testing:
https://play.decentraland.zone/branch/fix/blocking-promises-race-condition-lock-issue/index.html?ENV=org&position=21%2C-42&realm=unicorn-blue

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Common/AssetPromiseKeeper.cs
BrianAmadori: I found a cleaner way of implementing the fix, but in any case both approached missed a new case that I discovered today. Just added a few changes that should address all cases.

pravusjif (approved):

pbosio (approved):

PR #743, opened 2020-03-16
feat: message to request scenes info around a parcel

pbosio:
# What?
Added message from renderer to kernel to request scenes info around a parcel

# Why? 
We need to fetch the info of scenes around a specific parcel. Currently we can only request info around the parcel where the player is on and with a fixed amount of scenes around.

### DIFF
--- a/kernel/packages/shared/atlas/sagas.ts
+++ b/kernel/packages/shared/atlas/sagas.ts
@@ -12,10 +12,18 @@ import {
   FetchNameFromSceneJsonSuccess,
   marketData,
   QuerySceneName,
-  reportedScenes
+  reportedScenes,
+  ReportScenesAroundParcel
 } from './actions'
 import { getNameFromAtlasState, getTypeFromAtlasState, shouldLoadSceneJsonName } from './selectors'
-import { AtlasState, FETCH_NAME_FROM_SCENE_JSON, MarketEntry, SUCCESS_NAME_FROM_SCENE_JSON, MARKET_DATA } from './types'
+import {
+  AtlasState,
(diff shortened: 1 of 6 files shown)


kernel/packages/shared/atlas/sagas.ts
```diff
@@ -107,32 +117,51 @@ export function* checkAndReportAround() {
...
+  yield call(reportScenes, atlasState, targets)
+}
+
+export function* reportScenesAroundParcel(action: ReportScenesAroundParcel) {
+  let atlasState = (yield select(state => state.atlas)) as AtlasState
+  while (!atlasState.marketName['0,0']) {
+    yield take(MARKET_DATA)
+    atlasState = yield select(state => state.atlas)
+  }
+
```
moliva, line 143: Could we use the selector `getNameFromAtlasState` from `shared/atlas/selectors.ts` maybe? If not, please refactor to some `isMarketDataInitialized` selector.

kernel/packages/shared/atlas/sagas.ts
```diff
@@ -107,32 +117,51 @@ export function* checkAndReportAround() {
...
+  const targets = getScenesAround(action.payload.parcelCoord, action.payload.scenesAround, atlasState)
+  yield call(reportScenes, atlasState, targets)
+}
+
+function getScenesAround(parcelCoords: { x: number; y: number }, maxScenesAround: number, atlasState: AtlasState) {
```
moliva: Let's extract the parcelCoords type somewhere to be reused everywhere needed.

moliva (approved):
LGTM 👍

pablitar (approved):
Looks great!

PR #745, opened 2020-03-17
fix: network update position

moliva:
# What? 
- Bumps peer version to 0.1.35 with fix in network update to avoid disconnecting from a possibly _unknown_ peer
- Reintroduces changes to support network optimization protocol

### DIFF
--- a/kernel/package-lock.json
+++ b/kernel/package-lock.json
@@ -2541,9 +2541,9 @@
       "dev": true
     },
     "decentraland-katalyst-peer": {
-      "version": "0.1.33",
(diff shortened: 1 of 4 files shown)


pbosio (approved):

pablitar (approved):
LGTM!

Quick comment: I hope this fixes the crashes. If not, the next time instead of reverting we could try and avoid sending positionConfig. This should disable network optimization by position, and should work as before, but with the library bumped

moliva:
Noted! Let's merge this and see the outcome in `.zone` if possible.

PR #746, opened 2020-03-17
feat: navmap basic features

pravusjif:
the new navmap prefab is at `Assets/Scripts/MainScripts/DCL/Controllers/HUD/NavMap/NavMap`


pbosio:
Just one comment: Avatar Emotes icon is appearing over NavMap

Screen Shot 2020-03-24 at 00 46 33

pbosio (approved):

RominaMarchetti:
@pbosio that issue exceeds the nav map. We need to find a way to fix this because now canvas layers are ignoring their order (don't know why). This is noticeable in every full screen UI panel (with creators' scenes UI is the same :( ).

BrianAmadori (changes requested):
Found the following issues:

- When opening and closing the nav map, the minimap is displayed incorrectly (see screenshot below):
image

- Nav-map is not centered correctly when opening it.

- Scrolling input issues are all over the place, drag has too much momentum and sensivity is too low . (We can fix those issues as part of the scrolling issue, tho)

BrianAmadori (approved):

PR #747, opened 2020-03-17
feat: add comms tracking events

pablitar:
# What? 
This PR introduces events for sending tracking events of comms

# Why? 
To enable that data for diagnostic purposes

### DIFF
--- a/kernel/packages/shared/comms/v1/brokerWorldInstanceConnection.ts
+++ b/kernel/packages/shared/comms/v1/brokerWorldInstanceConnection.ts
@@ -452,4 +452,8 @@ export class BrokerWorldInstanceConnection implements WorldInstanceConnection {
     }
     return new SendResult(bytes.length)
   }
+
+  analyticsData() {
+    return
+  }
 }
--- a/kernel/packages/shared/comms/v1/brokerWorldInstanceConnection.ts
+++ b/kernel/packages/shared/comms/v1/brokerWorldInstanceConnection.ts
@@ -237,6 +237,10 @@ export class BrokerWorldInstanceConnection implements WorldInstanceConnection {
     this.connection.close()
   }
 
+  analyticsData() {
+    return
+  }
+
   private handleMessage(message: BrokerMessage) {
     const msgSize = message.data.length
 
@@ -451,9 +455,5 @@ export class BrokerWorldInstanceConnection implements WorldInstanceConnection {
       this.connection.sendUnreliable(bytes)
     }
     return new SendResult(bytes.length)
-  }
-
-  analyticsData() {
-    return
   }
 }
--- a/kernel/packages/shared/comms/v1/brokerWorldInstanceConnection.ts
+++ b/kernel/packages/shared/comms/v1/brokerWorldInstanceConnection.ts
@@ -238,7 +238,7 @@ export class BrokerWorldInstanceConnection implements WorldInstanceConnection {
   }
 
   analyticsData() {
-    return
+    return {}
   }
 
   private handleMessage(message: BrokerMessage) {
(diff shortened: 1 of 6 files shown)


kernel/packages/shared/comms/v2/LighthouseWorldInstanceConnection.ts
```diff
@@ -95,6 +95,16 @@ export class LighthouseWorldInstanceConnection implements WorldInstanceConnectio
     return this.cleanUpPeer()
   }
 
+  analyticsData() {
+    // For now, these are private. Next version of peer library will make them public
+    return {
+      // @ts-ignore
+      connectedPeers: this.peer.fullyConnectedPeerIds().map(it => it.slice(-6)),
```
moliva, line 103: Why do we send the last 6 characters of the id? Should we document this?

kernel/packages/shared/comms/v1/brokerWorldInstanceConnection.ts
```diff
@@ -237,6 +237,10 @@ export class BrokerWorldInstanceConnection implements WorldInstanceConnection {
     this.connection.close()
   }
 
+  analyticsData() {
+    return
```
moliva: Let's return an empty object here to avoid issues if statistics happen to be sent from local cli connection

kernel/packages/shared/comms/v2/LighthouseWorldInstanceConnection.ts
pablitar: This arised after speaking with @dnul. Since some of these lists could be very big, we should maybe use a part of the id. 

In general, we should do this all the way (in comms), but since that is more likely to cause some issues, I didn't want to go slicing ids in peer

kernel/packages/shared/comms/v1/brokerWorldInstanceConnection.ts
pablitar: I was thinking undefined could be part of the contract. I guess I could change the type to object | undefined

moliva: Alright! 👍

kernel/packages/shared/comms/v2/LighthouseWorldInstanceConnection.ts
moliva: Alright! 👍

kernel/packages/shared/comms/v1/brokerWorldInstanceConnection.ts
moliva: But wouldn't it result in a null reference error here https://github.com/decentraland/explorer/pull/747/files#diff-aa534620ff0363a557316b7564a11adeR676 ?

kernel/packages/shared/comms/v2/LighthouseWorldInstanceConnection.ts
pbosio: Can we add a comment there? to avoid any confusion in the future

pablitar: Done!

kernel/packages/shared/comms/v1/brokerWorldInstanceConnection.ts
pablitar: You are right. Just fixed it!

moliva: Amazing! 🌟

moliva (approved):
Great work!

pbosio (approved):
loving it!

PR #770, opened 2020-03-26
fix: chat accumulated messages performance

pravusjif:


kernel/packages/ui/avatar/chatWindow.ts
```diff
@@ -33,27 +33,18 @@ dcl.onEvent(event => {
...
-  messageText.textWrapping = true
-  messageText.outlineColor = Color4.Black()
+function updateMessagesLog(newMessage: MessageEntry) {
+  messagesLogText.value = ''
+  for (let i = 0; i < internalState.messages.length; i++) {
+    const currentMessage = internalState.messages[i];
+    const color = currentMessage.isCommand ? COMMAND_COLOR : 'white'
+
+    messagesLogText.value += `<color=${color}><b>${currentMessage.sender}:</b> ${currentMessage.message}</color>\n`
+  }
```
pbosio: Can we make this function parameterless? cause with this update we don't need that parameter any more

pbosio (approved):

kernel/packages/ui/avatar/chatWindow.ts
pravusjif: yeah good call

BrianAmadori (approved):

PR #774, opened 2020-03-26
chore: avoid external URL texture forced lowercase

pravusjif:
Removed the content provider `TryGetcontentURL()` usage for external URLs in DCLTexture.

This fix is for **externally-fetched URLs using the Texture resource from the SDK** (or DCLTexture in Unity)

How to test this:
- As there are no external-url textures in prod nor zone, at least you can check that everything is being loaded correctly, for example at https://play.decentraland.zone/branch/chore/AvoidExternalURLTextureForcedLowercase/index.html?position=60%2C-67
(in ZONE 60,67 there is a sprite loaded using the Texture resource from the SDK)


pbosio (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ContentProvider/ContentProvider.cs
```diff
@@ -109,9 +109,11 @@ public virtual string GetContentsUrl(string url)
             return null;
         }
 
-        public virtual bool TryGetContentsUrl_Raw(string url, out string result)
+        public virtual bool TryGetContentsUrl_Raw(string url, out string result, bool forceLowercaseURL = true)
         {
-            url = url.ToLower();
+            if (forceLowercaseURL)
+                url = url.ToLower();
```
BrianAmadori: There are a few problems with this approach, let me elaborate:

Remember that this is needed to match with the mappings dictionary, so this check should be in `BakeHashes` if we aim to use the dictionary. The thing is, if we are targeting external URLs, mappings shouldn't we used, so maybe the correct approach is to skip the `ContentProvider` usage altogether in that case.

BrianAmadori (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ContentProvider/ContentProvider.cs
pravusjif: Yeap that's better, i didn't know we could avoid using the TryGetContentURL()

BrianAmadori (approved):

PR #775, opened 2020-03-27
feat: new scene data fields + plumbing for navmap

BrianAmadori:
# What? 
This should add support for owner, thumbnails and description support in scene.json.

Also, `scene.json` plumbing was added to atlas sagas, so the `scene.json` data can be composed with the marketplace data before being sent to the nav map.

# Why? 
We need the new fields to support the navmap.


kernel/packages/shared/atlas/reducer.ts
```diff
@@ -30,68 +42,130 @@ export function atlasReducer(state?: AtlasState, action?: AnyAction) {
...
+  })
+
+  idToScene[landData.sceneId] = mapScene
+
+  return { ...state, tileToScene, idToScene }
+}
+
+function reduceDistrictData(state: AtlasState, action: AnyAction) {
+  state.hasDistrictData = true
+  return state
```
BrianAmadori: Remember to remove those comments. On a sidenote, district fetching should remain as a stub for now because we aren't used from unity side yet.

kernel/packages/shared/atlas/sagas.ts
```diff
@@ -64,130 +56,125 @@ function* fetchTiles() {
...
   })
-  window.unityInterface.UpdateMinimapSceneInformation(
-    typeAndNameKeys.map(key => ({
-      name: keyToTypeAndName[key].name,
-      type: keyToTypeAndName[key].type,
-      parcels: mapByTypeAndName[key]
-    }))
-  )
+
+  defaultLogger.info(`minimap`, minimapSceneInfoResult)
```
BrianAmadori: Remember to remove the logs (or put them under a disabled constant)

kernel/packages/shared/types.ts
```diff
@@ -155,24 +153,56 @@ export const TransparencyModes = {
...
+  origin?: string
+  projectId?: string
+}
+
 /// https://github.com/decentraland/proposals/blob/master/dsp/0020.mediawiki
-export interface IScene {
-  assets?: Record<any, string>
+export type SceneJsonData = {
+  display?: SceneDisplay
+  owner: string
```
BrianAmadori:
```suggestion
  owner?: string
```

kernel/packages/shared/atlas/sagas.ts
moliva: Sure enough. I will remove them, just added them for debugging during development.

kernel/packages/shared/utils.ts
```diff
@@ -0,0 +1,13 @@
+import { Vector2Component } from 'atomicHelpers/landHelpers'
+
+export function getTilesRectFromCenter(parcelCoords: Vector2Component, rectSize: number): string[] {
```
eordano: Might I suggest we name this file `getTilesRectFromCenter` instead of `utils.ts`? I'd rather have 1 function - 1 file rather than the generic "utils" name... sorry to bother you with this nit!

kernel/packages/shared/types.ts
```diff
@@ -353,88 +382,6 @@ export type WelcomeHUDControllerModel = HUDConfiguration & {
   hasWallet: boolean
 }
 
-export function normalizeContentMappings(
```
eordano, line 356: Thank you <3

kernel/packages/shared/selectors.ts
```diff
@@ -0,0 +1,108 @@
...
+      const file = key.toLowerCase()
+
+      ret.push({ file, hash: mappings[key] })
+    }
+  }
+
+  return ret
+}
+
+export function ILandToLoadableParcelScene(land: ILand): EnvironmentData<LoadableParcelScene> {
```
eordano, line 22: Are these functions used anywhere??? Can we delete them?

kernel/packages/shared/types.ts
```diff
@@ -1,12 +1,10 @@
-import { parseParcelPosition } from 'atomicHelpers/parcelScenePositions'
 import { Vector3Component } from '../atomicHelpers/landHelpers'
 import { QueryType } from 'decentraland-ecs/src/decentraland/PhysicsCast'
 
 export { Avatar, Profile, ColorString, WearableId, Wearable } from './profiles/types'
 
 export type MappingsResponse = {
   parcel_id: string
```
eordano, line 7: Maybe we can even go further and drop `parcel_id` and `root_cid` and use `sceneId` in preparation of using the content server v3.

kernel/packages/shared/atlas/selectors.ts
```diff
@@ -1,61 +1,90 @@
...
-  const key = `${x},${y}`
-  const name = state.sceneNames[key]
-    ? state.sceneNames[key]
-    : state.marketName[key] && state.marketName[key].name
-    ? state.marketName[key].name
-    : state.marketName[key] && state.marketName[key].estate_id && PlazaNames[state.marketName[key].estate_id]
-    ? PlazaNames[state.marketName[key].estate_id]
-    : EMPTY_PARCEL_NAME
-  if (name === 'interactive-text') {
+export function getSceneNameWithMarketAndAtlas(
```
eordano, line 62: Beautiful refactor. Would be nice to have some tests.

kernel/packages/shared/atlas/sagas.ts
```diff
@@ -39,14 +36,10 @@ export function* atlasSaga(): any {
   yield fork(fetchDistricts)
   yield fork(fetchTiles)
 
-  yield takeEvery(SCENE_START, querySceneName)
   yield takeEvery(SCENE_LOAD, checkAndReportAround)
-  yield takeEvery(FETCH_NAME_FROM_SCENE_JSON, fetchName)
 
-  yield takeLatest(RENDERER_INITIALIZED, reportScenesAround)
```
eordano, line 46: Could you explain a little why this is no longer needed?

kernel/packages/shared/atlas/reducer.ts
```diff
@@ -30,68 +41,137 @@ export function atlasReducer(state?: AtlasState, action?: AnyAction) {
...
+  return { ...state, idToScene }
+}
+
+function reduceSuccessDataFromSceneJson(state: AtlasState, landData: ILand) {
+  const tileToScene = { ...state.tileToScene }
+  const idToScene = { ...state.idToScene }
+
+  let mapScene: MapSceneData = { ...state.idToScene[landData.sceneId] }
+
+  // NOTE(Brian): this code is for the case in which market data comes first (most likely always)
```
eordano: Not a big fan of this explanation. Is there any way we can make this go away, some way the code makes this evident?

kernel/packages/shared/atlas/reducer.ts
```diff
@@ -1,25 +1,36 @@
...
+  sceneId: '',
+  name: '',
+  type: 0,
+  estateId: 0,
+  sceneJsonData: undefined,
+  alreadyReported: false,
+  requestStatus: 'loading'
 }
 
 export function atlasReducer(state?: AtlasState, action?: AnyAction) {
```
eordano, line 31: if we switch `AnyAction` to `QuerySceneData|FetchDataFromSceneJsonSuccess|FetchDataFromSceneJsonFailure| ReportedScenes` I believe the type casts on each `case` are not going to be necessary

eordano (approved):
A lot of nits and general suggestions. Great work 🍻

kernel/packages/shared/selectors.ts
moliva: It seems we are using this from the preview and builder.

kernel/packages/shared/atlas/sagas.ts
BrianAmadori: This is because now we aren't using the scenes lifecycle to fetch the scene data, as we need to take data from scenes that aren't even loaded. We just fetch them. So, with this decoupling the actions binding the lifecycle to the atlas saga were removed.

moliva: It is not needed as the events that trigger the report of minimap information are wired to scene lifecycle events which are fired after the renderer is initialized.

PR #776, opened 2020-03-27
feat: AudioStream component

pbosio:


moliva (approved):
Sounds good to me! 👍

kernel/packages/unity-interface/dcl.ts
```diff
@@ -116,6 +116,35 @@ const positionEvent = {
...
 
+/////////////////////////////////// AUDIO STREAMING ///////////////////////////////////
+
+const audioStreamSource = new Audio()
+
+teleportObservable.add(() => {
+  audioStreamSource.pause()
+})
+
+async function setAudioStream(url: string, play: boolean, volume: number) {
```
BrianAmadori, line 127: This code shouldn't be more maintainable if we move this method to another file? 
`dcl.ts` should try to contain the basic messaging plumbing only.

pravusjif (approved):

kernel/packages/unity-interface/dcl.ts
```diff
@@ -116,6 +116,35 @@ const positionEvent = {
   mousePosition: Vector3.Zero()
 }
 
+/////////////////////////////////// AUDIO STREAMING ///////////////////////////////////
+
+const audioStreamSource = new Audio()
+
+teleportObservable.add(() => {
+  audioStreamSource.pause()
```
BrianAmadori, line 124: As the audio is not spatialized, it wouldn't be more wise to listen to the player's position and stop it when the player is out of the scene? if not we are going to listen to the stream from very far away and it can be an unpleasant experience.

What happens if we teleport to coordinates inside the `AudioStream` range, it first pauses and then plays as intended?

What happens if we have 2 `AudioStream` components from 2 different scenes? They wouldn't override each other?

kernel/packages/unity-interface/dcl.ts
pbosio: This is all we need for now. I will totally agree to move it to other file in the future if we need to add more features or extend it functionality

kernel/packages/unity-interface/dcl.ts
pbosio: > As the audio is not spatialized, it wouldn't be more wise to listen to the player's position and stop it when the player is out of the scene? if not we are going to listen to the stream from very far away and it can be an unpleasant experience.

This is how it actually works. Audio can only be played if player is inside the same scene as the AudioStream component.
That subscription to the teleport event is just to make sure that, no matter what, the audio stops when player teleports.

> What happens if we teleport to coordinates inside the AudioStream range, it first pauses and then plays as intended?

Yes

> What happens if we have 2 AudioStream components from 2 different scenes? They wouldn't override each other?

As mentioned before, the audio is only played if the user is in the same scene.. if player move out of scene audio stops, if player move to a scene with other AudioStream, the audio of that scene is played

pbosio: Checks if audio can or cannot be played are on the unity side of the component in `UpdatePlayingState` method of `DCLAudioStream` class

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioStream.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            CommonScriptableObjects.playerCoords.OnChange -= OnPlayerCoordsChanged;
+            CommonScriptableObjects.rendererState.OnChange -= OnRendererStateChanged;
+            StopStreaming();
+        }
+
+        private bool AreCoordsInsideComponentScene(Vector2Int coords)
+        {
+            if (scene == null) return false;
+            return scene.parcels.Contains(coords);
```
BrianAmadori, line 47: We can check component's scene id against `SceneController.currentSceneId` instead of this expression.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioStream.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        private void UpdatePlayingState(bool forceStateUpdate)
+        {
+            if (gameObject.activeInHierarchy)
```
BrianAmadori: Too much `if...else` and flags spaguetti here. There isn't a way to improve this logic? Is a bit hard to follow and will be a struggle to modify it in the future.

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioStream.cs
eordano: I think the logic is pretty easy to follow. I would early-exit (`if (!gameObject.activeInHierarchy) return`) but other than that it's pretty easy to follow the next 4 ifs

pbosio: > I think the logic is pretty easy to follow. I would early-exit (`if (!gameObject.activeInHierarchy) return`) but other than that it's pretty easy to follow the next 4 ifs

I've have already made those changes. Waiting for CI tests and merge? :)

eordano (approved):

PR #819, opened 2020-04-15
feat: chat remake

BrianAmadori:
*NOTE:* To enable the new chat, add the `USE_NEW_CHAT` query parameter when opening Explorer (e.g. https://explorer.decentraland.zone/branch/feat/chat-remake/index.html?ENV=org&USE_NEW_CHAT ).

### Taskbar
The `Taskbar` element is intended to be used for window stacking, like i.e. the windows OS. This element for now is containing the `Chat World Window` but the idea is that we are going to reuse it for keeping track of the private messages windows, friend list windows, etc. 

The emotes button is also going to be kept here in the future.

### World Chat Window
This window has a `Chat Widget` as is in charge of passing the incoming messages from kernel to it. It also handles the `World Messages` and `Private Messages` tab, preview mode logic, etc. 

Preview mode is the mode in which the window set itself as transparent when out of focus, etc.

### Chat Widget
This widget is intended as general purpose, and the idea is we keep reusing it when we implement the friends window feature. It should be agnostic to specific windows behaviour or kernel messages. 

## UI elements details:
`HUD/ChatWidgetHUD/Resources/Chat Widget.prefab`
This is the generic chat widget that's going to be used for world and private messaging windows. 

The `Chat Widget` has a the following elements: 
* `ScrollRect` element
* Chat entries container: this is where `Chat Entry` elements will be added. For design purposes is possible to enparent `Chat Entry` elements manually here.
* Text input element

`HUD/ChatWidgetHUD/Resources/Chat Entry.prefab`
Represents a single message.

`HUD/WorldChatWindow/Resources/World Chat Window.prefab`
This is the main window for world chat. Includes a `Chat Widget` and the top bar with world/private filter buttons.

`HUD/TaskBar/Resources/Taskbar.prefab`
Currently this prefab contains a `World Chat Window` so it can be used have a visual glimpse of all the elements together.


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
```diff
@@ -0,0 +1,67 @@
...
+            var result = TrimAndSortChatMessages(view.entries.Select((x) => x.message).ToList());
+            view.RepopulateAllChatMessages(result);
+        }
+    }
+
+    public void FilterByType(ChatMessageType type)
+    {
+        var result = ChatController.i.entries.Where((x) => x.messageType == type).ToList();
+        result = TrimAndSortChatMessages(result);
+        view.RepopulateAllChatMessages(result);
```
pravusjif, line 45: Is it necessary to repopulate the entries list consantly? can't we just disable the gameobject of the messages we don't want to show (the ui would adapt automatically) and re-enable them when we need to show them again? for example when filtering messages.
Because right now all this sorting and traversing will happen every time we get a message, after we've reached the MAX_CHAT_ENTRIES right?

the next comment is related, please check that one also

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
```diff
@@ -0,0 +1,67 @@
...
+    {
+        var result = messages;
+
+        result.Sort((x, y) => { return x.timestamp > y.timestamp ? 1 : -1; });
+
+        if (result.Count > MAX_CHAT_ENTRIES)
+        {
+            int entriesToRemove = (result.Count - MAX_CHAT_ENTRIES);
+            result.RemoveRange(0, entriesToRemove);
+        }
```
pravusjif, line 58: I think instead of re-sorting and removing entries, every time we get to the MAX_CHAT_ENTRIES we could just remove the oldest message in the list/queue, reconfigure it, and re-add it at the back of the list/queue again and that would be it. (maybe using a queue instead of a list could make it easier as well)

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
```diff
@@ -0,0 +1,67 @@
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine.Events;
+using ChatMessage = ChatController.ChatMessage;
+using ChatMessageType = ChatController.ChatMessageType;
+public class ChatHUDController : IDisposable
```
pravusjif, line 8: I'm a bit confused as to why we have this class and `WorldChatWindowHUDController`, is the 'ChatHUD' a UI element inside the 'WorldChatWindowHUD'?

BrianAmadori: Yes, the idea is that the ChatHUD can be put inside the friends window when we get to it. I arranged the ChatHUD components so they are agnostic of the context of the HUD element.

The WorldChatWindow has the ChatHUD and the tabs for filtering world/PM, close button, etc.

pravusjif: I just saw this info in the description, i'm resolving this.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
BrianAmadori: This is a good idea, i'm going to give it a shot.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
BrianAmadori: Going to give a shot of optimizing it as you suggested.

pravusjif (approved):

PR #821, opened 2020-04-16
fix: pointer events hover on invisibles

pravusjif:
Tied pointer events interaction and hover feedback to its shape visibility.

This can be tested at 
https://play.decentraland.zone/branch/fix/PointerEventHoverOnInvisibles/index.html?ENV=org&position=53,3

when picking the keycard at the back of the building, now the interaction disappears after grabbing the object.

**As mentioned in the ZenHub issue:**
This "bug" keeps happening in interactions like the ones at Quest of the alcoholic scarecrow and Magic world, because **in those cases the pointer event that triggers the interaction is in a separate entity than the rendered object**... so in those cases rather than a bug is a **misuse of the SDK/Smart items**.


unity-client/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/OnPointerEventColliders.cs
```diff
@@ -68,6 +68,21 @@ Collider CreatePointerEventCollider(Renderer renderer)
             return meshCollider;
         }
 
+        public bool IsColliderInCollection(Collider targetCollider)
```
BrianAmadori: This method is not used, we should keep it?

unity-client/Assets/Scripts/Tests/CursorControllerTests/CursorControllerTests.cs
```diff
@@ -88,6 +90,79 @@ public IEnumerator OnPointerHoverFeedbackIsDisplayedCorrectly()
...
+
+            var cursorController = GameObject.FindObjectOfType<CursorController>();
+
+            // Check cursor shows normal sprite
+            Assert.AreEqual(cursorController.cursorImage.sprite, cursorController.normalCursor);
+
+            DCLCharacterController.i.PauseGravity();
+            DCLCharacterController.i.SetPosition(new Vector3(8, 1, 7f));
+
+            var cameraController = GameObject.FindObjectOfType<CameraController>();
```
BrianAmadori: `cameraController` is already part of `TestsBase`. Maybe this line isn't necessary.

unity-client/Assets/Scripts/Tests/CursorControllerTests/CursorControllerTests.cs
```diff
@@ -88,6 +90,79 @@ public IEnumerator OnPointerHoverFeedbackIsDisplayedCorrectly()
...
+
+            // Rotate camera towards the interactive object
+            var cameraRotationPayload = new CameraController.SetRotationPayload()
+            {
+                x = 45,
+                y = 0,
+                z = 0
+            };
+
+            cameraController.SetRotation(JsonConvert.SerializeObject(cameraRotationPayload, Formatting.None, new JsonSerializerSettings()
```
BrianAmadori: This can be heavily resumed as:
`cameraController.SetRotation(45, 0, 0)`

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/OnPointerEventColliders.cs
pravusjif: I'll remove it and if we change our policy of triggering pointer events with normal colliders in the future we can re-implement that utility method as it's simple enough.

PR #832, opened 2020-04-23
fix: drop external contracts dependency

eordano:
Since these things don't change too often; I'd rather have them in the client rather than block on an external fetch


moliva (dismissed):
Approving for the hotfix.
We might need to still fetch the contracts from the cloud and use the local information as a failsafe.

moliva (approved):
LGTM.

PR #844, opened 2020-04-27
feat: optimize circleci builds

eordano:
This commit introduces three changes to the CircleCI script:

1. Using `md5sum` instead of `shasum` reduces the time to detect
     whether the source code changed or not from about 1 minute to
     about 2 seconds (!!).
2. Before that, I was trying to reduce the time to calculate that
     checksum by only looking at the relevant files (that's why the
     change on `find` introducing all the `-o -iname \*.something`)
     In hindsight, the `md5sum` change makes this innecessary; the
     tradeoffs are:
      - If we miss some extension that *should* trigger the build,
        then we might get passing builds that should have stopped.
        But how frequent is a change that *only* changes a file
        that we might miss?
        * From the Unity side, we're pretty safe: anything .meta
          triggers a build cache invalidation
        * From the kernel side, the number of extensions on the
          folder is pretty small -- running this `find`:
```
  find kernel/packages -type f -exec basename {} +
    | grep -oE '[^.]+$' # show only extensions
    | sort | unique     # sort & unique
```
Gives a total of 9 extensions on my machine:
 ```
           - DS_Store
           - LICENSE
           - gitignore
           - js
           - json
           - md
           - proto
           - sh
           - ts
```
In order to determine which files to cache in Unity, I used the following script to list extensions:
```
find . -type f \( -not -path '*Library/*' \) # anything in lib can be ignored
  -exec basename {} + # take only the file name
  | grep -oE '[^.]+$'     # match only the ending extension
  | sort | uniq # sort and unique
```
And then to list the biggest files:
```
find . -type f \( -not -path '*Library/*' \) # same find as before
 \( -exec ls -l {} \; \)   # `ls -l` to get more info on each file
 | awk '{print $5 " " $9}' # print columns 5 and 9 of `ls` (size & name)
 | grep -E '^[0-9]{6,10}'  # grep only lines starting with 6 digit numbers
```

3. New cache: `Library` folder. Dropped because it makes the
    cache build fail.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,10 +93,6 @@ jobs:
           command: |
             find Assets -type f \( -not -path '*Plugins*' \) \( -iname \*.cs -o -iname \*.meta -o -iname \*.xml -o -iname \*.shader -o -iname \*.prefab -o -iname \*.yml -o -iname \*.mat -o -iname \*.json -o -iname \*.js -o -iname \*.jspre  -o -iname \*.jslib  -o -iname \*.hlsl  -o -iname \*.asmdef  -o -iname \*.csproj  \) \( -exec md5sum "$PWD"/{} \; \) | sort > ../.unitysources-checksum
       - restore_cache:
-          name: Restore unity Library Cache
-          keys:
-            - unity-libcache2-{{ .Branch  }}
-      - restore_cache:
           name: Restore decentraland-renderer if exists
           keys:
             - unity-build-{{ checksum "../.unitysources-checksum" }}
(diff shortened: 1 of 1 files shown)


eordano:
TL;DR -- Shaving 1 minute out from the hashing of source files; not related to caching unity library because it didn't work.

image
image

pravusjif (approved):
Looks good, nice optimizations to gain some speed on the CI.
Too bad we can't cache the unity project asset library yet, we should analyze if there is any way to do that on the CI at all, in the future.

moliva (approved):
LGTM.

nchamo (approved):
Great work!

PR #846, opened 2020-04-27
feat: chat private messaging ui

BrianAmadori:
Test check-list:
- Avatars profile card should open correctly
- Tab button animations when going from world to private should work correctly
- When clicking the private tab, all messages except the private ones should dissapear (so, all of them for now).
- Private messages are testeable using a small hack from Unity editor. Just right-click the `ChatController` component attached to the `SceneController` object and select `Fake Private Message` from the context menu. A few private messages should appear in the UI.


unity-client/Assets/Scenes/InitialScene.unity
```diff
@@ -307,7 +307,7 @@ PrefabInstance:
     - target: {fileID: 114754758112584508, guid: b387035aa46399445bdc7951a41d5213,
         type: 3}
       propertyPath: debugScenes
-      value: 0
+      value: 1
```
pravusjif: let's disable this toggle

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/WorldChatWindowHUDController.cs
```diff
@@ -2,23 +2,23 @@
...
     private ChatHUDController chatHudController;
     public WorldChatWindowHUDView view;
 
-    private string userName;
-
     private IChatController chatController;
     private IMouseCatcher mouseCatcher;
 
     internal bool resetInputFieldOnSubmit = true;
+    private int invalidSubmitLastFrame = 0;
```
pravusjif, line 17: are we using this somewhere?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/WorldChatWindowHUDController.cs
```diff
@@ -62,7 +57,33 @@ public void Dispose()
...
+
+    void OnEnablePrivateTab()
+    {
+        view.chatHudView.CleanAllEntries();
+        var result = chatController.GetEntries().Where((x) => x.messageType == ChatMessage.Type.PRIVATE).ToList();
+
+        foreach (var v in result)
+        {
+            OnAddMessage(v);
+        }
```
pravusjif, line 80: Can't we just hide/show the filtered messages by toggling their gameobjects? it would be more performant i think, also if we keep the private and non-private messages separately we would avoid traversing the whole entries list every time the tab is clicked.

BrianAmadori: I don't think we need performance improvements on something so sparse as clicking on a tab, also this won't work well with this scenario:

- We have 30 private messages in 100 world chat entries (100 because its trimmed) but actually the private history is of 150.
- We change to private tab
- 30 messages would be shown, when with the 100 cap we should repopulate the list with 100 private messages.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/WorldChatWindowHUDController.cs
BrianAmadori: Yep, is to deal with a bug where `onSubmit` and the return key callback are called on the same frame, opening and closing the window again. We have an equivalent fix in the `UITextInput` UI component.

unity-client/Assets/Scenes/InitialScene.unity
BrianAmadori: done

RominaMarchetti:
About the  checklist: everything  seems to be working fine!

The only issue I've detected was an overlap  between messages. I took some screenshots when I saw it, then I've tried to reproduce it but I just couldn't. See file attached,  please.
Screen Shot 2020-04-28 at 22 43 23

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/WorldChatWindowHUDController.cs
pravusjif: We could load the previous private messages at the beginning of the explorer and have the correct private messages in a separate collection, but OK we can tackle this if we notice problems

pravusjif (approved):

RominaMarchetti (approved):

PR #849, opened 2020-04-27
feat: friends HUD

pravusjif:


RominaMarchetti:
-Friend **menu button** missing on the row (it should only appear on hover)
-**Add Friend** button missing on the search bar (only visible if a player name is written)
-A friend request sent should only  display  a **Cancel CTA**.

PR #850, opened 2020-04-27
feat: video player

pbosio:
video streaming MVP

can be tested here: https://play.decentraland.zone/branch/feat/video-player/index.html?ENV=zone&position=52%2C-58&realm=heimdallr-amber


kernel/packages/decentraland-ecs/src/decentraland/Components.ts
```diff
@@ -848,3 +851,72 @@ export class OnAnimationEnd extends OnUUIDEvent<'onAnimationEnd'> {
...
+    super()
+    this.url = url
+  }
+}
+
+/**
+ * @public
+ */
+@DisposableComponent('engine.VideoTexture', CLASS_ID.VIDEO_TEXTURE)
+export class VideoTexture extends ObservableComponent {
```
pravusjif, line 873: Can't we make `VideoTexture` extend from `Texture` ?

pravusjif (approved):

kernel/packages/decentraland-ecs/src/decentraland/Components.ts
pbosio: Nup... I tried that, apparently you can't extend components

BrianAmadori: Why not? I did this extensively for UI components.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Video/Tests/VideoTests.cs
```diff
@@ -0,0 +1,190 @@
...
+        protected override IEnumerator SetUp()
+        {
+            yield return base.SetUp();
+            DCLVideoTexture.isTest = true;
+        }
+
+        [UnityTest]
+        public IEnumerator VideoTextureIsAttachedAndDetachedCorrectly()
+        {
+            // TEST: DCLVideoTexture is created correctly
```
BrianAmadori: Instead of using comments, those test segments should be promoted to a full test each

BrianAmadori (changes requested):

kernel/packages/decentraland-ecs/src/decentraland/Components.ts
pbosio: > Why not? I did this extensively for UI components.

That's because UIShape is not really a component (It's missing` @DisposableComponent`)
and even if extending `Texture` was possible we need to be able to discriminate them, cause we don't want to enable to add a `VideoTexture` in every component that has a `Texture` field... like for example Material's `bumpTexture` or `refractionTexture`

BrianAmadori (approved):

PR #864, opened 2020-05-05
feat: open external url from sdk

pbosio:
Add to the SDK the function `openExternalURL(url)` to opening a link in a new tab and should only work when called from inside a pointerEvent


pravusjif (approved):

pablitar (approved):
Looks great!

kernel/packages/scene-system/scene.system.ts
```diff
@@ -219,6 +225,18 @@ export default class GamekitScene extends Script {
...
 
+        openExternalUrl(url: string) {
+          if (that.allowOpenExternalUrl) {
+            that.events.push({
+              type: 'OpenExternalUrl',
+              tag: '',
+              payload: url
+            })
+          } else {
+            this.error('openExternalUrl cant only be used inside a pointerEvent')
```
moliva, line 236: ```            this.error('openExternalUrl can only be used inside a pointerEvent')```

kernel/packages/scene-system/scene.system.ts
```diff
@@ -583,4 +601,15 @@ export default class GamekitScene extends Script {
 
     return data
   }
+
+  private isPointerEvent(event: any): boolean {
+    switch (event.type) {
+      case 'uuidEvent':
+        if (event.data.payload as InputEventResult) {
```
moliva, line 608: Why the `as` here?

PR #865, opened 2020-05-07
feat: chat persistent whisper and reply

pravusjif:
Added /whisper or /w persistence on usage and /r command functionality


BrianAmadori (commented):
I'd be nice to have at least a single test to cover this behaviour.

PR #878, opened 2020-05-08
feat: chat new preview mode

BrianAmadori:
link to test: https://play.decentraland.zone/branch/feat/chat-new-preview-mode/index.html?ENV=org&USE_NEW_CHAT

## WHAT?
- Chat window no longer can be fully closed, is either in normal or in "preview" mode.
- Removed World/Private tabs
- Removed open chat tooltip, as now its always open.
- In preview mode: 
    - Only messages with less than 10 seconds of lifetime are shown (this time can be changed in `Chat Entry` prefab)
    - When a message gets older than 10 seconds, it starts to fade out (fade out anim also tweakeable from the `Chat Entry` prefab)


pbosio (approved):

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatEntry.cs
```diff
@@ -66,6 +72,39 @@ public void Populate(Model chatEntryModel)
...
+            this.enabled = false;
+            return;
+        }
+
+        this.enabled = true;
+    }
+
+    private void Update()
+    {
+        double fadeTime = (double)(model.timestamp / 1000.0) + timeToFade;
```
pravusjif, line 96: I'd love to have an escape line here to avoid calculating the time stuff if the entry is fully visible, etc. but not sure we can do it.

pbosio: > I'd love to have an escape line here to avoid calculating the time stuff if the entry is fully visible, etc. but not sure we can do it.

When `SetFadeout` is set to disable it disable the component, so there are no calls to `Update`

PR #890, opened 2020-05-11
feat: private chat window

pravusjif:


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PrivateChatWindow/PrivateChatWindowHUDController.cs
```diff
@@ -0,0 +1,89 @@
...
+public class PrivateChatWindowHUDController : IHUD
+{
+    private ChatHUDController chatHudController;
+    public PrivateChatWindowHUDView view;
+
+    private IChatController chatController;
+
+    internal bool resetInputFieldOnSubmit = true;
+
+    public void Initialize(IChatController chatController, IMouseCatcher mouseCatcher)
```
BrianAmadori: Here, we lack some kind of "user ownership" of the window. We should define some interface that'd allow the FriendsHUD/TaskbarHUD to re-populate the window properly.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PrivateChatWindow/PrivateChatWindowHUDController.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        if (resetInputFieldOnSubmit)
+        {
+            view.chatHudView.ResetInputField();
+            view.chatHudView.FocusInputField();
+        }
+
+        var data = new ChatMessage()
+        {
+            body = msgBody,
```
BrianAmadori: With some kind of "user ownership" we could just append the `/w <user> ` to the message and the rest of the private message fields in `ChatMessage` so it is sent correctly.

BrianAmadori (changes requested):

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PrivateChatWindow/PrivateChatWindowHUDController.cs
```diff
@@ -51,21 +75,22 @@ public void SendChatMessage(string msgBody)
 
         var data = new ChatMessage()
         {
-            body = msgBody,
+            body = $"/w {conversationUserName} " + msgBody,
             sender = UserProfile.GetOwnUserProfile().userId,
```
BrianAmadori, line 78: We don't need to set the `recipient` here with the current userId?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PrivateChatWindow/PrivateChatWindowHUDController.cs
```diff
@@ -51,21 +75,22 @@ public void SendChatMessage(string msgBody)
...
 
     public void SetVisibility(bool visible)
     {
+        if (view.gameObject.activeSelf == visible) return;
+
         view.gameObject.SetActive(visible);
 
         if (visible)
-        {
             view.StartCoroutine(ForceLayoutDelayed());
```
BrianAmadori: Please use `Utils.ForceUpdateLayout` here. Also, this should be under view's `OnEnable` as in `FriendsTabViewBase`.

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PrivateChatWindow/PrivateChatWindowHUDController.cs
pravusjif: This way, we don't need to set the receiver because it's like sending a global message with the whisper command.

Can we send private messages to kernel without using the whisper commands? I don't think so, but if we could, then we should set the receiver as you say, and send the message in a pipeline that accepts private messages without parsing the /w command

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PrivateChatWindow/PrivateChatWindowHUDController.cs
pravusjif: done

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PrivateChatWindow/PrivateChatWindowHUDController.cs
BrianAmadori: Oh right, this its already handled in the ` /w`, then kernel loops back with the message containing the proper recipient/sender values. nvm.

BrianAmadori (approved):

PR #910, opened 2020-05-14
fix: flip picture frame image

pbosio:
Picture frame images were horizontally mirrored


pravusjif (approved):

pravusjif:
So the problem was in the fbx? I should have known, those artists.... XD

moliva (approved):

PR #916, opened 2020-05-14
feat: VideoTexture support video from scene's file

pbosio:
# What? 
support loading video files from scene 
mp4, mov, ogg, webm


pravusjif (approved):
Can we have a little test scene with a small video to know exactly how this is supposed to be used?

pbosio:
> Can we have a little test scene with a small video to know exactly how this is supposed to be used?

I will add a test scene about in next PR when adding new functionalities to the video... I sent this PR so I don't send a big PR later

BrianAmadori (approved):

PR #928, opened 2020-05-19
[Client Frontend & Backend] - Replace Swanky rarity to Rare

RominaMarchetti:
- [x] Frontend: all swanky prefabs renamed and labels edited.

To checkout rarity go to the avatar editor and open a green NFT info card: instead of Swanky you should read Rare.

---------------------
- [x] Backend


kernel/packages/shared/profiles/sagas.ts
```diff
@@ -150,6 +161,7 @@ export function* initialLoad() {
       const catalog = collections
         .reduce((flatten, collection) => flatten.concat(collection.wearables), [] as Wearable[])
         .map(overrideBaseUrl)
+        .map(overrideSwankyRarity)
```
nachomazzara, line 165: Is this because of the wearable-api or stored profiles?

moliva: This data comes from the wearable-api. The idea is to avoid issues in this transition.

nachomazzara: Nice! I've implemented the change, but not deployed (will do on Tuesday) in case you want to add an issue to remove this override later.

moliva: Good point, will do that then 😬

eordano (approved):

PR #930, opened 2020-05-20
feat: NFT info prompt

pbosio:
test-scene `-103.102.nft`

Screen Shot 2020-06-04 at 11 39 37


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NFTPromptHUD/NFTPromptHUDView.cs
```diff
@@ -0,0 +1,306 @@
...
+    IWrappedTextureAsset imageAsset = null;
+
+    bool backgroundColorSet = false;
+    string marketUrl = null;
+
+    private void Awake()
+    {
+        buttonClose.onClick.AddListener(Hide);
+        buttonCancel.onClick.AddListener(Hide);
+        buttonOpenMarket.onClick.AddListener(OpenMarketUrl);
```
sandrade-dcl, line 56: It would be a good practice to unsubscribe these subscriptions somewhere along the monobehaviour life-cycle (in this case in the OnDestroy method).

sandrade-dcl (changes requested):
Take a look to my suggestion about unsubscriptions and see if it make sense in this case. Apart from that, everything looks fine! :)

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NFTPromptHUD/NFTPromptHUDView.cs
pbosio: In this case, buttons being children of the GameObject with that MonoBehavior attached, If the GameObject is destroyed that means the buttons will be destroyed as well... so I don't think unsubscribing to events is really necessary

sandrade-dcl: Ok! Then there is no problem! Approved ;-)

sandrade-dcl (approved):

moliva (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -712,6 +715,9 @@ public bool ProcessMessage(MessagingBus.QueuedSceneMessage_Scene msgObject, out
                     case MessagingTypes.OPEN_EXTERNAL_URL:
                         OnOpenExternalUrlRequest?.Invoke(scene, payload.OpenExternalUrl.Url);
                         break;
+                    case MessagingTypes.OPEN_NFT_DIALOG:
```
pravusjif, line 718: I see we don't have these 2 cases in our WSSController.cs to have this working from the unity editor, we should add these, right?

pravusjif (approved):
Running local test scenes is broken right now... so I can't test this.

The code looks OK, just left a comment

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
pravusjif: OK it's not needed because it uses protobuff messaging

PR #945, opened 2020-05-26
feat: refactor `dcl.ts`

eordano:
Edit: Much smaller refactor!

Uses `globalDCL` in `shared` so the refactor of `unity-interface` is much leaner.

Easier to review commit-by-commit:

* `fix: drop all relative imports cross-packages`: Drops some references from `shared` to `unity-interface`. The idea is that only `entrypoints` should import from `unity-interface` and `shared` should be unity-dependency-free
* `chore: extract interfaces for unity and browser`: Creates the typings for `rendererInterface` and `browserInterface`
* `feat: use globalDCL instead of importing from unity-interface`: Renames any mention of `unityInterface` for `globalDCL.rendererInterface`
* `fix: drop window usage`: drops a number of uses of `window` or `globalThis` and uses `globalDCL`


pravusjif (approved):
Tried the build as if it were a release candidate and it worked fine!
The code looks pretty organized as well.

moliva (approved):

PR #950, opened 2020-05-26
feat: Allow the explorer to load and execute frames in background

pablitar:
# What? 
This patches requestAnimationFrame to enable it to work in background assuming it is not throttled. It also adds an inaudible audio to trick chrome into not thottling the tab (be sure to interact with the site to ensure that the audio starts playing)


moliva (approved):
Great work! 👏

pbosio (approved):

PR #955, opened 2020-05-28
fix: treat unavailable as online

moliva:
# What? 
Treat unavailable as online to avoid showing a user not actively sending status as offline.


marcosnc (approved):
LGTM

PR #956, opened 2020-05-28
fix: Conflicts pressing RETURN key on AddFriends window

sandrade-dcl:
# What? 
If we press the ENTER key after write a friend name in the Add Friends input field, the friend window is closed and the worl chat is opened.


pravusjif (approved):

BrianAmadori (approved):
Looks good, as a suggestion i'd improve the conditional legibility a bit by using local variables.

PR #959, opened 2020-05-28
fix: Hide JumpIn button when not receives realm info from a friend

sandrade-dcl:
# What? 
We should hide the JumpIn button when we don't receive the realm information.


pravusjif (approved):

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/FriendsHUD/Scripts/JumpInButton.cs
```diff
@@ -84,7 +84,9 @@ private void UpdateInfo(Vector2 coords, string realmServerName, string realmLaye
 
     private void ResfreshInfo()
```
BrianAmadori:
```suggestion
    private void RefreshInfo()
```

sandrade-dcl: Done! Thanks!

PR #963, opened 2020-05-29
fix: clean folders before AssetBundle migration

BrianAmadori:


unity-client/Assets/Batch/AssetBundleBuilder.cs
```diff
@@ -415,7 +436,7 @@ public bool DownloadAndConvertAssets(MappingPair[] rawContents, System.Action<Er
 
             startTime = Time.realtimeSinceStartup;
 
-            InitializeDirectoryPaths(false);
+            InitializeDirectoryPaths(true);
```
BrianAmadori, line 439: This is the actual change, sorry for the re-formatting.

PR #967, opened 2020-05-29
fix: mouse false lock when exiting emotes panel

pravusjif:
WHY?
When we click a UI button that is supposed to lock the mouse, on certain buttons (like the ones in the avatar expressions panel, including the overlay area to close the panel) we get a false mouse-locked state, because the request to the browser to lock the pointer is not done in the same frame as the user input (because we mostly use the onClick event in buttons, not the onPointerDown that would allow that)

WHAT?
* Implemented a UI.Button child class that has the onPointerDown event exposed
* Updated AvatarExpressions HUD and WelcomeHUD prefab buttons involved in locking the mouse to use the new Button class and the onPointerDown event


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/Button_OnPointerDown.cs
```diff
@@ -0,0 +1,15 @@
+﻿using UnityEngine;
+using UnityEngine.Events;
+using UnityEngine.EventSystems;
+using UnityEngine.UI;
+
+// In oder to lock/unlock the cursor in the same frame a button is clicked we need to use OnPointerDown (event not accesible in the regular Button)
```
sandrade-dcl: I would put this class comment in this way:

`/// <summary>`
`/// In oder to lock/unlock the cursor in the same frame a button is clicked we need to use OnPointerDown (event not accesible in the regular Button)`
`/// </summary>`

This way, the IDE's intellisense will detect the comment as "something to show" when someone is going to use this class, and it will be more useful.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ExpressionsHUD/ExpressionsHUDView.cs
```diff
@@ -43,7 +43,7 @@ private void Awake()
 
         for (int i = 0; i < hideContentButtons.Length; i++)
         {
-            hideContentButtons[i].onClick.AddListener(HideContent);
+            hideContentButtons[i].onPointerDown += HideContent;
```
sandrade-dcl, line 46: Shouldn't we unsubscribe these subscriptions somewhere along the life-cycle of the view? Maybe in the `OnDestroy` event?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WelcomeHUD/WelcomeHUDView.cs
```diff
@@ -29,10 +29,7 @@ public static WelcomeHUDView CreateView(bool hasWallet)
 
     public void Initialize(UnityAction OnConfirm, UnityAction OnClose)
     {
-        confirmButton.onClick.RemoveAllListeners();
-        confirmButton.onClick.AddListener(OnConfirm);
-
-        closeButton.onClick.RemoveAllListeners();
-        closeButton.onClick.AddListener(OnClose);
+        confirmButton.onPointerDown += OnConfirm;
+        closeButton.onPointerDown += OnClose;
```
sandrade-dcl: The same as before, maybe we should unsubscribe these subscriptions somewhere.

sandrade-dcl (changes requested):
Apart from the suggestions about unsubscriptions (we can discuss if they are necessary or not), I think it is ready to be merged. Also I have been testing the build URL and everythink looks work fine! :)

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/Button_OnPointerDown.cs
pravusjif: done!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ExpressionsHUD/ExpressionsHUDView.cs
pravusjif: Done!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WelcomeHUD/WelcomeHUDView.cs
pravusjif: done!

sandrade-dcl (approved):

PR #972, opened 2020-06-01
fix: user external time API

nchamo:
# What? 
We are using an external API to fetch the current time.

# Why? 
When we used the local time, if users had the wrong time on their computers, they wouldn't be able to login


moliva (approved):
LGTM!

pablitar (approved):

PR #974, opened 2020-06-01
fix: Rejoining rooms when reconnection fails and has to join other realm

pablitar:
# What? 
Making it so the current rooms are stored in the world instance connection instead of the peer, so reconnection can keep those rooms

This fixes a bug that would make a reconnection to other realm not to establish communications correctly until the user moved and the rooms changed.


nchamo (approved):
Looks good :)

moliva (approved):
Great!

PR #979, opened 2020-06-01
fix: unsettle position to avoid concurrent position updates

moliva:
What
------
- Added check to avoid position updates when position is not settled (due to teleport or initial load).
- Changed flow to unsettle position as soon as we recognize the need (i.e. teleport as of today)

Why
----
These checks and flow of events, where enabling concurrent position updates that mangled with the logic of when to load scenes that led to teleports hanging.


pablitar (approved):
Nice!

I think now we have some redundant behaviour because I had to do something similar when changing realms.

Could you check if this can be reused somehow in here https://github.com/decentraland/explorer/blob/master/kernel/packages/shared/comms/index.ts#L206?

Maybe it automatically makes the boolean I added redundant

pbosio (approved):

PR #980, opened 2020-06-01
fix: use configured realm when existing

moliva:
# What? 
Uses configured realm from URL when existing

# Why? 
This parameter was being ignored when a realm was cached, which is not the intended behavior.

This PR updates the flow to first check for a configured realm, when it does exist use it (whether from a cached candidate or forcing a sync init to look for it).
If it is not the case, check for a cached realm or eventually initialize the Catalysts as usual and make the pick.


pbosio (approved):

pablitar (approved):
LGTM!

PR #984, opened 2020-06-02
fix: Friends window is not re-opening after pressing ESC and L key

### LINKED ISSUE #983: Friends window is not re-opening after pressing ESC and L key
## Bug description?
If we have the Friends window open and press the ESC key, we will not be able to open it again using the L key.

## How to reproduce it?
1. Open the Friends window clicking on the taskbar button (or using the L key).
2. Press ESC key in order to close the window.
3. Press L key for open it again.
4. Notice that the Friends window is not opening.

## Expected behaviour
We should be able to re-open the Friends window as many times as we want usgin the L key.

sandrade-dcl:
# What? 
If we have the Friends window open and press the ESC key, we will not be able to open it again using the L key.

**More details:** [https://github.com/decentraland/explorer/issues/983](https://github.com/decentraland/explorer/issues/983)

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/TaskbarHUDController.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/TaskbarHUDController.cs
@@ -86,17 +86,17 @@ public class TaskbarHUDController : IHUD
 
     private void ToggleFriendsTrigger_OnTriggered(DCLAction_Trigger action)
     {
-        OnPressL();
+        OnFriendsToggleInputPress();
     }
 
     private void ToggleWorldChatTrigger_OnTriggered(DCLAction_Trigger action)
     {
-        OnPressReturn();
+        OnWorldChatToggleInputPress();
     }
 
     private void CloseWindowTrigger_OnTriggered(DCLAction_Trigger action)
     {
-        OnPressEsc();
+        OnCloseWindowToggleInputPress();
     }
 
     private void View_OnChatToggleOn()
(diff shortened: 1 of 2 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/TaskbarHUDController.cs
```diff
@@ -341,6 +341,21 @@ public void OnPressEsc()
         worldChatWindowHud.view.ActivatePreview();
     }
 
+    private void OnPressL()
```
pravusjif: I'd change the name of this method to `OnToggleInputPress` or similar, because in the future we may change the key used to toggle this.
Also, I'm not sure why this is here if we have all the input detection logic in InputController, maybe `OnToggleInputPress` is not good enough to depict what the method does, can you think of a better method name?

pravusjif (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/TaskbarHUDController.cs
sandrade-dcl: You are right, I have updated the name of the next methods:

- `OnPressL` to `OnFriendsToggleInputPress`.
- `OnPressReturn` to `OnWorldChatToggleInputPress`.
- `OnPressEsc` to `OnCloseWindowToggleInputPress`.

I think that makes it much clearer.

Note: These methods implements the specific logic (for this controller) that will be executed when each InputAction_Trigger (from InputController) be triggered. That is why they are here.

Thanks!

pravusjif (approved):

PR #985, opened 2020-06-02
feat: Parallel Catalyst & Unity Loading

pablitar:
# What? 
This PR refactors initShared so it can be split in two parts: Essentials, which are quick loading global variables to be used in other initialization processes, and the rest.

Then, the initializer can start loading the unity client only with the essentials, and then await the rest.

This should shave a couple of seconds of the loading process.


pbosio (approved):

moliva (approved):
LGTM.

PR #990, opened 2020-06-03
feat: Add a toggle key for hide/show all the UI

sandrade-dcl:
# What? 
We need a toggle key for hide/show all the UI in the screen. This will allow to make better screenshots or recordings to be shared by the community.


RominaMarchetti:
I've tried to use the other shortcuts while I the UI was invisible. 
Issue  1: after turning it on, when first I clicked M/Tab, instead of seeing the world plus its UI the map appeared. 
Issue 2: clicking enter while UI is off. The chat window opens and then U key doesn't show the UI because it is writing instead XD

Solution: block all key controls until player turns UI on back.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/ShowHideUIByTrigger.cs
```diff
@@ -0,0 +1,50 @@
...
+    private CanvasGroup canvasGroup;
+    private bool isVisible;
+
+    private void Awake()
+    {
+        toggleTrigger = Resources.Load<InputAction_Trigger>(TOGGLE_UI_VISIBILITY_ASSET_NAME);
+        toggleTrigger.OnTriggered += ToggleTrigger_OnTriggered;
+
+        canvasGroup = GetComponent<CanvasGroup>();
+        isVisible = canvasGroup.alpha > 0f;
```
pravusjif: I don't understand how this works for canvases created/configured **after** the triggering has happened.
for example:
1. I use the toggle key and then have no UI
2. I go into a new scene and the scene ui is invisible

I don't see a `ToggleTrigger_OnTriggered()` call inside this `Awake()`, so I don't understand how this works, can you explain that to me? thanks

It is also a bit confusing that the first `isVisible` value is based on the canvasGroup alpha, when that value can be different among canvases (in case any of our canvases already had a canvasGroup for example) and thus they could have a different isVisible initial value when it should be the same for all at all times.

sandrade-dcl: - Related to why is it working for canvases created after the triggering happened: It is because almost all the prefabs involved in this feature have been prefabs that are loaded from the beggining: `Player`, `AvatarHUD`, `ExpressionsHUD`, `Taskbar`, etc.. But you are right, I didn't realize that could there exist other UI game objects created dinamically and then this flow wouldn't works for them. So I'm going to change the flow in order to take on account these cases. Good point, thanks!

- Related to base initially the `isVisible` variable on the alpha value of the CanvasGroup: You're right, I'm going to create a global boolean variable (Sriptable Object) to store the current status of the UI visibility and then my script will use it in order to know how to initialize itself. Also, this SO variable will be necessary for solving the previous point and for blocking all the input keys (the request from Romi). Thanks!

sandrade-dcl: Done!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/ShowHideUIByTrigger.cs
```diff
@@ -1,44 +1,29 @@
...
 /// <summary>
 /// Attaching this component to a canvas, will hide/show it after triggering the ToggleUIVisibility input action.
 /// </summary>
 [RequireComponent(typeof(CanvasGroup))]
 public class ShowHideUIByTrigger : MonoBehaviour
 {
-    private const string TOGGLE_UI_VISIBILITY_ASSET_NAME = "ToggleUIVisibility";
-
-    private InputAction_Trigger toggleTrigger;
+    //private InputAction_Trigger toggleTrigger;
```
pravusjif: Let's remove this commented code

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/ShowHideUIByTrigger.cs
sandrade-dcl: Done!

PR #994, opened 2020-06-03
fix: check tld from cache

moliva:
# What? 
Use cache when TLD when corresponds to current URL.


pravusjif (approved):
thx! now it works as expected, I just tested it

PR #995, opened 2020-06-03
chore: optimize hover feedback

pravusjif:
Improved hover feedback to use 1 canvas and recycle it instead of 1 per pointer event

Profiled at the genesis plaza spawn point and with this optimization:
- In Genesis plaza, Instead of having like 170 canvases for the hover feedback (one for each pointer event component) now we have always just 1 and recycle it (since there will never be more than 1 hover feedback at a time)
- We gain between 0.5 and 1.5 ms of CPU (measured looking at the editor player cpu stats)


sandrade-dcl (approved):
Everything looks fine. The unique thing we will have to be careful is with the `InteractionHoverCanvas` prefab because it has been modified too in my PR about `feat: Add a toggle key for hide/show all the UI`, so we will be pending along the next merges to `master`.

PR #1016, opened 2020-06-08
feat: Show nearby users on the map

### LINKED ISSUE #996: Other users are not shown in the map
After we reimplemented the map, the other users are not shown anymore.

A possible implementation is to create a set or list of users to show in the map with their positions and maybe other data, and then send that list to the map component.
Initially we could populate the list with the users that can be seen in the world (near the current position), but later we can add also all the other users in the real by calling this endpoint: https://peer.decentraland.org/comms/layers/amber/users
Doing it this way, the functionality in the map component should not be changed.

sandrade-dcl:
# What? 
Show nearby users on the map.


unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
```diff
@@ -92,5 +109,27 @@ void SetMinimapRepresentationActive(bool active)
...
+        private void OnEntityTransformChanged(DCLTransform.Model updatedModel)
+        {
+            lastAvatarPosition = updatedModel.position;
+
+            UpdateAvatarIconInMinimap(new MinimapMetadata.MinimapUserInfo
+            {
+                userId = model.id,
+                userName = model.name,
+                worldPosition = updatedModel.position
+            });
```
pbosio: Maybe it's a good idea to have only one instance of `MinimapMetadata.MinimapUserInfo` and modify it's properties before calling `UpdateMinimapUserInformation` to avoid that amount of `new MinimapMetadata.MinimapUserInfo`? It's that possible?

sandrade-dcl: Yes! you're right and that is possible. It's a good idea to have only one instance. I have applied the change. Thanks! :)

pbosio (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
```diff
@@ -83,6 +91,11 @@ public override IEnumerator ApplyChanges(string newJson)
             avatarName.SetName(model.name);
             SetMinimapRepresentationActive(true);
             everythingIsLoaded = true;
+
+            avatarUserInfo.userId = model.id;
+            avatarUserInfo.userName = model.name;
+            avatarUserInfo.worldPosition = lastAvatarPosition != null ? lastAvatarPosition.Value : minimapRepresentation.transform.position;
+            UpdateAvatarIconInMinimap(avatarUserInfo);
```
pravusjif, line 98: I would prefer a minimap-agnostic approach in the AvatarShape side (using AvatarShape events from some Minimap class), but i think that would be more difficult, so It's fine this way in my opinion.

pravusjif (approved):

PR #1018, opened 2020-06-09
fix: profiling panel issues

BrianAmadori:
Testeable here:
[https://explorer.decentraland.zone/branch/fix/profiling-panel-issues/index.html?ENGINE_DEBUG_PANEL](https://explorer.decentraland.zone/branch/fix/profiling-panel-issues/index.html?ENGINE_DEBUG_PANEL)


unity-client/Assets/Scripts/MainScripts/Debugging/Panels/StatsPanelToggler.cs
```diff
@@ -18,28 +17,19 @@ public class StatsPanelToggler : MonoBehaviour
...
             RefreshShortcutText();
         }
 
         void Update()
         {
-            UpdateAllProfilingInput();
             UpdateDeepProfilingInput();
 
-            if (allProfilingEnabled)
+            if (Input.GetKeyUp(KeyCode.P))
```
sandrade-dcl, line 30: Maybe we should manage this input action from the `InputController`, in order to have all the possible inputs centralized.

sandrade-dcl (approved):

pravusjif (approved):
We should hide the panel and have it toggle with a URL param or similar

pbosio:
> We should hide the panel and have it toggle with a URL param or similar
Pravs I think that it is there: `ENGINE_DEBUG_PANEL`

pbosio:
> We should hide the panel and have it toggle with a URL param or similar

Pravs I think that it is already there: `ENGINE_DEBUG_PANEL`

pbosio (approved):

unity-client/Assets/Scripts/MainScripts/Debugging/Panels/StatsPanelToggler.cs
BrianAmadori: Going to skip this for now because I need to merge this now, and this is debugging code, so is not that important to have the key handling through the `InputController`. Maybe we can revisit this improvement later.

PR #1025, opened 2020-06-10
fix: optimize requests to OpenSea

pbosio:
Group together requests into a single request.
Cache responses.


pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea.cs
```diff
@@ -1,33 +1,41 @@
...
+            Request request = requestController.AddRequest(assetContractAddress, tokenId);
+            yield return request.OnResolved(
+                (assetResponse) =>
                 {
-                    AssetResponse response = Utils.FromJsonWithNulls<AssetResponse>(request.downloadHandler.text);
-                    onSuccess?.Invoke(ResponseToNFTInfo(response));
-                }
-                else
+                    NFTInfo nftInfo = ResponseToNFTInfo(assetResponse);
+                    onSuccess?.Invoke(nftInfo);
```
moliva: Why don't you cache the response before resuming invocation flow?

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea_Internal/Request.cs
```diff
@@ -0,0 +1,196 @@
...
+                {
+                    group = requestGroup[i];
+                    break;
+                }
+            }
+            if (group == null)
+            {
+                group = CreateNewGroup();
+            }
+            return group.AddRequest(assetContractAddress, tokenId);
```
moliva, line 35: Could it happen that a single assetContractAddress, tokenId pair is added more than once to the same request? Should we avoid this situation?

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea_Internal/Request.cs
```diff
@@ -0,0 +1,196 @@
...
+            yield return new WaitForSeconds(WAIT_TIME);
+            isOpen = false;
+            this.onGroupClosed?.Invoke(this);
+            this.onGroupClosed = null;
+
+            string url = API_URL_ASSETS;
+
+            for (int i = 0; i < requests.Count; i++)
+            {
+                url += requests[i].ToString();
```
moliva: There is a maximum number of characters for a single URL, we might need to check this to limit the number of assets added in each group to avoid surpassing it.

moliva (commented):
Adding a couple of questions.

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea_Internal/Request.cs
pbosio: > Could it happen that a single assetContractAddress, tokenId pair is added more than once to the same request? Should we avoid this situation?

It shouldn't happen, but It could happen.
In case it happen it won't break anything anyway :P

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea_Internal/Request.cs
pbosio: > There is a maximum number of characters for a single URL, we might need to check this to limit the number of assets added in each group to avoid surpassing it.

Thanks, I will look into that

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea.cs
pbosio: > Why don't you cache the response before resuming invocation flow?

You are right, I will change it to cache the response first

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea_Internal/Request.cs
```diff
@@ -0,0 +1,196 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Networking;
+
+#pragma warning disable CS0162
+
+namespace DCL.Helpers.NFT.Markets.OpenSea_Internal
+{
+    internal class RequestController
```
BrianAmadori: Maybe we can rename this to `OpenSeaRequestController`? As i feel the name seems too generic for this class purpose and could cause some confusions. Same with the rest of the class names. Yes, they are enclosed by a custom namespace, but namespace scope is too lenient because of `using`.

BrianAmadori (approved):

moliva (approved):

PR #1028, opened 2020-06-11
feat: World Chat Entry's contextual menu

### LINKED ISSUE #1021: World Chat Entry's contextual menu
UX description below the image

- Right click in world chat messages opens a contextual menu showing player's name plus the following options: Passport / Block / Report.
- The panel will always at the right top corner of the scroll view.
-  The menu will close if:
1. Choose one of the panel options
2. The Player presses Esc key once
3. The Player clicks any part of the world chat widget

-Disclaimer- 
The chat widget and the contextual menu, in case it is opened, will close if the player clicks the world.

sandrade-dcl:
# What? 

- Right click in world chat messages opens a contextual menu showing player's name plus the following options: Passport / Block / Report.

- The panel will always open at the top of the clicked message

- The menu will close if:

1. Choose one of the panel options
2. The Player presses Esc key once
3. The Player clicks any part of the world chat widget

-Disclaimer-
The chat widget and the contextual menu, in case it is opened, will close if the player clicks the world.

### DIFF
--- a/kernel/packages/shared/chat/sagas.ts
+++ b/kernel/packages/shared/chat/sagas.ts
@@ -179,7 +179,7 @@ function* handleSendMessage(action: SendMessage) {
       messageType: ChatMessageType.PUBLIC,
       messageId: uuid(),
       timestamp: Date.now(),
-      sender: currentUser.userId || currentUser.profile.name || 'unknown',
+      sender: currentUser.userId || currentUser.profile.userId || 'unknown',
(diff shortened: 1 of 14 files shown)


pravusjif (approved):
My only concern is that we should probably use the new `UserContextMenu` class for the context menu in the Friend entries.

moliva (approved):
LGTM.

PR #1029, opened 2020-06-11
feat: init pm post load

moliva:
# What? 
Make private messaging subsystem initialize after the initial world load.

# Why? 
It speeds up initial loading.


BrianAmadori (approved):
Works as a charm

pablitar (approved):

moliva:
Cherrypicking commits into base, to avoid other commits from master.

PR #1033, opened 2020-06-12
fix: wait for renderer enable before nft requests

pbosio:
Don't try to load nft images when renderer is deactivated.

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_NFT.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_NFT.cs
@@ -10,6 +10,8 @@ namespace DCL.Components
         public bool withCollisions;
         public Color backgroundColor;
 
+        string assetUrl;
+
         public override void Unload()
         {
             loaderController.OnLoadingAssetSuccess -= CallOnComponentUpdated;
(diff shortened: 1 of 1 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_NFT.cs
```diff
@@ -37,10 +40,19 @@ public override void Load(string src, System.Action<LoadWrapper> OnSuccess, Syst
 
             loaderController.OnLoadingAssetSuccess += CallOnComponentUpdated;
 
-            loaderController.OnLoadingAssetSuccess += () => OnSuccess(this);
-            loaderController.OnLoadingAssetFail += () => OnFail(this);
+            assetUrl = src;
 
-            loaderController.LoadAsset(src, true);
+            if (CommonScriptableObjects.rendererState)
```
BrianAmadori: This wouldn't be like this?
```suggestion
            if (CommonScriptableObjects.rendererState.Get())
```

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_NFT.cs
pbosio: `BaseVariable<T>`  has an implicit cast to `T` , boolean in this case, but I'll change it anyway to avoid any confusion

BrianAmadori (approved):

pravusjif (approved):

PR #1035, opened 2020-06-12
fix: Avoid showing duplicate avatars when there are multiple peer messages with the same identity

pablitar:
# What? 
This PR adds code to ensure that there is only one peer tracking info for each identity, and should ensure that there are no more avatar duplicates seen by the players


moliva (approved):
🙌

pbosio (approved):

PR #1037, opened 2020-06-13
fix: update amd loader for better debugging and understandability

menduz:
# What?

Proper names and error catching in amd-loader

# Why?

There was several scenes halting not knowing why. This PR fixes some bugs I fixed a while ago in [ulla](https://github.com/arduz-io/ulla) and adds a lot of readability to the code.

+ Also eliminates obscure names and make things easier to read.
+ Also compiles with strict mode

### DIFF
--- a/kernel/packages/decentraland-amd/package.json
+++ b/kernel/packages/decentraland-amd/package.json
@@ -12,11 +12,17 @@
   "files": [
     "dist"
   ],
+  "prettier": {
+    "semi": true,
+    "singleQuote": false,
+    "printWidth": 120
+  },
   "license": "MIT",
   "devDependencies": {
     "expect": "^23.6.0",
     "mocha": "^5.2.0",
-    "typescript": "^3.1.2",
-    "uglify-js": "^3.4.9"
-  }
+    "terser": "^4.7.0",
+    "typescript": "^3.9.5"
+  },
+  "dependencies": {}
 }
--- a/kernel/packages/decentraland-amd/package.json
+++ b/kernel/packages/decentraland-amd/package.json
@@ -14,7 +14,7 @@
   ],
   "prettier": {
     "semi": false,
-    "singleQuote": false,
+    "singleQuote": true,
     "printWidth": 120
   },
   "license": "MIT",
--- a/kernel/packages/decentraland-amd/package.json
+++ b/kernel/packages/decentraland-amd/package.json
@@ -13,7 +13,7 @@
     "dist"
   ],
   "prettier": {
-    "semi": true,
+    "semi": false,
     "singleQuote": false,
     "printWidth": 120
   },
--- a/kernel/packages/decentraland-amd/tsconfig.json
+++ b/kernel/packages/decentraland-amd/tsconfig.json
@@ -1,10 +1,10 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es2017",
     "diagnostics": false,
-    "inlineSourceMap": false,
-    "inlineSources": false,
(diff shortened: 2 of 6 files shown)


kernel/packages/decentraland-amd/tsconfig.json
```diff
@@ -1,10 +1,10 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es2017",
```
eordano, line 3:
```suggestion
    "target": "es5",
```

kernel/packages/decentraland-amd/tsconfig.json
```diff
@@ -13,9 +13,18 @@
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true,
     "noUnusedLocals": true,
-    "lib": ["es5"],
+    "strict": true,
+    "strictNullChecks": true,
+    "lib": [
+      "es2017"
```
eordano, line 19:
```suggestion
      "es5"
```

kernel/packages/decentraland-amd/package.json
```diff
@@ -12,11 +12,17 @@
   "files": [
     "dist"
   ],
+  "prettier": {
+    "semi": true,
```
eordano:
```suggestion
    "semi": false,
```

kernel/packages/decentraland-amd/package.json
```diff
@@ -12,11 +12,17 @@
...
+    "singleQuote": false,
```
eordano:
```suggestion
    "singleQuote": true,
```

kernel/packages/decentraland-amd/tsconfig.json
menduz: It uses sets and array.from, there is no real reason to stick to ES5 anymore

kernel/packages/decentraland-amd/package.json
menduz: I did not want to change the styles that were already in place for the file. may that be another pr?

kernel/packages/decentraland-amd/package.json
menduz: same

kernel/packages/decentraland-amd/package.json
eordano: then let's rather drop this section

menduz:
Styles fixed. What do we resolve with the ES6 features?

@nearnshaw had his scene fixed but raised a concern about other scene getting broken because of this. Could you Nico post more details here?

nearnshaw:
I tried this on a few existing scenes, and it solved problems on some but raised new ones on others.

Specifically, I tried this out on the Genesis Plaza and Soho Plaza scenes, that both have a lot of different files importing from each other, but that currently work well.
In both cases the scene stoped loading using this. The console claimed that it couldn't load any of the modules, saying that perhaps there were circular dependencies. 
In the code there are dependencies between different .ts files, and there were also libraries that were referenced  several times from multiple files, but no circular dependencies between objects, systems, etc.

https://github.com/decentraland-scenes/Genesis-Plaza
https://github.com/decentraland-scenes/Soho-Plaza

image


We can't merge this if it breaks existing scenes like that, there may be many affected scenes that don't get fixed.

kernel/packages/decentraland-amd/tsconfig.json
eordano: Thanks! That's related to the `lib`, right? Let's leave the `target` as es5

PR #1042, opened 2020-06-17
fix: scenes ui stretching

pbosio:
Scenes UI were being stretch cause `ConstrainedPanel` was scaled in Y axis


pravusjif (approved):
I tested a couple of scenes with UI and didn't see any issue.
The only thing that concerns me a bit is the pivot setting **after** setting the size, i would think the pivot setting should be before, but somehow this is working fine, maybe the same pivot setting has already been confifured on the canvas in unity editor.

pbosio:
> I tested a couple of scenes with UI and didn't see any issue.
> The only thing that concerns me a bit is the pivot setting **after** setting the size, i would think the pivot setting should be before, but somehow this is working fine, maybe the same pivot setting has already been confifured on the canvas in unity editor.

I did that change just for your peace of mind :P 
according to unity docs `RectTransform.sizeDelta: The size of this RectTransform relative to the distances between the anchors` so it seems it doesn't really matter the pivot... but just in case... :)

sandrade-dcl (approved):

PR #1044, opened 2020-06-18
fix: Don't send position if it hasn't change since last time

pablitar:
# What? 
To reduce bandwidth usage, we send positions through comms only if they have changed or if 5 seconds have passed since last update (to ensure eventual consistency if the last position didn't reach the other peers)


moliva (approved):
LGTM.

pbosio (approved):

PR #1049, opened 2020-06-22
chore: Refactorize UserContextMenu to be used by friend entries too

sandrade-dcl:
# What? 
Refactorize the UserContextMenu class to be used by friend entries too and avoid to repeat 2 very similar implementations.


BrianAmadori (approved):
Good job!. Tested in build and seems to work ok.

pravusjif (approved):

PR #1056, opened 2020-06-24
feat: Don't connect to peers that are too far away

pablitar:
# What? 
This PR updates the Peer library to use the new optimizations that don't connect to far away peers.

There are three mechanisms in place:
1. If a peer shares at least a room with another, then they can connect (this is for allowing communications in large scenes)
2. If a peer is less than 8 parcels away of another, then they can connect
3. If a peer is more than 10 parcels away of another and don't share a room, then they'll drop connection

Since the peer use local structures to track other peers data, and those structures may be out of date, we are trying to be less restrictive than we could be. Still, this should be a lot more accurate than what we have today.


kernel/packages/shared/comms/index.ts
```diff
@@ -651,7 +651,10 @@ export async function connect(userId: string) {
               if (context && context.currentPosition) {
                 return context.currentPosition.slice(0, 3)
               }
-            }
+            },
+            maxConnectionDistance: 4,
+            nearbyPeersDistance: 5,
+            disconnectDistance: 5
```
eordano, line 657: should this be `10`?

pablitar: This is somewhat confusing, but 5 is ok.

Since we are using the "rings" mechanisms for distance, this actually means that the disconnect distance is for any peer being in the ring 5, it being anything with distance greater than 160 (10 parcels).

Since we are using the same distance function in lighthouse, this is actually defined here: https://github.com/decentraland/catalyst/blob/c84d4ed0893ebb7caa462f40179f727e1c5a9311/commons/utils/Positions.ts#L1

And yes, this means that eventually we will need to update the catalysts too. But fortunately it is not critical, since the effect we will have until we update the catalysts is that when the peer queries the lighthouse for nearby peers, it will respond with every peer, since the lighthouse will be using the old function which only had up to ring 3. But the peer will then sort the information locally. This would only be problematic if we had thousands of peers on each layer.

If I'm being honest, I don't like how confusing this is, and I'm not sure it is worth the confusion haha. Maybe we can change the distance function to be a simple parcel distance...

eordano: Thanks for clarifying!

eordano (approved):
I love this change, thanks.

moliva (approved):

pravusjif:
Tested the build with no problems, went to the art event that has many people right now

PR #1063, opened 2020-06-25
fix: check resize URLs in fallback service

### LINKED ISSUE #1061: Avatar snapshot not loading for some users
After the performance improvements in #1027, some users reported that their avatar snapshot never appeared.

image
image

This problem was reported by users in different realms, with and without the lambdas improvements.

Apparently, it can be fixed by manually saving a new version of the profile.

moliva:
# What? 
Fixes #1061 by adding a check to fallback resize service URLs and patches affected profiles on login.


pbosio (approved):

pablitar (approved):
LGTM!

PR #1065, opened 2020-06-25
fix: pointer lock

pbosio:


pablitar (approved):
Tested it a bit. Seems to work great! Couldn't reproduce the problem trying for some time.

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/MouseCatcher/MouseCatcher.cs
```diff
@@ -40,20 +44,41 @@ public void LockCursor()
...
             OnMouseLock?.Invoke();
+#endif
         }
 
-        //Externally called by the browser
         public void UnlockCursor()
         {
             Utils.UnlockCursor();
+#if !WEB_PLATFORM            
             OnMouseUnlock?.Invoke();
```
BrianAmadori, line 56: This means the events will not fire after a lock change in build? This isn't affecting any of our systems?

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/MouseCatcher/MouseCatcher.cs
pbosio: > This means the events will not fire after a lock change in build? This isn't affecting any of our systems?

For web build the event was moved to the function called by kernel

        //Externally -ONLY- called by the browser
        public void UnlockCursorBrowser(int val)
        {
            bool lockPointer = val != 0;
            Utils.BrowserSetCursorState(lockPointer);
            if (lockPointer)
            {
                OnMouseLock?.Invoke();
            }
            else
            {
                OnMouseUnlock?.Invoke();
            }
        }

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/MouseCatcher/MouseCatcher.cs
BrianAmadori: We are triggering the mouse lock internally too, and `TaskbarHudController` depends on the event to trigger correctly on all cases. This may introduce bugs. (going to test it now to make sure)

BrianAmadori: Ok nvm, I suppose that when we trigger the mouse lock internally, the browser event triggers and the roundtrip is completed

BrianAmadori (approved):

PR #1069, opened 2020-06-29
test:fractions

EibrielInv:
# What? 
A test about how well the Explorer handles numbers that can not be perfectly represented in Binary by the SDK

# Why? 
Setting a fractional number in the Transform can lead to unexpected results
imagen


nchamo:
Hey @EibrielInv !

I'm closing this, since we did the fix in https://github.com/decentraland/explorer/pull/1199

Thank you for reporting this, and taking the time to build a scene to reproduce it!

PR #1071, opened 2020-06-30
feat: loading screen improvements (first iteration)

moliva:


pbosio:
Can we do something about the TOS checkbox and unity loading on the background? It's impossible to check that checkbox and there is no indication on what's going on :)

moliva:
We put a spinner, for the time being, until the Unity module is initialized to avoid the TOS check box issue.

RominaMarchetti:
@moliva is it possible to change the spiner? It looks too **bold** :P 

/See asset attached/
Spinner

I want it to look like [this](https://www.figma.com/file/akmeRX6da0lnNuJuN50eMC/LoadingScreen?node-id=167%3A2)

BrianAmadori (commented):
The tooltip images are being requested continuously while the client loads. Eye (?

moliva:
The tooltip images are requested by the browser and get a 304 status, but still the requests are taking ~300ms. I pushed a fix to this caching images and making the browser load them from its own memory, making local requests of ~3ms instead.

RominaMarchetti (approved):

PR #1074, opened 2020-07-01
feat: Updated peer library to use relay suspension protocol

pablitar:
# What? 
Updated peer library to be able to use relay suspension protocol & linked configuration.json to enabling and configuring the protocol.

See: https://github.com/decentraland/catalyst/pull/155


kernel/packages/shared/comms/index.ts
```diff
@@ -625,12 +626,13 @@ export async function connect(userId: string) {
         break
       }
       case 'v2': {
+        await ensureMetaConfigurationInitialized()
```
pablitar, line 629: @moliva Any objection to adding this method? Is there another way to await for meta config?

moliva: I think it is ok. Metaconfiguration should be one of the first checkpoints setting the client up and should be up by this point. Being that we are not currently in a saga we have to use this kind of structures, but could later be changed to `waitForMetaConfigurationInitialization` (in `shared/meta/sagas` module) to avoid repetition.

moliva (approved):
LGTM!

pbosio (approved):

PR #1075, opened 2020-07-01
feat: Improve Loading Screen Feedback

sandrade-dcl:
# What? 
Give more details in the scenes loading process, showing the number of assets currently downloading and the number of components currently loading.


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -90,8 +92,9 @@ private void Update()
 
         protected virtual string prettyName => sceneData.basePosition.ToString();
 
-        protected void RefreshName()
+        protected void RefreshSceneState()
```
BrianAmadori: The `RefreshName()` method was intended for debugging purposes, here we are adding another responsibility to it. Maybe its a bit more maintainable to just make `state` a setter and call the event from there?.

```
public int stateValue;
public int state 
{
    get { return stateValue; }
    set 
    {
        if ( stateValue != value )
            OnStateRefreshed?.Invoke(this);  

        stateValue = value;
    } 
}
```

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
sandrade-dcl: You're right, it is better to separate resposibilities. Done! Thanks :)

pravusjif (approved):

BrianAmadori:
Merging this on monday after the weekend to avoid fires.

PR #1079, opened 2020-07-03
feat: adding pin catalyst param

moliva:
# What? 
Pins a selected Catalyst server by passing the `PIN_CATALYST=https://catalyst.url` to Explorer URL, using all services from that catalyst, avoiding pinging other servers and disabling asset bundles centralized service.

E.g.:`https://explorer.decentraland.zone/branch/feat/pin-catalyst/index.html?PIN_CATALYST=https://peer.decentraland.zone`


kernel/packages/decentraland-loader/lifecycle/manager.ts
```diff
@@ -101,7 +101,7 @@ export async function initParcelSceneWorker() {
     metaContentService: DEBUG
       ? resolveUrl(document.location.origin, '/local-ipfs')
       : getFetchMetaContentService(window.globalStore.getState()),
-    contentServerBundles: DEBUG ? '' : getServerConfigurations().contentAsBundle + '/',
+    contentServerBundles: DEBUG || PIN_CATALYST ? '' : getServerConfigurations().contentAsBundle + '/',
```
eordano, line 105: Not sure about precedence here, is this the same as 
```suggestion
    contentServerBundles: (DEBUG || PIN_CATALYST) ? '' : getServerConfigurations().contentAsBundle + '/',
```
?

moliva: Indeed, it's the same result.

PR #1081, opened 2020-07-06
 fix: performance drop on expression animation call

pravusjif:
**WHY?**
Right now, every time we have an `OnUpdate` call on the profile, the `AvatarHUDController` processes the whole profile catalogue, and this causes a big performance drop (see image attached). 

Screen Shot 2020-07-06 at 11 33 55

In some cases, like every time we call an avatar expression animation, the AvatarHUD is hidden and not being used, and we could avoid having this heavy performance hit.

**WHAT?**
Added escape check in `AvatarHUDController.LoadUserProfile()` to avoid the whole processing if the rendering is enabled and the AvatarHUD hidden.

Test instructions:
- The performance impact fix can only be seen when profiling in unity editor
- Test that when opening the avatar hud, we still see the avatar rendered in there


BrianAmadori (approved):
Tested and it seems to work correctly. The tests should be fixed tho.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/AvatarEditorHUDController.cs
```diff
@@ -61,7 +62,12 @@ public void SetCatalog(WearableDictionary catalog)
 
     public void LoadUserProfile(UserProfile userProfile)
     {
-        if (userProfile?.avatar == null || string.IsNullOrEmpty(userProfile.avatar.bodyShape)) return;
+        LoadUserProfile(userProfile, false);
+    }
+
+    public void LoadUserProfile(UserProfile userProfile, bool forceLoading)
+    {
+        if ((!forceLoading && renderingEnabled && !view.isOpen) || userProfile?.avatar == null || string.IsNullOrEmpty(userProfile.avatar.bodyShape)) return;
```
BrianAmadori: Maybe we can use `Application.isBatchMode` here to skip that conditional, so the tests don't break?

sandrade-dcl (approved):

PR #1088, opened 2020-07-07
feat: avatars pooling

pravusjif:
What this PR brings to the table:

- [x] Potential pooling support for all components
- [x] Pooling implemented for `AvatarShape`
- [x] Wearables are re-used (instead of being always instantiated). The material colors are applied after recycling and reverted on release.

Minor improvements:
- Mark green blocker pools as persistent, so the pool is never cleared up.
- Improvements on the `ParcelScenesCleaner` algorithm. Found a bug that made entities not being cleared up properly after teleports and could potentially destroy a scene out of nowhere.

More last minute fixes:
- fix: all avatar loading logic was being executed when updating any avatar with same wearables.
- fix: all avatar loading logic was being executed when performing emotes.
- fix: existing wearables rootBone was not being updated when changing bodyShape, leading to missing or floating wearables in random cases.
- fix: rootBone was being assigned needlessly on unequipped wearables when changing wearables.
- fix: now avatars can be loaded even when gameObjects are inactive--the loading will resume when the object is re-activated


sandrade-dcl (commented):
I'm trying to generate bots with `bots.create()` but I can't because it is throwing a null reference exception in the `UpdateFromJSON()` (BaseComponent, line 63). It seems the `updateHandler` variable comes null.

Edit: Ok, it was a problem in my local project. I regenerated the Library folder and it works.

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/AvatarEditorHUDView.cs
```diff
@@ -300,6 +303,8 @@ public void SetVisibility(bool visible)
         avatarEditorCanvas.enabled = visible;
         avatarEditorCanvasGroup.blocksRaycasts = visible;
         isOpen = visible;
+
+        // characterPreviewController.gameObject.SetActive(visible);
```
BrianAmadori: Remove commented line

BrianAmadori (approved):

sandrade-dcl (approved):
**Only as annotation for the future improvements:** I have noticed the framerate drop that we have while we are with a large number of avatars in the scene (from 60fps to 35-40 aprox.), is only happening when the camera is rendering the avatars. In fact if we look at other place where there are not avatars (even with all these avatars near of us), the framerate go up to 60fps.

PR #1089, opened 2020-07-08
feat: chat message when friend goes online

pbosio:
Screen Shot 2020-07-06 at 16 33 40


pravusjif (approved):

RominaMarchetti (approved):

kernel/packages/shared/comms/chat.ts
```diff
@@ -25,3 +27,41 @@ export function notifyStatusThroughChat(status: string) {
...
+    }
+
+    globalThis.globalStore.dispatch(
+      messageReceived({
+        messageId: uuid(),
+        messageType: ChatMessageType.SYSTEM,
+        timestamp: Date.now(),
+        body: message
+      })
+    )
```
moliva: I'd rather use the `shared/comms/chats.ts#notifyStatusThroughChat` function instead to avoid repetition.

kernel/packages/shared/comms/chat.ts
```diff
@@ -1,9 +1,11 @@
-import { MessageEntry, ChatMessageType } from 'shared/types'
+import { MessageEntry, ChatMessageType, PresenceStatus, UpdateUserStatusMessage } from 'shared/types'
 import { uuid } from 'atomicHelpers/math'
 import { StoreContainer } from '../store/rootTypes'
 import { messageReceived } from '../chat/actions'
+import { getProfile } from 'shared/profiles/selectors'
 
 declare const globalThis: StoreContainer
+let friendStatus: any = {}
```
moliva: Let's improve type check here.

moliva (changes requested):

kernel/packages/shared/comms/chat.ts
pablitar: Agree. This is basically a `Record<string, PresenceStatus>` right?

kernel/packages/shared/comms/chat.ts
pablitar: Agree. This could be replaced with: `notifyStatusThroughChat(message)`. Maybe it could also be renamed to `sendSystemMessageThroughChat`, but I'm not convinced that would be better.

pablitar (approved):
LGTM!

moliva (approved):
Nice! 🙌

PR #1093, opened 2020-07-08
fix: add hiccup metrics

BrianAmadori:
This PR adds hiccup and percentile metrics to the current FPS meter. Remember FPS meter is activated with the `/showfps` chat command.

- Avg hiccup ms: How high are hiccups in average? Lower is better.
- Max hiccup count per minute: In the worst case, how many hiccups occur per minute. Lower is better. 
- 30 fps percentile %:  How much of 30fps or faster frames exists relative to slower frames. Near to 100% is better.


sandrade-dcl (approved):
Fantastic tool to measure the hiccups during our performance-related tests, thanks! :)

pravusjif (approved):

PR #1097, opened 2020-07-09
feat: custom startcoroutine

BrianAmadori:
Wrote this PR earlier this year in order to have a more centralized approach of all our subsystems that used throttling.

## Explanation
Right now, the following systems use coroutines with independent throttling:

- `AssetPromiseKeeper` to resolve many promises without hiccups.
- `GLTFSceneImporter` to throttle the parsing/loading of models
- `AssetPromise_AB` to throttle the loading of individual AssetBundles assets.
- `MessagingControllerManager` to throttle messages
- `SceneController` to throttle decoding
- `MemoryManager` and `ParcelScenesCleaner` to throttle memory unloading
- `MaterialCachingHelper` to put materials inside our persistent cache and optimize rendering

On top of that we are using coroutines for other subsystems that have no throttling, like:

- `SceneBoundariesChecker` that checks periodically objects are inside the scene bounds
- `AvatarShape` loading
- Thumbnail loading
- `ComponentUpdateHandler` that manages an internal queue of `ApplyChanges` calls (so they don't interrupt each other if a message is processed before the previous `ApplyChanges` finished).

And so on.

The issue of having so many subsystems running on tandem, is that regardless of what budget are we setting, there is no encompassing system that manages the CPU load between them. There's very few code in place that prevents that all of them use all their throttling capacity, and when that happens we have hiccups. 

The idea of this PR is to gain full control of what is going to be processed each frame, this way it'd be way easier to achieve a stable framerate.

In this implementation, we aim to centralize all our sub-systems execution in a single custom coroutine runner that has a single specified budget. This way we can ensure this budget each frame. To have a proper control of each frame, we have to prioritize those systems, so they execute i.e. in the following cascading order:

1. Decoding messages
2. Enqueue and process messages
3. Resolve promises
4. Load asset bundles and GLTFs
5. Memory management, parcels cleaner, etc.

The custom coroutine runner executes all of those in the determined order, and checks in every `yield return` if the CPU budget is filled out. This also increases the granularity of the process spreading, improving the throttling accuracy.


BrianAmadori:
This is nowhere near being merged. Will close it for now.

PR #1099, opened 2020-07-10
fix: stuck local build issue

BrianAmadori:
Local build was failing due to `CachedXMLHTTPRequest` logic issues.

These logic issues involved the fact that the private functions of our plugin were overrided as intended (i.e. `send()`). But the defined properties (defined with `Object.defineProperty`) were still using the old plugin!.

Changed our plugin name and disabled the built-in plugin using the `Module` flag to prevent this weird behaviour.

Also, this means that `hotpatch.sh` is not needed anymore, so removed it from the build system.


pravusjif (approved):
it worked fine for me

eordano (approved):

eordano:
Confirmed patch not necessary anymore

PR #1101, opened 2020-07-10
fix: asset bundle load hiccups

BrianAmadori:


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
```diff
@@ -24,35 +22,17 @@ public class AssetPromise_AB : AssetPromise_WithUrl<Asset_AB>
...
-            {"ltf", 8},
-            {"glb", 9}
-        };
-
-        private IOrderedEnumerable<string> assetsToLoad;
+        public static AssetBundlesLoader assetBundlesLoader = new AssetBundlesLoader();
 
         public AssetPromise_AB(string contentUrl, string hash) : base(contentUrl, hash)
         {
+            assetBundlesLoader.Start();
```
BrianAmadori, line 37: Beware! here you are re-starting the loader for every promise that gets created! This needs a null check.

sandrade-dcl: I know, that's why I added that null check on the loader `Start()` function:

```
public void Start()
{
     if (assetBundlesLoadingCoroutine != null)
          return;

     assetBundlesLoadingCoroutine = CoroutineStarter.Start(LoadAssetBundlesCoroutine());
}
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetBundlesLoader.cs
```diff
@@ -8,9 +8,10 @@ namespace DCL
 {
     public class AssetBundlesLoader
     {
-        private const int SKIPPED_FRAMES_AFTER_BUDGET_TIME_FOR_NEARBY_ASSETS = 1;
-        private const int SKIPPED_FRAMES_AFTER_BUDGET_TIME_FOR_DISTANT_ASSETS = 10;
-        private const float MAX_SQR_DISTANCE_FOR_QUICK_LOADING = 5000f;
+        private const float MAX_LOAD_BUDGET_TIME = 0.004f;
+        private const int SKIPPED_FRAMES_AFTER_BUDGET_TIME_IS_REACHED_FOR_NEARBY_ASSETS = 1;
+        private const int SKIPPED_FRAMES_AFTER_BUDGET_TIME_IS_REACHED_FOR_DISTANT_ASSETS = 5;
+        private const float MAX_SQR_DISTANCE_FOR_QUICK_LOADING = 8000f;
```
BrianAmadori: This distance is inside our loading radius, right? How are we relating it to that radius?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetBundlesLoader.cs
```diff
@@ -87,13 +87,13 @@ private IEnumerator LoadAssetBundlesCoroutine()
...
                     assetBundleInfoToLoad = assetBundlesReadyToBeLoaded.Dequeue();
-                    yield return LoadAssetsInOrder(assetBundleInfoToLoad, SKIPPED_FRAMES_AFTER_BUDGET_TIME_FOR_NEARBY_ASSETS);
+                    yield return LoadAssetsInOrder(assetBundleInfoToLoad, SKIPPED_FRAMES_AFTER_BUDGET_TIME_IS_REACHED_FOR_NEARBY_ASSETS);
                 }
 
                 while (assetBundlesWaitingForLoad.Count > 0 && assetBundlesReadyToBeLoaded.Count == 0)
                 {
                     assetBundleInfoToLoad = assetBundlesWaitingForLoad.Dequeue();
-                    yield return LoadAssetsInOrder(assetBundleInfoToLoad, SKIPPED_FRAMES_AFTER_BUDGET_TIME_FOR_DISTANT_ASSETS);
+                    yield return LoadAssetsInOrder(assetBundleInfoToLoad, SKIPPED_FRAMES_AFTER_BUDGET_TIME_IS_REACHED_FOR_DISTANT_ASSETS);
```
BrianAmadori, line 107: What happens if any asset inside this low priority queue is loading slowly and we get near to it? It shouldn't be promoted to the high priority one?

BrianAmadori:
I think we should re-open this PR so the commit count gets reasonable. Having 61 commits for this changeset will get the git log dirty.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetBundlesLoader.cs
sandrade-dcl: For now `MAX_SQR_DISTANCE_FOR_QUICK_LOADING` is only an approximate value based on a distance measured by me on the world. If we have the loading radius accessible, I think it would be a good idea to base the `MAX_SQR_DISTANCE_FOR_QUICK_LOADING` value on it (for example: using a percentage of the radius value), what do you think?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetBundlesLoader.cs
sandrade-dcl: Yeah, I already had that case in mind and I was planning, as a second iteration, to implement some kind of system able to prioritize the loading of some bundles from the "low priority" queue. But I still don't have a final solution to apply.

BrianAmadori: Are you sure we can push this to a second iteration? If we have ABs with a big asset list, this can downgrade the experience too much. 

i.e.:
* User starts exploring the world
* AB start downloading from far away, slowly
* User gets near the scene before the ABs finish loading
* User has to wait too much time for them to finish because they were enqueued in the low priority list

sandrade-dcl: With "second iteration" I wanted to say "I'll do it during this implementation, after these first tests", sorry for the confussion :D

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetBundlesLoader.cs
BrianAmadori: Yeah, the ideal is to adapt to the loading radius, but for now we can simply tune this value knowing on advance the radius in meters.

I think right now the loading radius is of 4 parcels (the unloading radius being 8, to prevent having loading/unloading overheads overlap).

So, each parcel is 16 meters. 16 * 4 =  64 meters. 64 squared = 4096. So the slow loading queue shouldn't never trigger with 8000?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetBundlesLoader.cs
sandrade-dcl: I have added a system for re-prioritize the "low priority" queue. The approach is: each time a new AB is marked for load, before add it to any queue, we review the "low priority" queue and we go taking ABs while their distances to the player are less than the configured limit (`MAX_SQR_DISTANCE_FOR_QUICK_LOADING`). All these taken AB's are added in the "hig priority" queue.

In order to avoid making too many calls to this "reprioritization" function, I have added a `TIME_BETWEEN_REPRIORITIZATIONS` (currently set as `1 sec`) that will indicates how many time should pass between each re-prioritization.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetBundlesLoader.cs
sandrade-dcl: I have set the `MAX_SQR_DISTANCE_FOR_QUICK_LOADING` as `6000f` and I have added a temporal UI tool (it will be deleted after) in order to view how both queues are being filling in the build runtime.

PR #1111, opened 2020-07-15
fix: work around for multiple welcome chat messages

moliva:
# What? 
Makes sure that welcome message in chat is only shown at the start of the client.


BrianAmadori (approved):
Tested and help appeared only once!.

sandrade-dcl (approved):

PR #1112, opened 2020-07-15
chore: tweak hiccup metric

BrianAmadori:
Use frames as margin instead seconds would allow for a cleaner metric. Using minutes would mean that we are always sampling hiccups around a variable number of frames.


sandrade-dcl (approved):
Much clearer the new "Hiccup Rate" metric, thanks!

pravusjif (approved):

PR #1113, opened 2020-07-16
fix: Make the Loading Bar realistic

sandrade-dcl:
# What? 
Make the Loading Bar realistic.


RominaMarchetti:
@sandrade-dcl  It looks weird for me just seen the first badge of messages without the bar (related with comms and authentication), and suddenly it appears almost complete. Can we show it since the very beginning?

pravusjif (approved):

BrianAmadori (approved):

BrianAmadori:
The comms/communication step may get a bit long, and when that happens the bar get stuck for a while. Maybe we can start increasing the bar a bit until we get to the unity step? This is only a minor detail, looking great otherwise.

PR #1119, opened 2020-07-19
chore: tweak hiccup metric (part 2)

eordano:
This builds on the previous hiccup metric improvements, changing:

* Use the latest info, so we don't carry error data from too many frames ago
* Add Number of seconds spent in hiccups in the last 1000 frames
* Add Amount of frames in the last 1000 that are hiccups

Also, adding tests and splitting out FPSDisplay so it isn't in the big MainScripts assembly.

image


eordano:
I think this gives a more accurate view of when the experience is running badly:

image

I'd say we should target less than 1% of "bad frames" (maximum 10 hiccups every 1000 frames, maximum 1 second loss in hiccups every 100 seconds). I'm fairly certain that hiccup frames are strictly related to loading.

sandrade-dcl (approved):

BrianAmadori (approved):
I have a few mini-changes in mind to improve this code but for now it looks as good enough!

PR #1124, opened 2020-07-20
fix: avatar editor thumbnail issues

BrianAmadori:
This PR contains the following:

- `ThumbnailManager` refactor, removed all redundant logic already covered by our beloved `AssetPromiseKeeper_Texture`.
- fix: `AssetPromise_Texture` was not counting references correctly, leading to thumbnails dissapearing when they shouldn't.
- add code to make sure we don't reload the same texture due to refCount going to zero inter-frame.


sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Texture/Asset_Texture.cs
```diff
@@ -19,13 +22,17 @@ public void ConfigureTexture(TextureWrapMode textureWrapMode, FilterMode texture
 
         public void CopyTextureFrom(Texture2D sourceTexture)
         {
+            if (this.texture == null)
```
pravusjif: Is this check correct or should it be `texture != null` ?

pravusjif (approved):

PR #1130, opened 2020-07-22
refactor: avoid modifying DOM when values are equal

moliva:
# What? 
Only modify DOM values in loading screen, when the new values are actually different from the current ones.

This avoids bothering requests in the network tab in dev tools with the cache disabled as well as elements blinking in an intermittent way in elements tab.


BrianAmadori (approved):
Network panel looks way cleaner now. Thanks!

sandrade-dcl (approved):

PR #1132, opened 2020-07-22
chore: use action payload to report scenes around parcel

pbosio:


pravusjif (approved):
looks good, worked fine for me

sandrade-dcl (approved):

PR #1133, opened 2020-07-22
fix: early rendering activation

pravusjif:
- Fixed first time currentScene setup to add the missing rendering lock

- Added fix for unneeded extra scenes sorting calls


sandrade-dcl (approved):

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -191,6 +193,10 @@ private void SortScenesByDistance()
                     if (scene.sceneData.id != globalSceneId && characterIsInsideScene)
                     {
                         currentSceneId = scene.sceneData.id;
+
+                        if (firstSort)
```
BrianAmadori, line 197: Not a fan of this approach because suddenly, the scenes sorting method has the loading lock responsibility. Ideally i'd think of a way to keep this responsibility outside.

PR #1136, opened 2020-07-23
feat: explore -part1- fetch hotscenes

pbosio:
*fetch scenes across realms with people in it
*added hidden chat command to show the first 5 scenes with people in it (helpful for debugging porpoises and meeting new people :P)


kernel/packages/shared/social/hotScenes.ts
```diff
@@ -0,0 +1,123 @@
...
+    baseCoord: baseCoord.split(',').map((str) => parseInt(str, 10)) as ParcelArray,
+    scene: sceneJsonData,
+    usersCount: 1
+  }
+}
+
+async function getCrowdedScenesFromCandidate(candidate: Candidate): Promise<CandidateCrowdedScene[]> {
+  let scenes: Record<string, CandidateCrowdedScene> = {}
+
+  const tiles = candidate.layer.usersParcels?.map((value) => `${value[0] | 0},${value[1] | 0}`) ?? []
```
moliva: If `usersParcels` is populated, its values ought to be properly initialized. If still, we want to catch any issue with them I would filter out the parcel before taking it into account, as we don't actually know the proper coordinate.

kernel/packages/shared/social/hotScenes.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  for (const tile of tiles) {
+    let id = (await fetchSceneIds([tile]))[0]
+    let land = null
+
+    if (id) {
+      land = (await fetchSceneJson([id]))[0]
```
moliva: Might prove better to fetch both ids and data for the whole set of tiles before iterating though the collection.

moliva (changes requested):

moliva (approved):
Great work! 😄

pablitar (approved):
LGTM!

PR #1137, opened 2020-07-24
fix: avatar glitches when wearables are missing from profile

BrianAmadori:
Before:
image

After:
image


sandrade-dcl (approved):
Code reviewed and the fix tested in the build. It is working fine!

pravusjif (approved):
awsomeeeeee

pbosio (approved):

PR #1140, opened 2020-07-28
feat: explore -part2- unity controller and messaging with kernel

pbosio:


kernel/packages/shared/social/hotScenes.ts
```diff
@@ -4,10 +4,17 @@ import { StoreContainer } from 'shared/store/rootTypes'
 import { fetchSceneIds } from 'decentraland-loader/lifecycle/utils/fetchSceneIds'
 import { fetchSceneJson } from 'decentraland-loader/lifecycle/utils/fetchSceneJson'
 import { SceneJsonData } from 'shared/types'
+import { reportScenesFromTiles } from 'shared/atlas/actions'
 import { getSceneNameFromAtlasState, postProcessSceneName } from 'shared/atlas/selectors'
 
 declare const globalThis: StoreContainer
 
+declare const window: {
```
moliva, line 12: You could use `UnityInterfaceContainer` from `dcl.ts` here instead.

moliva (approved):

kernel/packages/shared/social/hotScenes.ts
pbosio: > You could use `UnityInterfaceContainer` from `dcl.ts` here instead.

Wouldn't that end up in a circular dependency issue?

moliva: Indeed, good catch.

kernel/packages/shared/social/hotScenes.ts
```diff
@@ -30,6 +37,12 @@ type CandidateCrowdedScene = {
   usersCount: number
 }
 
+export type HotSceneInfo = {
+  baseCoords: { x: number; y: number }
+  usersTotalCount: number
```
pravusjif, line 33: maybe it's more useful to have a collection of the user-IDs/addresses that are inside the scene?
(we can later calculate/take the count() on that collection)

pbosio: > maybe it's more useful to have a collection of the user-IDs/addressed that are inside the scene?
> (we can later calculate/take the count() on that collection)

Wouldn't be better to have a separate api to get all connected players and their location?
Anyway we can also add that when we need it, do we need that now?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HotScenesController/HotScenesController.cs
```diff
@@ -0,0 +1,70 @@
...
+        if (updatePayload.chunkIndex == 0)
+        {
+            isUpdating = true;
+            hotScenesList.Clear();
+        }
+
+        hotScenesList.AddRange(updatePayload.scenesInfo);
+        OnHotSceneListChunkUpdated?.Invoke();
+
+        if (hotScenesList.Count >= updatePayload.totalScenes)
```
pravusjif: if we send "chunksCount" or "lastChunkIndex" (instead of `totalScenes`) we can compare `updatePayload.chunkIndex >= lastChunkIndex` instead of accessing the more costly `hotScenesList.Count`.

But i doubt the current approach would affect performance at all.

kernel/packages/shared/social/hotScenes.ts
pravusjif: nah, it's fine.

pravusjif (approved):

kernel/packages/shared/social/hotScenes.ts
BrianAmadori: We can't move `UnityInterfaceContainer` to a new file to solve the circular dependency and be able to use it nicely?

pbosio: > We can't move `UnityInterfaceContainer` to a new file to solve the circular dependency and be able to use it nicely?

if we choose to go that way, we should create another PR for that... cause, as far as I can see, we have multiple scripts doing this same thing, so we should create a PR to fix'em all 😄

kernel/packages/shared/social/hotScenes.ts
```diff
@@ -4,10 +4,17 @@ import { StoreContainer } from 'shared/store/rootTypes'
...
 
 declare const globalThis: StoreContainer
 
+declare const window: {
+  unityInterface: {
+    UpdateHotScenesList: (info: HotSceneInfo[]) => void
+  }
+}
+
 type CrowdedSceneRealmInfo = {
```
BrianAmadori: Why we have the concept of `CrowdedScenes` here? Maybe it would unify the naming better if we call this type `HotSceneRealmInfo`?

BrianAmadori (approved):

kernel/packages/shared/social/hotScenes.ts
BrianAmadori: O.K. by me, but if the improvement is feasible without much hassle and makes sense, pretty please let's open another PR and fix it. I'm a bit concerned about not improving anything because of bureaucracy.

PR #1144, opened 2020-07-30
fix: private chat not displaying timestamp toast

pbosio:
A reference was missing from the prefab causing a Null Reference Exception


RominaMarchetti (approved):

sandrade-dcl (approved):

pravusjif (approved):
thx for fixing this!

PR #1145, opened 2020-07-30
fix: chat heads container not expanding with it content

pbosio:
This issue make impossible to add new buttons to the taskbar.

Screen Shot 2020-07-29 at 19 42 51

* added content size fitter to the chat heads container
* force container's parent layout to update when content change
* hide container if there is no content to show

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/ChatHeadGroupView.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/ChatHeadGroupView.cs
@@ -18,6 +18,7 @@ public class ChatHeadGroupView : MonoBehaviour
     private IChatController chatController;
     private IFriendsController friendsController;
     private ulong rendererStateTimeMark;
+    private RectTransform contentParentRT;
 
     public void Initialize(IChatController chatController, IFriendsController friendsController)
     {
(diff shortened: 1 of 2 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/ChatHeadGroupView.cs
```diff
@@ -38,6 +39,19 @@ public void Initialize(IChatController chatController, IFriendsController friend
...
+    {
+        contentParentRT = container.parent as RectTransform;
+        gameObject.SetActive(false);
+    }
+
+    private void Update()
+    {
+        // NOTE: Update is only enabled when SetParentContainerAsDirty is called 
+        DCL.Helpers.Utils.ForceUpdateLayout(contentParentRT, delayed: true);
+        this.enabled = false;
```
sandrade-dcl, line 52: Only a question: Why can't we do this directly in the `SetParentContainerAsDirty()` function, instead of here in the `Update()` event?

sandrade-dcl (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/ChatHeadGroupView.cs
pbosio: > Only a question: Why can't we do this directly in the `SetParentContainerAsDirty()` function, instead of here in the `Update()` event?

Is just to avoid the case where multiples chat heads are added and set the layout as dirty. This way we ensure to only call it once and avoid the performance cost of the operation

sandrade-dcl: Ok, thanks! you have my approve! :)

sandrade-dcl (approved):

pravusjif (approved):

BrianAmadori (approved):

PR #1147, opened 2020-07-30
fix: reconnection issues & realm optimization

moliva:
# What? 
This PR comprises of two changes:
- Fixes reconnection strategy for comms initial connection
- Triggers a reconnection whenever the DAO scan is completed to avoid using a cached realm when a better option is available (it still honors the realm configured by the user)


pbosio (approved):

pablitar (approved):
Great work!

moliva:
🚀

PR #1150, opened 2020-07-31
feat: shadergraph toon shader

ShibuyaMartin:
- [X] Main shader integration
- [x] Parametrize shader to have different light positions on different contexts (avatar editor, in-world, halloween mode)
- [x] Fix transparencies (extract source code and make blending modes parametrizable)


unity-client/Assets/Rendering/Utils/ShaderUtils.cs
```diff
@@ -0,0 +1,40 @@
+﻿using UnityEngine;
+
+namespace DCL.Helpers
+{
+    public static class ShaderUtils
+    {
+        public static readonly int _LightDir = Shader.PropertyToID("_LightDir");
```
pbosio, line 7: Just an observation here: Is it possible to remove the underscore prefix?
I know this class already existed and was promoted to it own file and that fields are named to match shader's property name.
But it doesn't match any of our notations in this project, plus, even if it's not standardized by Microsoft, a really great number of C# devs reserve the underscore prefix to distinguish private fields

pbosio (approved):

unity-client/Assets/Rendering/Utils/ShaderUtils.cs
pravusjif: I'd like this too XD

pravusjif (approved):

unity-client/Assets/Rendering/Utils/ShaderUtils.cs
BrianAmadori: Yeah, your arguments make sense @pbosio . Just let's discuss this as another issue and dedicate another PR to it 👍

PR #1154, opened 2020-08-03
feat: audio implementation

Landeplage:
Initial implementation of audio for HUD-elements. Currently only the avatar menu, avatar editor and the settings menu have sounds.


sandrade-dcl (approved):
I have found some missing sounds:
- If we open the world chat with the Enter key, there is not sound.
- The [U] key (to show/hide the UI) has not sound assigned.
- If we change the sound toggle in our Settings window, the audio is not being muted.

We can merge this PR and go fixing these kind of things on next PR's.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/Audio/HUDAudioPlayer.cs
```diff
@@ -0,0 +1,190 @@
...
+        fadeIn,
+        fadeOut,
+        notification,
+        sendChatEntry,
+        receivePrivateChatEntry,
+        receiveGlobalChatEntry,
+        cameraToThirdPerson,
+        cameraToFirstPerson,
+        tinyHover
+    }
```
BrianAmadori, line 29: We will have to do some work here because this doesn't look like it scales well. After looking around the code, I think we could have some sort of audio event specification as an `ScriptableObject`, so we pass the SO reference instead of this enum and the `GetEvent` declarations below.

To exemplify, let's say we have a `HUDStuffView` that needs to play sound `beep`:
- We create an SO object and place it somewhere on the project 
- We add this SO to the `AudioContainer` class
- We add a SO reference to the `HUDStuffView` prefab. Let's call this reference is called `beepSound`
- Then we can just do `HUDAudioPlayer.i.Play( beepSound );`

This would save us the maintenance of this file and make the system very scalable. I understand you may want to do something specific to the sounds, like in this case:
```
                if (!listItemAppearHasPlayed)
                {
                    eventListItemAppear.SetPitch(listItemAppearPitch);
                    eventListItemAppear.Play(true);
                    listItemAppearPitch += 0.15f;
                    listItemAppearHasPlayed = true;
                }
```

For handling those specific cases, you can just do it like this:

Let's say you have the simple `ScriptableObject` definition for your `AudioEvent`:

```
public class AudioEvent : ScriptableObject
{
    public void Play( bool oneShot ) 
    {
        ...
    } 
}
```

We can make the `Play` method virtual, like this...
 
```
   public virtual void Play( bool oneShot )
   {
        ... play code ...
   }
```

We can later implement an `AudioEvent` child, and put our custom stuff there, like this:

```
public class AudioEvent_PitchUp : AudioEvent 
{
    int lastPlayedFrame;
    float pitch;
    public float pitchIncrement = 0.15f;

    public override void Play(bool oneShot)
    {
            if (lastPlayedFrame != Time.frameCount)
            {
                SetPitch(pitch);
                base.Play(true);
                pitch += pitchIncrement;
                lastPlayedFrame = Time.frameCount;
            }
    }
}
```

BrianAmadori (approved):
Awesome work! 👍

Issue #847, opened 2020-04-27
Cross imports between modules -> module is not a constructor


nearnshaw:
This is a problem I've been experiencing for years, but it's always been elusive and hard to reproduce. I think I now have a better understanding of when it happens.

Two things need to happen, for this problem to appear:

- There have to be imports in game.ts from a second .ts file
- The second .ts file needs to import an external library, like Identity or web3-provider

If these two things happen, then the second .ts file is not loaded, and you get an error message in console saying that the file "is not a constructor":

![Screen Shot 2020-04-27 at 14.38.52.png](https://images.zenhubusercontent.com/5d9b940e491c060001c8647b/b6ae8f82-f38f-48ea-8502-8785a16716ea)

Here's a minimal example that Carl Fravel put together to show this problem in action:

[https://app.zenhub.com/files/220994476/b53e73d5-82c7-46aa-a25e-7f7caca11cf9/download](https://app.zenhub.com/files/220994476/b53e73d5-82c7-46aa-a25e-7f7caca11cf9/download)

In DemoClass.ts, I added some commented lines at the end that show how the problem can also happen if you import a different library, like Identity.

If you rewire things so that there are no cross imports between the two .ts files, you can avoid it. If I put the same cod that's in DemoClass.ts in game.ts, then it works

menduz:
Does it show any errors before that one? like "these modules did not load"?

nearnshaw:
@menduz  I just double checked and there are no previous error messages to this one

There are a LOT of junk messages that seem to be irrelevant (+ than 3 whole pages), but nothing that's an error, and nothing that references the module name


<img width="683" alt="Screen Shot 2020-05-11 at 15 41 44" src="https://user-images.githubusercontent.com/3507907/81598615-f9e67500-939d-11ea-996e-c3dc894c6dd5.png">

<img width="681" alt="Screen Shot 2020-05-11 at 15 41 31" src="https://user-images.githubusercontent.com/3507907/81598598-f2bf6700-939d-11ea-8e1e-b17bddcdd69a.png">

<img width="688" alt="Screen Shot 2020-05-11 at 15 39 19" src="https://user-images.githubusercontent.com/3507907/81598501-cad00380-939d-11ea-9f03-b91d491ed9be.png">

PR #1157, opened 2020-08-04
fix: Issue with module imports

pablitar:
Fixes (#847)
- When you had multiple files in a scene, and a file referenced an internal dependency, you'd get an "XXXClass is not a constructor" error:
- Changed makefile and build process to add dcl-amd code from this same repo

Bunch of other fixes were added after some testing with existing scenes:
- Extracted exports from modules for it to not be contaminated by dcl-amd data
- Added some missing calls to handlers
- Added a check in scene system to wait for external modules to load before sending "initMessagesFinished". This should avoid showing the world before the scene has finished loading in some cases
- Fixed issue with relative paths
- Added support for "index" modules


kernel/packages/decentraland-amd/src/amd.ts
```diff
@@ -1,264 +1,346 @@
...
+    let i = idParts.length
     while (--i) {
-      switch (id[0]) {
-        case "..":
-          newContext.pop();
-        case ".":
-        case "":
-          idParts.shift();
-          changed = true;
+      switch (idParts[0]) {
```
nchamo, line 319: Nice catch!

kernel/packages/decentraland-amd/src/amd.ts
```diff
@@ -1,264 +1,346 @@
...
+        if (typeof dcl !== 'undefined') {
+          dcl.loadModule(moduleName).then((descriptor: ModuleDescriptor) => {
+            let createdModuleExports = {}
+
+            for (let i in descriptor.methods) {
+              const method = descriptor.methods[i]
+              createdModuleExports[method.name] = createMethodHandler(descriptor.rpcHandle, method)
+            }
+
+            // This is somewhat repeated with the ready function above... Should refactor with clear head
```
nchamo, line 275: Agree, they are pretty similar

nchamo (approved):
Daamn Pablo, great work!

pbosio (approved):

PR #1159, opened 2020-08-04
fix: disable console log for scenes in production

BrianAmadori:
This PR should me merged after `fix/native-messages-layer-part-1`

In some areas there are scenes that output logs like no tomorrow. This clogs up the CPU like horse manure on a pipe. Here's the fix.


BrianAmadori:
Closed and re-opened with a cleaner commit count here: https://github.com/decentraland/explorer/pull/1164

PR #1161, opened 2020-08-05
feat: fixed canvas resolution

BrianAmadori:
Use `unityInterface.SetTargetHeight(<max-height>)` in dev tools console to play with values. Default is 720.


pravusjif:
Tested walking a specific path inside Genesis Plaza, with medium settings, on a macbook pro chrome resolution 2880x1800:

in **master** I'm getting hiccups in the last 1k frames as high as 880
Screen Shot 2020-08-06 at 10 42 05

in **this PR** I'm getting hiccups in the last 1k frames as high as 220
Screen Shot 2020-08-06 at 10 39 41

So I'd say the implementation is definitely an improvement in performance, but of course the change is pretty noticeably in the UI and postprocesses
Screen Shot 2020-08-06 at 10 52 22
Screen Shot 2020-08-06 at 10 55 36

However in 1080p it looks fine and still has much better performance values than the native retina display resolution:
Screen Shot 2020-08-06 at 10 57 59

So we should definitely have this setting somehow in the settings panel, and the quality presets could also use it

pravusjif:
Maybe it's better to have the 1080p as default (instead of 720), for high resolutions

BrianAmadori:
Re-opened as another PR with cleaner commits https://github.com/decentraland/explorer/pull/1165

PR #1163, opened 2020-08-07
fix: link errors to source map

nchamo:
Before this change, when an error happened, it wasn't possible to understand the cause:
Screen Shot 2020-08-07 at 14 13 20

Clicking on the link would take you nowhere, and it was hard to debug issues.
We realized that the source map was actually present, but the `sourceURL` parameter was not set. Therefore, Chrome would show it as anonymous, and it wouldn't provide a link to the actual source map.

By simply adding this extra parameter, now a link is provided to the actual code:
Screen Shot 2020-08-07 at 14 12 14

It is important to mention that the code is the generated javascript, and is not the same as the actual `.ts` files generated by content creators.

We might want to consider taking the time to implement a more robust, user-friendly solution that generates these scene files.


pablitar (approved):
Nice work!

nchamo:
@EibrielInv, forgot to cc you. On the forum you asked for something similar.

Even though it's not as helpful as a I would like it to be, it's definitely better than before

PR #1166, opened 2020-08-10
fix: Improve performance of UI updates (Part I)

sandrade-dcl:
Improve the performance related to the hiccups provoked by the UI updates (part 1).


pravusjif (approved):
beautiful, works fine at Salmonomicon and Tomb Chaser

pbosio (approved):

BrianAmadori (approved):
As we discussed in pm, visual tests were not being evaluated correctly. But when fixed they passed! So this is a ✔️

PR #1174, opened 2020-08-12
chore: Added script to fetch scene contents

pablitar:
Added script to fetch scene contents, for testing purposes


pbosio (approved):

kernel/scripts/fetchSceneContents.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+type Content = {
+  file: string
+  hash: string
+}
+
+async function main() {
+  const query = sceneId ? `id=${sceneId}` : `pointer=${pointer}`
+  const url = `${contentServerUrl}/entities/scene?${query}`
```
nchamo: We want to deprecate this API.

Let's try to use `/deployments`. You can use `entityType=scene` and `&entityId={sceneId}` or `&pointer={pointer}`

kernel/scripts/fetchSceneContents.ts
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * The scene will be downloaded from https://peer.decentraland.org/content. This can be overriden with the CONTENT_SERVER_URL variable
+ * The scene will be downloaded in the public/downloaded-scenes folder. This can be overriden with the OUTPUT_DIR variable
+ */
+
+import fetch from 'node-fetch'
+import * as fs from 'fs'
+
+const sceneId = process.env.SCENE_ID
+const pointer = process.env.POINTER
```
nchamo: Nit: Maybe calling it parcel would be clearer?

kernel/scripts/fetchSceneContents.ts
nchamo: Actually, you could use the catalyst-client (I think it's already a dependency for kernel)

The API:
https://github.com/decentraland/catalyst-client

nchamo (approved):
Great work!!!

PR #1178, opened 2020-08-13
fix: lazy physics update

BrianAmadori:
Disabled `Physics.autoSimulation` and `Physics.autoSyncTransforms`. Using manual `Physics.SyncTransform` calls instead to avoid unwanted physics hiccups.

What to test:

- [x] Animation-based moving platforms have to work (i.e. air balloon and elevators at genesis plaza)

- [x] When the world adjust its position, the character shouldn't teleport or do anything weird (just walk straight for long time to check this)

- [x] System-based moving platforms have to work (any platforming game, i.e. dcl train)


sandrade-dcl (approved):
Awesome! :) I have tested it (with Unity Profiler) and I have been able to confirm that all the annoying spikes we had related to physics have dessapeared. Good job!

pravusjif (approved):
works great

PR #1179, opened 2020-08-13
fix: scene's ui rendering over client's ui

pbosio:
Explicitly set canvas' sortingOrder instead of depending on it default value. It was set to -1 just in case but setting it to 0 also works.

BEFORE: https://play.decentraland.org/?position=-55%2C1 or https://play.decentraland.zone/?ENV=org&position=-55%2C1
TEST LINK: https://play.decentraland.zone/branch/fix/scene-ui-over-dcl/index.html?ENV=org&position=-55%2C1

NOTE: Can also be reproduced opening the "controls" HUD (C key)
THE ISSUE
Screen Shot 2020-08-13 at 14 54 56


pravusjif (approved):
finally!!!

sandrade-dcl (approved):

PR #1188, opened 2020-08-18
chore: re-enabled email request popup

pbosio:
Popup should appear a minute after the renderer is enabled.
Use RESET_TUTORIAL flag to reset tutorialStep if needed.
Popup should prompt to all users, so, for testing porpoises don't use that flag the first time.


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/EmailPromptHUD/EmailPromptHUDController.cs
```diff
@@ -1,34 +1,142 @@
...
+    }
 
-        inputField.onValueChanged.AddListener(value =>
+    void CreateTimer()
+    {
+        if (!isEnable)
+            return;
+
+        isTimerCreated = true;
+        timerRoutine = CoroutineStarter.Start(WaitForSeconds(60));
```
sandrade-dcl: It would be fine to define a constant for this value of 60 seconds 👍

sandrade-dcl (approved):
I guess there is a reason for the delay of 60 seconds to show the pop-up, right? For me it is a bit weird... Otherwise, everything's fine! :)

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/EmailPromptHUD/EmailPromptHUDController.cs
pbosio: You are right. I will create a constant for that. Thanks

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/EmailPromptHUD/EmailPromptHUDController.cs
```diff
@@ -1,34 +1,144 @@
...
-    public TMP_InputField inputField;
-    public GameObject cryptoUserMessage;
-    public GameObject nonCryptoUserMessage;
-    public Button sendButton;
+    const float POPUP_DELAY = 60;
 
-    void Awake()
+    EmailPromptHUDView view;
+
+    bool isRendererEnabled = false;
```
BrianAmadori: Why we are using this bool instead of querying the renderer state `ScriptableObject`?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/EmailPromptHUD/EmailPromptHUDController.cs
```diff
@@ -1,34 +1,144 @@
...
+    bool isTimerComplete = false;
+    bool isTimerCreated = false;
+    bool isEnable = false;
```
BrianAmadori:
```suggestion
    bool isEnabled = false;
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/EmailPromptHUD/EmailPromptHUDController.cs
```diff
@@ -1,34 +1,144 @@
...
+    void CreateTimer()
     {
-        bool hasWallet = UserProfile.GetOwnUserProfile().hasConnectedWeb3;
+        if (!isEnable)
+            return;
 
-        cryptoUserMessage.SetActive(hasWallet);
-        nonCryptoUserMessage.SetActive(!hasWallet);
+        isTimerCreated = true;
+        timerRoutine = CoroutineStarter.Start(WaitForSeconds(POPUP_DELAY));
```
BrianAmadori: I have a few issues with the approach of this class:

* Too many flags: having logic based on a bunch of flags that can be true or false always lead to issues maintaining/reading the code down the road. This is because you lose track of the state, and the code flow is unclear.

* If we are already starting a coroutine here, and the code is dependant on it, why not just make the whole logic run on a single IEnumerator instead of all those flags flying around? like this:

```
IEnumerator StartEmailLogic_FindBetterName()
{
    yield return WaitForSecondsCache.Get(seconds);
    yield return WaitUntil( () => rendererIsEnabled );
    ShowEmailPopup();
}
```

BrianAmadori (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/EmailPromptHUD/EmailPromptHUDController.cs
pbosio: when you are right you are right 😄

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/EmailPromptHUD/EmailPromptHUDController.cs
```diff
@@ -1,34 +1,109 @@
...
+        isPopupRoutineRunning = false;
+    }
 
-        cryptoUserMessage.SetActive(hasWallet);
-        nonCryptoUserMessage.SetActive(!hasWallet);
+    IEnumerator ShowPopupDelayed(float seconds)
+    {
+        isPopupRoutineRunning = true;
+        yield return new WaitUntil(() => CommonScriptableObjects.rendererState.Get());
+        yield return WaitForSecondsCache.Get(seconds);
```
BrianAmadori, line 84: What if i'm in the loading state when the time passes? Maybe we can add another `rendererState` check?

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/EmailPromptHUD/EmailPromptHUDController.cs
pbosio: Yes, I can add that. I didn't thought it would make much of a difference if the popup is displayed while the renderer is disabled

PR #1192, opened 2020-08-20
Initial refactor for SceneController.

marcosnc:


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Environment.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace DCL
+{
+    public class Environment : Singleton<Environment>
+    {
+        public MessagingControllersManager messagingControllersManager { get; }
+
+        /*
+         * TODO: Continue moving static beans to this class
```
BrianAmadori: We should avoid nasty Java concepts like "beans" out of this repo 😛

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Environment.cs
```diff
@@ -0,0 +1,34 @@
...
+         * 
+        MemoryManager memoryManager;
+        PointerEventsController pointerEventsController;
+        ParcelScenesCleaner parcelScenesCleaner; // This is a static member of ParcelScene
+        PoolManager poolManager; // This should be created through a Factory, and that factopry should execute the code in the method EnsureEntityPool
```
BrianAmadori, line 15: We should remove those comments.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingBus.cs
```diff
@@ -183,13 +181,13 @@ public void Enqueue(QueuedSceneMessage message, QueueMode queueMode = QueueMode.
 
                 if (type == MessagingBusType.INIT)
                 {
-                    manager.pendingInitMessagesCount++;
+                    Environment.i.messagingControllersManager.pendingInitMessagesCount++;
```
BrianAmadori, line 184: Please keep the `manager` reference. This is cached for a reason, this is a performance critical code and believe it or not, the `i` getter was consuming a few ms in high stress cases.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -127,120 +58,18 @@ void Awake()
                 WebInterface.StartDecentraland();
             }
 
-            ParcelScene.parcelScenesCleaner.Start();
+            ParcelScene.parcelScenesCleaner.Start(); // Cambiar por la inicialización del Enviroment.parcelScenesCleaner
```
BrianAmadori: Remove all those spanish comments please.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -314,274 +126,220 @@ private void Update()
             }
         }
 
-        public void CreateUIScene(string json)
+        public void EnsureEntityPool()  // Mover al PoolManagerFactory
```
BrianAmadori, line 129: Remove spanish comment

BrianAmadori (changes requested):
Congratz on your first Unity PR! 🙂 

Left some minor requests.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingBus.cs
marcosnc: I'm strongly against doing that. Having a bean reference means that reference can be changed and it will make the code harder to read.
On the other hand, using a local reference vs using a remote reference should not have any impact on performance. I would recommend we test it to be sure, but should not produce any difference.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
marcosnc: You're right! Sorry about this!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingBus.cs
marcosnc: You were right! I've made some performance test and then decided to change some structures here.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Environment.cs
marcosnc: I would prefer to keep them until we move all the singleton instances to this class. They are just here to remember what we need to do.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingBus.cs
marcosnc: After making some changes, now the non-local access is as fast as the local one. Please see the changes in the static variable in Environment, the removal of the accessor for the messagingControllerManager member, and also the changes made to Singleton.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Environment.cs
BrianAmadori: Fair enough.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Environment.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public readonly MessagingControllersManager messagingControllersManager;
+
+        /*
+         * TODO: Continue moving static instances to this class. Each static instance should be converted to a local instance inside this class.
+         *
+        MemoryManager memoryManager;
+        PointerEventsController pointerEventsController;
+        ParcelScenesCleaner parcelScenesCleaner; // This is a static member of ParcelScene
+        PoolManager poolManager; // This should be created through a Factory, and that factopry should execute the code in the method EnsureEntityPool
```
BrianAmadori, line 15:
```suggestion
        PoolManager poolManager; // This should be created through a Factory, and that factory should execute the code in the method EnsureEntityPool
```

BrianAmadori (approved):
Good job!. Make sure you give the build a whirl before merging, so we make sure the `Singleton<T>` change didn't break anything.

PR #1193, opened 2020-08-20
feat: add ability to attach entity to player

nchamo:
# What is this PR doing?
Entities can be attached to each other by using the `setParent` method. However, there are cases where we might want to attach entities to other objects that are not entities created by the same scene (for example, the player's avatar). This PR does just this.

We are now allowing entities to be attached to two new objects:
* The player's avatar: `AvatarPosition`
* The player itself: `Player`

In the first case, the entity is attached to the avatar, and moves with it. In the case of the `Player`, the entity is also attached to the avatar, but it rotates with the camera when the first person mode is activated.

# Implementation details
## SDK
1. We created two new global entities (the player and the avatar), each with their own fixed id. This is because we had to make too many changes to support the entity hierarchy without entities.
1. We also created the concept of `Attachable`, so we can add new objects in the future.

## Unity
1. We created a new `CommonScriptableObject` called `CameraMode`. This allowed us to listen to changes to the camera mode, without having to modify the `CameraController` itself.
1. We added two new gameobjects to the CharacterController. One represents the `AvatarPosition` and the other represents the `Player`
1. We modified the `boundariesChecker`, so that we can add entities to it that are constantly checked. 
1. At first, we had the `Player` gameobject switch between the `CharacterController` and the `MainCamera` in terms of who the parent was. This change happened when the camera mode changed. The problem was that there is an animation that occurs when switching from third person mode, to the first person. So entities were pushed forward. So instead of changing parents, we just listen to changes on the camera rotation.


unity-client/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/Tests/UUIDComponentTests.cs
```diff
@@ -28,7 +28,7 @@ protected override IEnumerator SetUp()
             DCLCharacterController.i.characterController.enabled = false;
 
             cameraController.SetRotation(0, 0, 0, new Vector3(0, 0, 1));
-            cameraController.SetCameraMode(CameraStateBase.ModeId.FirstPerson);
+            //cameraController.SetCameraMode(CameraMode.ModeId.FirstPerson);
```
nchamo: I had to comment this, because I couldn't reference the `CameraMode` from here. I'm not sure how to make it accesible

kernel/packages/decentraland-ecs/src/ecs/Attachable.ts
```diff
@@ -0,0 +1,17 @@
+import { IEntity, IEngine } from "./IEntity"
+
+/**
+ * Entities can be attached to each other by using the `setParent` method. However, there are cases where we might want to attach entities
+ * to other objects that are not entities created by the same scene (for example, the player's avatar). For those cases, we have this class.
+ */
+export abstract class Attachable {
```
pablitar, line 7: Why use an abstract class and not an Interface?

Is it to be able to use `Attachable.AVATAR`?

kernel/packages/decentraland-ecs/src/ecs/Entity.ts
```diff
@@ -279,8 +280,19 @@ export class Entity implements IEntity {
...
+    let newParent: IEntity | null
+
+    // Check if parent is of type Attachable
+    if (_parent && 'getEntityRepresentation' in _parent) {
+      if (!this.engine) {
+        throw new Error(`In order to set an attachable as parent, you first need to add the entity to the engine.`)
+      }
+      newParent = _parent.getEntityRepresentation(this.engine)
+    } else {
+      // @ts-ignore
```
pablitar, line 293:
```suggestion
```
I think this is not necessary.

kernel/packages/decentraland-ecs/src/ecs/Entity.ts
```diff
@@ -305,22 +317,23 @@ export class Entity implements IEntity {
       delete currentParent.children[this.uuid]
     }
 
+    // Make sure that the parent and child are both on the engine, or off the engine, together
     if (newParent !== null && newParent.uuid !== '0') {
       if (!newParent.isAddedToEngine() && this.isAddedToEngine()) {
         // tslint:disable-next-line:semicolon
-        ;(this.engine as IEngine).removeEntity(this)
+        (this.engine as IEngine).removeEntity(this)
```
pablitar: I know you didn't actually change this code, but we may take the opportunity :)
```suggestion
        this.engine!.removeEntity(this)
```

kernel/packages/decentraland-ecs/src/ecs/Entity.ts
```diff
@@ -305,22 +317,23 @@ export class Entity implements IEntity {
...
       }
       if (newParent.isAddedToEngine() && !this.isAddedToEngine()) {
         // tslint:disable-next-line:semicolon
-        ;((newParent as Entity).engine as IEngine).addEntity(this)
+        ((newParent as Entity).engine as IEngine).addEntity(this)
```
pablitar: I know you didn't actually change this code, but we may take the opportunity :)
```suggestion
        (newParent as Entity).engine!.addEntity(this)
```

kernel/packages/decentraland-ecs/src/ecs/Entity.ts
pablitar: Also, I think you can delete the `// tslint:disable-next-line:semicolon` comments

unity-client/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/Tests/UUIDComponentTests.cs
pablitar: How does this affect the tests?

kernel/packages/decentraland-ecs/src/ecs/Entity.ts
nchamo: I can't remove this. The build will fail, because it thinks that `_parent` can be `Attachable`

kernel/packages/decentraland-ecs/src/ecs/Attachable.ts
nchamo: No, I can only use it like `Attachable.AVATAR` with abstract classes

pablitar (approved):
LGTM!

kernel/packages/decentraland-ecs/src/ecs/Attachable.ts
pablitar: Yep, that's what I said :D

nchamo: I was trying to confirm it, but I clearly didn't phrase it correctly haha

unity-client/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/Tests/UUIDComponentTests.cs
BrianAmadori: I think the OnClick tests expect the first person mode, be wary with this change.

BrianAmadori: Make sure the assembly definition of this test (the .asmdef file in the same folder of the project, modificable via Unity's inspector) is referencing the `CommonScriptables` assembly.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/Attachable/PlayerReference.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    protected override void Cleanup()
+    {
+        CommonScriptableObjects.cameraMode.OnChange -= OnCameraModeChange;
+        CommonScriptableObjects.cameraForward.OnChange -= UpdateForward;
+    }
+
+    private void UpdateForward(Vector3 newFordward, Vector3 prev)
+    {       
+        transform.forward = newFordward;
```
BrianAmadori:
```suggestion
        transform.forward = newForward;
```

kernel/packages/decentraland-ecs/src/ecs/Entity.ts
pablitar: Weird. Locally with `make watch` it seems to work

nchamo: This is what failed
https://app.circleci.com/pipelines/github/decentraland/explorer/4918/workflows/d35fcba1-172d-4318-9d7a-fa97e81db84e/jobs/30707

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/Attachable/AvatarReference.cs
```diff
@@ -0,0 +1,23 @@
+using UnityEngine;
+using DCL.Models;
+using DCL.Controllers;
+using System.Collections.Generic;
+
+public class AvatarReference : Attachable
+{         
+
+    public static AvatarReference i { get; private set; }
```
BrianAmadori: Not sure if i'm convinced of this singleton setup. We can't put these under an Attachable manager or put the references under another object?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -473,6 +473,20 @@ public void SetEntityParent(string entityId, string parentId)
                 return;
             }
 
+            if (me != null && parentId == "PlayerEntity")
```
BrianAmadori: As this is a performance critical path, maybe we can group the `me != null` checks? Any small performance saving here is gold.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/Attachable/AvatarReference.cs
nchamo: Can you point me to another case where we do any of those approaches? Just so I can see how you do it

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
nchamo: Done!

kernel/packages/decentraland-ecs/src/ecs/IEntity.ts
```diff
@@ -41,7 +44,7 @@ export interface IEntity {
 
   isAddedToEngine(): boolean
   getParent(): IEntity | null
```
pravusjif, line 46: shouldn't `getParent()` support `Attachable` as the `setParent()` ?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/Attachable/Attachable.cs
```diff
@@ -0,0 +1,54 @@
...
+        entity.gameObject.transform.SetParent(transform, false);
+        entity.OnRemoved += OnEntityRemoval;        
+    }
+
+    protected virtual void Setup() { }
+    protected virtual void Cleanup() { }
+
+    private void Awake()
+    {
+        // Setup
```
pravusjif: Since the code is clear enough, I'd remove the following redundant comments:
`// Setup`
`// Listen to changes on the player position`
`// Stop listening to changes`
`// Clean up`

kernel/packages/decentraland-ecs/src/ecs/IEntity.ts
nchamo: Actually, we can't 😢  

Since `getParent` is used to traverse across the hierarchy in different parts of the SDK, an entity (or `IEntity`) needs to be returned.

It is definitely inconsistent, but is one of the things we need to change and do better with the next version of the SDK

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/Attachable/AvatarReference.cs
pravusjif: I think Brian means not having this monobehaviour at all, just having the gameobject or transform (of the gameobject that currently holds this monobehaviour) as a reference in some other class public property or similar.

I'd suggest doing that.

For example maybe in a property in SceneController (and put the gameobject, that will be referenced, as a child of the SceneController), but since we are refactoring the SceneController, not sure that's the best place

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/Attachable/PlayerReference.cs
```diff
@@ -0,0 +1,52 @@
+using UnityEngine;
+using DCL.Models;
+using DCL.Controllers;
+using System.Collections.Generic;
+
+public class PlayerReference : Attachable
+{         
+
+    public static PlayerReference i { get; private set; }
```
pravusjif: Same as in `AvatarReference`, this monobehaviour could be a non-singleton and be a reference in another accessible object. (if we take that approach on the `AvatarReference` I'd suggest making this change in this monobehaviour as well, not deleting the monobehaviour like in `AvatarReference`, just removing the singleton functionality).

pravusjif (changes requested):

pravusjif:
This may sound silly but it's a super minor change: could we rename the "PlayerReference" GameObject to "PlayerEntityReference" and "AvatarReference" to "AvatarEntityReference"?

Sorry, it's just that I really think that will be less confusing (also for future devs coming to the project), as we also have the "Player" GameObject, "CharacterController" GameObject, "AvatarRenderer" GameObject...

For example someone may ask "what's the difference between the Player gameobject and the PlayerReference gameobject?"... if it's called "PlayerEntityReference" I think it's clearer that it means an SDK entity

----------------------------------------

Another request is that the newly created gameobjects (the ones I'd like to be renamed) should be inside the CharacterController prefab, not in the Scene (PM me if you don't know how to use prefabs)

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/Attachable/AvatarReference.cs
pravusjif: I just saw that the 2 new GameObjects (for the Avatar and for the Player) are children of the CharacterController GameObject, so maybe we could reference those in a new Monobehaviour in the CharacterController GameObjrect?

BrianAmadori (approved):
It's looking great 👍. I'd add a couple of tests that ensure the special parenting cases keep working as intended--check that the parent is correct, is correctly added as persistent to the bounds checker, etc.

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneBoundariesController/SceneBoundariesChecker.cs
```diff
@@ -91,6 +91,15 @@ public void AddPersistent(DecentralandEntity entity)
             persistentEntities.Add(entity);
         }
 
+        /// <summary>
+        /// Returns whether an entity was added to be concistently checked
```
```suggestion
        /// Returns whether an entity was added to be consistently checked
```

PR #1197, opened 2020-08-20
fix: floating point issues

caminoezequiel:
# What?
Fixes fractional number for `PB_Vector3` in protobuf

# Why?
Setting a fractional number in the Transform can lead to unexpected results


kernel/packages/shared/proto/engineinterface_pb.d.ts
```diff
@@ -1326,6 +1326,12 @@ export class PB_AvatarShape extends jspb.Message {
   getUsedummymodel(): boolean
   setUsedummymodel(value: boolean): void
 
+  getExpressiontriggerid(): string
```
nchamo, line 1329: What are all these changes?

unity-client/Assets/Scripts/MainScripts/DCL/Models/Protocol/Engineinterface.cs
```diff
@@ -3423,16 +3423,16 @@ public sealed partial class PB_Vector3 : pb::IMessage<PB_Vector3> {
           default:
             _unknownFields = pb::UnknownFieldSet.MergeFieldFrom(_unknownFields, input);
             break;
-          case 13: {
-            X = input.ReadFloat();
+          case 9: {
```
nchamo, line 3426: Where do these numbers come from? Is this code autogenerated?

nchamo (commented):
Looks good, just a few questions 😄

unity-client/Assets/Scripts/MainScripts/DCL/Models/Protocol/Engineinterface.cs
BrianAmadori: Yep, this should be autogenerated by the protobuf compiler

BrianAmadori:
Congratulations on your first PR 🙂  

Now, are we sure that this actually fixes the precision issue? How so? Because floating point precision errors are related with using numbers with big integer components, and in that test this seems to be not the case.

Also, the position deviation shown in the tests is very large, so this makes the case even more suspicious.

kernel/packages/shared/proto/engineinterface_pb.js
```diff
@@ -4837,21 +4837,21 @@ proto.engineinterface.PB_Vector3.serializeBinaryToWriter = function(message, wri
   var f = undefined;
   f = message.getX();
   if (f !== 0.0) {
-    writer.writeFloat(
+    writer.writeDouble(
```
BrianAmadori, line 4840: If this change is really necessary, please make sure to modify the `.proto` templates too so the code is auto-generated as intended.

BrianAmadori (changes requested):

kernel/packages/shared/proto/engineinterface_pb.d.ts
caminoezequiel: @nchamo theses methods were autogenerated when I ran [protoc](https://github.com/decentraland/explorer/pull/1197/files/1acb6f1e98f6bb1eb9977d676520df1f56d5623c#diff-b3ea3aa7518ce3b22deb6fcd4967e0dd)

kernel/packages/shared/proto/engineinterface_pb.js
caminoezequiel: @BrianAmadori yes, I did. I updated this file [.proto](https://github.com/decentraland/explorer/pull/1197/files/1acb6f1e98f6bb1eb9977d676520df1f56d5623c#diff-ef819b8b253c3a3525fe2fcf12e43ce2) I don't know if there is another file where I should change it, do you know? 

After changed `.proto` file I ran these [commands](https://github.com/decentraland/explorer/pull/1197/files/1acb6f1e98f6bb1eb9977d676520df1f56d5623c#diff-b3ea3aa7518ce3b22deb6fcd4967e0dd) please let me know if it is ok.

caminoezequiel:
replaced by: https://github.com/decentraland/explorer/pull/1199

PR #1200, opened 2020-08-21
feat: enabling eth_signTypedData_v4

nchamo:
As part of the Matic integration, we need to allow `eth_signTypedData_v4` messages. This is needed to bridge between the Eth and Matic chains.

@HPrivakos has a [library](https://github.com/HPrivakos/decentraland-crypto-utils/) in place that exposes this integration, but in order for it to work, we need to allow `eth_signTypedData_v4` messages.

This is how it is used from the said library:
```
requestManager.provider.sendAsync(
      {
        method: "eth_signTypedData_v4",
        params: [fromAddress, dataToSign],
        jsonrpc: "2.0",
        id: 1,
      } as RPCSendableMessage
)
```


eordano:
would this prompt up metamask, or would it use the ephemeral key? 

If (1); then I think it'd be good to have a UI popping before Metamask does so, explaining that the transaction comes from the owner of the parcel and not from Decentraland.

If (2); then I think the feature is broken :P

HPrivakos:
It comes from Metamask address, like a normal web3 call
Current web3 calls doesn't show any UI before opening metamask

nchamo:
Hey @eordano !

First, as @HPrivakos said, the ephemeral key isn't used for these types of operations, so we should be ok on that side.

Then, regarding the pop-up, I agree that it would be nice to add a a message that explains that the scene is trying to perform a blockchain operation. Now, just to add some context, right now a lot of scenes do these operations without a decentraland pop-up, and only the Metamask one. If we were to add one, we should try to make it consistent.

Also, said pop-up should have the option to "Don't show again on this scene", so we don't show both the decentraland and Metamask pop-up all the time. However, this feature on itself would require a few changes:
1. We realised that we don't have a good way to identify a scene. The entity id changes with each deployment, and the parcels where it was deployed might change (for example a 3x3 scene might be reduced to a 3x2). 
1. Where should we store this preference? On the profile? On the local storage?

I think it would be best to enable the Matic integration, and work on this pop-up or underlaying permission system on another issue. 

Do you think that makes sense?

cc @marcosnc

pablitar (approved):

PR #1204, opened 2020-08-21
release: 2020-08-21

BrianAmadori:
What to test:

- Check out upcoming event coords: 72,15
- Try to teleport as much as possible, particularly try to teleport when there's still loading stuff/green blockers around
- If the places you are teleporting to have avatars, is better


BrianAmadori:
cancelled until next week

PR #1209, opened 2020-08-25
fix: Avatars not visible in preview mode

pablitar:
The global scene was not being initialized while in preview mode. This caused avatars not to be visible while in preview mode.


nchamo (approved):
Great work!

BrianAmadori (approved):

pravusjif (approved):

PR #1213, opened 2020-08-26
Feat: allow SDK move to inside scene

caminoezequiel:
# What? 
Adds method `movePlayerTo` that allows move the player inside of scene

# Why?


kernel/public/test-scenes/0.101.move-inside-scene/scene.json
```diff
@@ -0,0 +1,48 @@
+{
+  "display": {
+    "title": "interactive-text",
+    "favicon": "favicon_asset"
+  },
+  "contact": {
+    "name": "Ezequiel",
+    "email": "ezequiel@decentraland.org"
```
nchamo, line 8: It's not necessary for you to define this. It's ok if you want to leave this like it is, just wanted to mention it :)

kernel/packages/shared/apis/RestrictedActionModule.ts
```diff
@@ -0,0 +1,63 @@
...
+  isPositionValid(position: Vector3) {
+    return this.getSceneData().scene.parcels.some((parcel) => {
+      const { x, y } = parseParcelPosition(parcel)
+      return isInParcel(position, gridToWorld(x, y, new Vector3()))
+    })
+  }
+
+  @exposeMethod
+  async requestMoveTo(newPosition: Vector3, cameraTarget?: Vector3): Promise<void> {
+    // validar que tenga permission
```
nchamo: We would need to translate all these comments into english

kernel/packages/shared/apis/RestrictedActionModule.ts
```diff
@@ -0,0 +1,63 @@
...
+    const basePosition = new Vector3()
+    gridToWorld(base.x, base.y, basePosition)
+
+    return basePosition.add(newPosition)
+  }
+
+  isPositionValid(position: Vector3) {
+    return this.getSceneData().scene.parcels.some((parcel) => {
+      const { x, y } = parseParcelPosition(parcel)
+      return isInParcel(position, gridToWorld(x, y, new Vector3()))
```
nchamo: I think you can avoid passing the `new Vector3()`

kernel/packages/shared/apis/RestrictedActionModule.ts
```diff
@@ -0,0 +1,63 @@
...
+  hasPermission(permission: Permission) {
+    const json = this.getSceneData()
+    const list = json.requiredPermissions || []
+    return list.indexOf(permission) !== -1
+  }
+
+  calculatePosition(newPosition: Vector3) {
+    const base = parseParcelPosition(this.getSceneData().scene.base)
+
+    const basePosition = new Vector3()
```
nchamo, line 34: Could this be changed as
```
    const baseWorld = gridToWorld(base.x, base.y)
    return baseWorld.add(newPosition)
```
?

nchamo (commented):
Looks really good. Just added a few comments. 

As we said, we would need to add a few tests 😄

kernel/public/test-scenes/0.101.move-inside-scene/game.ts
```diff
@@ -0,0 +1,13 @@
+import { BoxShape, engine, Entity, log, moveTo, OnPointerUp, Transform, Vector3 } from 'decentraland-ecs/src'
+
+const e = new Entity()
+e.addComponent(new BoxShape())
+e.addComponent(new Transform({ position: new Vector3(4, 0, 6) }))
+e.addComponent(
+  new OnPointerUp((e) => {
+    log('clicked: ', e)
+    moveTo({ x: 10, y: 0, z: 10 }, { x: 4, y: 1, z: 6 })
```
nearnshaw: How about renaming the function to `movePlayerTo`, instead of `moveTo` ?

I think it would be clearer, otherwise people might wonder if that's to move an entity, not the player

kernel/public/test-scenes/0.101.move-inside-scene/game.ts
```diff
@@ -0,0 +1,13 @@
+import { BoxShape, engine, Entity, log, moveTo, OnPointerUp, Transform, Vector3 } from 'decentraland-ecs/src'
+
+const e = new Entity()
+e.addComponent(new BoxShape())
+e.addComponent(new Transform({ position: new Vector3(4, 0, 6) }))
+e.addComponent(
+  new OnPointerUp((e) => {
```
nchamo: Just tried the scene locally, and it works great!!
Super minor change to make the scene more user friendly:
We could change `OnPointerUp` for `OnPointerDown` and add a hover text that says something like "Move player"
You can see here how to set the hover text https://docs.decentraland.org/development-guide/click-events/

kernel/packages/shared/apis/RestrictedActionModule.ts
```diff
@@ -0,0 +1,63 @@
...
+    // validar que tenga permission
+    if (!this.hasPermission(Permission.ALLOW_MOVE_INSIDE_SCENE)) {
+      defaultLogger.error(`Permission "${Permission.ALLOW_MOVE_INSIDE_SCENE}" is required`)
+      return
+    }
+    // calcular nueva position
+    const position = this.calculatePosition(newPosition)
+    // validate new position is inside of any scene
+    if (!this.isPositionValid(position)) {
+      defaultLogger.error('Error: Position is out of scene')
```
pablitar: We could maybe log the position here
```suggestion
      defaultLogger.error('Error: Position is out of scene', position)
```

kernel/packages/shared/apis/RestrictedActionModule.ts
caminoezequiel: @nchamo no, we don't. If we change it we loss the method `add` because `gridToWorld` returns a `Vector3Component` and it doesn't have the method `add`.

kernel/public/test-scenes/0.101.move-inside-scene/game.ts
caminoezequiel: @nearnshaw yes, of course! that name is better. thanks

kernel/packages/shared/apis/RestrictedActionModule.ts
caminoezequiel: yes, I'll remove it.

pbosio (approved):

kernel/packages/shared/apis/RestrictedActionModule.ts
nchamo: Oh, ok. Closing this then

kernel/packages/shared/apis/RestrictedActionModule.ts
```diff
@@ -0,0 +1,62 @@
+import { exposeMethod, registerAPI } from 'decentraland-rpc/lib/host'
+import { ExposableAPI } from './ExposableAPI'
+import defaultLogger from '../logger'
+import { unityInterface } from '../../unity-interface/UnityInterface'
+import { ParcelIdentity } from './ParcelIdentity'
+import { Vector3 } from '../../decentraland-ecs/src/decentraland/math'
+import { gridToWorld, isInParcel, parseParcelPosition } from '../../atomicHelpers/parcelScenePositions'
+
+enum Permission {
+  ALLOW_MOVE_INSIDE_SCENE = 'ALLOW_MOVE_INSIDE_SCENE'
```
nchamo: I would also change this from `ALLOW_MOVE_INSIDE_SCENE` to `ALLOW_TO_MOVE_PLAYER_INSIDE_SCENE`, to make it consistent with the other renaming

nchamo (commented):
Thanks for the changes!

Just added a really small comment. With that change and tests, we should be ready to go!

kernel/test/unit/RestrictedActionModule.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+      const module = new RestrictedActionModule(options)
+
+      await module.movePlayerTo(new Vector3(21, 0, 32))
+      sinon.verify()
+    })
+
+    it('should fail when scene does not have permissions', async () => {
+      // remove permissions
+      scene.land.sceneJsonData.requiredPermissions = []
```
nchamo: This worries me a little bit.

If this test was executed before the other two, then would those fail?

Maybe we should have a method that takes the list of requiredPermissions (the enum), and returns the scene object?

nchamo (approved):
Great work!!

Just a little comment on the tests, and we would be ready to go

PR #1216, opened 2020-08-26
feat: track interface

2fd:
# What? 
- implements a generic interface to send events to Segment
- tracks when the email prompt appears and when it is skipped by the user

# Why? 
- we need better visibility about how is performing these new features


marcosnc (approved):

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/EmailPromptHUD/EmailPromptHUDController.cs
```diff
@@ -31,6 +31,7 @@ public void SetVisibility(bool visible)
             Utils.UnlockCursor();
             view.gameObject.SetActive(true);
             view.showHideAnimator.Show();
+            WebInterface.ReportAnalyticsEvent("open email popup");
```
BrianAmadori, line 34: Sounds like we should put all these strings as constants under a common static class.

BrianAmadori (approved):
It looks good! 👍 Just left a small detail.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/EmailPromptHUD/EmailPromptHUDController.cs
pbosio: Really? This are like super specific strings... I could understand if you want them as a constant in that class, but I fail to see the porpoise on having a common static class for this 🤔

PR #1220, opened 2020-08-27
fix: animation cull issues

### LINKED ISSUE #1217: Animation culling: issue with animation-based elevators
This is related to the following cases:

- When using the elevator going to the air trolley ride (or whatever that is) if you are not looking down, the elevator would never start.
- If you get near any avatar, and the avatar is on the edge of the screen, you can see it going to pose T.
- If you get near the scarface zeppelin in genesis plaza and face the opposite direction, you can see how the shadow stops.

BrianAmadori:
fixes: https://github.com/decentraland/explorer/issues/1217

- fix: When using the elevator going to the air trolley ride (or whatever that is) if you are not looking down, the elevator would never start.
- fix: If you get near any avatar, and the avatar is on the edge of the screen, you can see it going to pose T.
- fix: If you get near the scarface zeppelin in genesis plaza and face the opposite direction, you can see how the shadow stops.

The fix works by deactivating the animation culling (this means set it to `AlwaysAnimate`) when you get less than 20 meters of distance with relation to the closest point of the animated renderer bounds.

Also, this PR introduces the `CullingController` and `RendererTracker` systems that use the optimized `CullingGroup` Unity class. They will be useful when we implement small object culling for exteriors.

- [X] Basic systems implementation
- [X] Working fix
- [ ] Unit tests

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/RendereableAssetLoadHelper.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/RendereableAssetLoadHelper.cs
@@ -104,7 +104,9 @@ namespace DCL.Components
 
         public void Unload()
         {
-            Environment.i.cullingController.RemoveRenderersFromGameObject(this.loadedAsset);
+            if (this.loadedAsset != null)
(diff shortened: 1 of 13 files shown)


BrianAmadori:
Closing the PR because the approach is not optimal. Will disable the culling and re-implement this fix in a better way later.

PR #1223, opened 2020-08-27
fix: entites out of boundaries material on remove

pbosio:
In some cases, using debug (like builder), if a mesh is removed while out of boundaries it may stay cached with the red "out of boundaries" material instead of it original one.


sandrade-dcl (approved):

BrianAmadori (changes requested):
Cool! We should add test to cover this use case and its good to go.

pbosio:
> Cool! We should add test to cover this use case and its good to go.

I've tried to create a test for that already, but I couldn't reproduce the issue with the test 🤷‍♂️

pravusjif (approved):

sandrade-dcl (approved):

pravusjif (approved):

BrianAmadori (approved):

PR #1225, opened 2020-08-27
fix: (builder) add support for teleport, nft dialog and external links

pbosio:


sandrade-dcl (approved):

pravusjif (approved):

unity-client/Assets/Builder/Scripts/DCLBuilderBridge.cs
```diff
@@ -523,6 +524,11 @@ private void SetPlayMode(bool isPreview)
             cameraController?.gameObject.SetActive(isPreviewMode);
             cursorController?.gameObject.SetActive(isPreviewMode);
 
+            if (!isPreview)
+            {
+                HideHUDs();
```
BrianAmadori, line 529: If its not preview, we don't show the HUDs again?

pbosio: If builder change mode from Preview to Editor (preview being first person / character view) we hide all HUDs...
for example an smart item that prompt a teleport request, we show it in preview mode but need to hide it when it changes to editor mode.
There is something I need to fix related to this (somthing about the external urls hud), but not being 100% related to builder, I will be creating a new PR for that.

BrianAmadori (approved):

unity-client/Assets/Builder/Scripts/DCLBuilderBridge.cs
BrianAmadori: The actual question was more like, why we don't have a `ShowHUDs` or something in case the `isPreview` comes as true? Anyway i gave an approve, but the asymmetric behavior in that method bugged me.

PR #1226, opened 2020-08-27
fix: (builder) disabled canvas resizing logic

pbosio:


sandrade-dcl (approved):

pravusjif (approved):

BrianAmadori (approved):

kernel/packages/unity-interface/UnityInterface.ts
```diff
@@ -90,10 +94,12 @@ export class UnityInterface {
     this.Module = this.gameInstance.Module
     _gameInstance = gameInstance
 
-    if (this.Module) {
-      window.addEventListener('resize', this.resizeCanvasDelayed)
-      this.resizeCanvasDelayed(null)
-      this.waitForFillMouseEventData()
+    if (!EDITOR) {
+      if (this.Module !== undefined) {
```
sandrade-dcl, line 98: I have just noticed that this line will provoke the null exception that we got before fix it on this [**PR**](https://github.com/decentraland/explorer/pull/1214).

We should create another PR and change this line by this one:
`if (this.Module) {`

sandrade-dcl (commented):
The canvas null reference is happening again. We should fix it.

kernel/packages/unity-interface/UnityInterface.ts
pbosio: Thx Santi! I'll do that

PR #1227, opened 2020-08-28
fix: avoid undefined name on SDK getUserData

nchamo:
We are going a step forward into handling profiles consistently, whether they are other users' profiles, or our own.

In this case, we are defaulting to a `Guest-` name, if one is not defined


pablitar (approved):
LGTM!

PR #1228, opened 2020-08-28
fix: materials sharing texture issue

BrianAmadori:
When two material components shared the same texture and any of the materials was disposed, the texture was being disposed too when it shouldn't.

This affected particularly the alcoholic scarecrow scene.

From @nearnshaw:

>In one scenario, try spawning south of the scene, around -12,-34 and walking up. You’ll see that the scarecrow scene is full of white boxes that should be invisible click areas w invisible textures. The same happens to the scene on the West from there, that has a large white wall that should be invisible. Also the Woner World scene to the north has these white cubes inside.

> In another scenario, start spawning north of the scene, around -14,-18, and walk south. (reloading fresh, not teleporting)  You’ll see first that the Wonder World scene is fine, the white cubes disappear when the loading is finished. The scene to the West of the scarecrow now has a large Black wall, not white like before, and in the scarecrow scene some of the cubes are invisible, others are black

This PR includes:

- [x] The fix
- [x] Tests that prove the issue is fixed
- [x] Refactor to material tests so they use our updated `TestHelpers` methods and "should" nomenclature. 
- [x] General code cleanup around the Material components


unity-client/Assets/Scripts/MainScripts/DCL/Components/Textures/DCLTexture.cs
```diff
@@ -1,8 +1,8 @@
 using DCL.Components;
 using DCL.Controllers;
-using DCL.Helpers;
 using System;
 using System.Collections;
+using Cinemachine;
```
sandrade-dcl: Is it needed to import `Cinemachine` here?

sandrade-dcl (approved):

pbosio (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/Textures/DCLTexture.cs
BrianAmadori: Nope, fixed it :)

pravusjif (approved):

PR #1231, opened 2020-08-31
fix: upside down nftshapes

pravusjif:
Re-applied Z rotation to classic NFTShape prefab to fix upside-down picture frames.

You can test that NFTshapes in genesis plaza and in SOHO plaza are all fine.


sandrade-dcl (approved):

pbosio (approved):

BrianAmadori (changes requested):
Can we please add a small test to ensure the rotation is correctly applied and we don't stumble with this regression again?

BrianAmadori (approved):
Tests added but were marked as explicit because they fail on CI for unknown reason. We have to upload visual tests to CI in order to debug this.

PR #1234, opened 2020-08-31
fix: gifs firefox crash

pravusjif:
Added chrome check to use the unity plugin GIF processing on browsers with no 2D context in OffsetCanvas (like firefox)


kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -343,6 +344,13 @@ export class BrowserInterface {
   }
 
   async RequestGIFProcessor(data: { imageSource: string; id: string; isWebGL1: boolean }) {
+    const isChrome = !!window.chrome && (!!window.chrome.webstore || !!window.chrome.runtime)
```
BrianAmadori: It shouldn't be better to use this check? So the gif still works on non-chrome browsers with offscreen canvas down the road? (and if firefox add support it'd fix itself)

```
if (typeof OffscreenCanvas !== "undefined") {     
    isSupported = true; 
} else {     
    isSupported = false; 
}
```

pravusjif: good idea, thx!

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -343,6 +344,14 @@ export class BrowserInterface {
   }
 
   async RequestGIFProcessor(data: { imageSource: string; id: string; isWebGL1: boolean }) {
+    // tslint:disable-next-line
+    const isSupported = typeof OffscreenCanvas !== "undefined"
```
pablitar: Have you tested this in Firefox? Keep in mind that `OffscreenCanvas` seems to be undefined only in worker context, and not in the main thread.

Also, we should test if the OffscreenCanvas has `2d context` support.

pravusjif: I'll add the `typeof OffscreenCanvasRenderingContext2D === "function"` check, thx for pointing it out!

BrianAmadori (approved):

sandrade-dcl (approved):
Code reviewed and fix tested on Firefox: I have loaded the Superrare scene where there are a lot of gif's and the navigator have not crashed. The GIF's have been loaded slower than in Chrome version, just as we expected.

pablitar (approved):
Great work!! :D

PR #1235, opened 2020-08-31
Feat: allows to user save his interests

caminoezequiel:
# What? 
Allows to user save his interests

# Why? 
...


pablitar (approved):
Great work!

PR #1236, opened 2020-08-31
fix: openExternalURL HUD add animator and fixed initial config

pbosio:


sandrade-dcl (approved):

BrianAmadori (approved):
Tested the external link dialog box around genesis plaza and worked like a charm.

PR #1244, opened 2020-09-01
fix: load avatars when exporting a scene

nchamo:
Right now, when using the cli to export a scene, the avatars fail to load and the client crashes.

This is mainly because the default profile is exposed only on `localhost:8000`. This causes the command `dcl start` to work, but it fails when exporting.


pablitar (approved):
LGTM!

BrianAmadori (approved):

PR #1250, opened 2020-09-02
fix: avoid rotating the player if no camera target is set

nchamo:
Previously, we used to rotate the camera, even if no target was set, when teleporting the player.

Now, we keep this behaviour when the player teleports into a scene. But now, when using the method `movePlayerTo`, we maintain the same rotation if no `camera target` is set.


caminoezequiel (approved):

pablitar (approved):
LGTM!

PR #1251, opened 2020-09-02
feat: New Taskbar implementation

sandrade-dcl:
Implementation of the new taskbar.

**Notes:**
- Use `ENABLE_NEW_TASKBAR` param in the url to see this feature.
- You can use `ENABLE_EXPLORE_HUD` param in the url to see the Explore button.


RominaMarchetti (approved):
UX approved :D

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/AvatarEditorHUDView.cs
```diff
@@ -348,6 +350,9 @@ public void SetVisibility(bool visible)
         avatarEditorCanvas.enabled = visible;
         avatarEditorCanvasGroup.blocksRaycasts = visible;
         isOpen = visible;
+
+        if (!isOpen)
+            OnClose?.Invoke();
```
pbosio: Can we check here if it was previously visible? calling `SetVisibility(false)` multiple times will trigger the `OnClose` event every time. It should never happen but just in case...

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ExploreHUD/ExploreHUDController.cs
```diff
@@ -5,7 +5,7 @@
 
 public class ExploreHUDController : IHUD
 {
-    internal ExploreHUDView view;
+    public ExploreHUDView view;
```
pbosio: If we need this `public` can we at least limit the access to set `view`
```suggestion
    public ExploreHUDView view { get; private set; }
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ExploreHUD/ExploreHUDView.cs
```diff
@@ -10,6 +10,8 @@ internal class ExploreHUDView : MonoBehaviour
     [SerializeField] internal Button_OnPointerDown togglePopupButton;
     [SerializeField] internal Color[] friendColors = null;
 
+    public event System.Action OnClose;
```
pbosio: Is it possible to move this to the controller instead of the view? and if that is possible, can we make the class internal again?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/SettingsHUD/SettingsHUDView.cs
```diff
@@ -6,6 +6,8 @@ public class SettingsHUDView : MonoBehaviour
     {
         public bool isOpen { get; private set; }
 
+        public event System.Action OnClose;
```
pbosio: Is it possible to move this event to the controller?

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/TaskbarHUDController.cs
```diff
@@ -133,15 +147,59 @@ private void ChatHeadsGroup_OnHeadOpen(TaskbarButton taskbarBtn)
...
+    private void View_OnBackpackToggleOff()
+    {
+        avatarEditorHud.SetVisibility(false);
+    }
+
+    private void View_OnExploreToggleOn()
+    {
+        exploreHud.SetVisibility(true);
+        OnAnyTaskbarButtonClicked?.Invoke();
+    }
```
pbosio, line 176: It was a requirement for the explore popup to be able to have it on screen while chatting

pbosio (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/AvatarEditorHUDView.cs
sandrade-dcl: You're right, done! Thanks for the advice :)

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ExploreHUD/ExploreHUDController.cs
sandrade-dcl: Ok, we can limit the access through a property. Done! Thanks!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ExploreHUD/ExploreHUDView.cs
sandrade-dcl: In the same way we are doing in other views like `WorldChatWindowHUDView`, `PrivateChatWindowHUDView`, `FriendsHUDView`, etc. the responsibility of notify about the window closing is of the own view and not the controller (anyway we can discuss it with the team in order to do a possible refactor in all classes). Anyway, even moving this event to the controller, we would continue needing to be public because it the class is used in other test classes like `TaskbarHUDShould`, so I cannot convert it to internal :'(

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/SettingsHUD/SettingsHUDView.cs
sandrade-dcl: For now we are following the same resposibility principle than other views (as I have told in the previous comment, we can discuss it).

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ExploreHUD/ExploreHUDView.cs
BrianAmadori: @sandrade-dcl By using the `[assembly: InternalsVisibleTo("<assembly-name>")]` we can make the internals visible to the testing classes, so maybe you can use this to keep the internal class?. Just look for any `AssemblyInfo.cs` file. This is a pattern we currently use quite broadly!.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/HelpAndSupportHUD/HelpAndSupportHUDView.cs
```diff
@@ -0,0 +1,58 @@
...
+        public static HelpAndSupportHUDView Create()
+        {
+            HelpAndSupportHUDView view = Instantiate(Resources.Load<GameObject>(PATH)).GetComponent<HelpAndSupportHUDView>();
+            view.Initialize();
+            return view;
+        }
+
+        public void SetVisibility(bool visible)
+        {
+            gameObject.SetActive(visible);
```
BrianAmadori: We should use the `ShowHideAnimator` here, so we have the fade in/out effect overrideable by custom anims.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/SettingsHUD/SettingsHUDView.cs
BrianAmadori: I think Pato may be onto something here. In the rest of our views, the events are triggered by UI interactions, i.e. `OnClose` would mean the user clicked on the close button. Here, the event is triggered by a state change, and the state owner is the controller.

sandrade-dcl: Ok, I will take a look to it, thanks guys.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ExploreHUD/ExploreHUDView.cs
sandrade-dcl: Ok, I will take a look to it, thanks guys.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/HelpAndSupportHUD/HelpAndSupportHUDView.cs
sandrade-dcl: You're right, I will use it. Thanks!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/TaskbarHUDController.cs
sandrade-dcl: I have added a piece of code to take on account this specific rule. It's working. Thanks Pato!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ExploreHUD/ExploreHUDView.cs
sandrade-dcl: Done! Thanks to the AssemblyInfo, I have been able to keep the class as internal, just as we wanted.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/SettingsHUD/SettingsHUDView.cs
sandrade-dcl: Done!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/HelpAndSupportHUD/HelpAndSupportHUDView.cs
sandrade-dcl: Done! I applied the `ShowHidAnimator` in the `HelpAndSupportHUD` and `SettingsHUD`.

pbosio (approved):

BrianAmadori (approved):
Amazing work! 👍 

Some feedback regarding UX @RominaMarchetti :

1) The grey bar looks too distracting for me. Maybe we can put some alpha or gradient there to ease this feeling? Ok, we can collapse it, but why not tone down a bit so I can have the bar open to see notifications and such without the distraction?. Or at least do that when the mouse is locked.

2) When choosing to hide the UI, there's no feedback or button to how to open again, so I can get stuck here.

3) I think the navigation would be smoother if the popups are closed when clicking outside the window. With the taskbar at bottom I have to move the mouse all way up to do this.

sandrade-dcl:
Hey @BrianAmadori, only for clarify: This PR has not yet been touched by @RominaMarchetti 's magic. All the visible art of the new Taskbar is from me (you know... programmer art) xD

PR #1252, opened 2020-09-02
Chore: add catalyst-client to update profile

caminoezequiel:
# What? 
Adds `catalyst-client` to save the profile, removes duplicated code

### DIFF
--- a/kernel/packages/shared/profiles/sagas.ts
+++ b/kernel/packages/shared/profiles/sagas.ts
@@ -1,6 +1,6 @@
 import { Store } from 'redux'
 import { EntityType } from 'dcl-catalyst-commons'
-import { CatalystClient, DeploymentBuilder, DeploymentData } from 'dcl-catalyst-client'
+import { ContentClient, DeploymentBuilder, DeploymentData } from 'dcl-catalyst-client'
 import { call, fork, put, race, select, take, takeEvery, takeLatest } from 'redux-saga/effects'
 
 import { getFromLocalStorage, saveToLocalStorage } from 'atomicHelpers/localStorage'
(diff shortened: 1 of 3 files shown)


kernel/packages/shared/profiles/sagas.ts
```diff
@@ -697,22 +618,32 @@ export async function modifyAvatar(params: {
...
-    content
+    contentFiles,
+    metadata
   )
-  return deploy(url, data)
+  // sign the entity id fetchMetaContentServer
+  const authChain = Authenticator.signPayload(identity, preparationData.entityId)
+  // Build the client
+  const baseUrl = url.replace('/content', '')
+  const catalyst = new CatalystClient(baseUrl, 'explorer-profile')
```
nchamo: You can use a `ContentClient` instead of a `CatalystClient`, and avoid removing the `/content`

kernel/packages/shared/profiles/sagas.ts
```diff
@@ -61,26 +63,9 @@ import {
...
 import { ensureServerFormat } from './transformations/profileToServerFormat'
-import {
-  Catalog,
-  Profile,
-  WearableId,
-  Wearable,
-  Collection,
-  Entity,
-  EntityField,
-  ControllerEntity,
```
nchamo, line 72: I think we can also delete plenty of these types form the `./types` file

nchamo (commented):
Looks great!

Love removing all that code

nchamo (approved):

kernel/packages/shared/profiles/sagas.ts
caminoezequiel: We are using that types, do you think we should use from another definition?

PR #1253, opened 2020-09-02
fix: badly optimized animations related crash

BrianAmadori:
Addresses #1219 

In SoYou scene we had a fairly consistent crash that corresponded with the use of many badly optimized animations that took around 400mb, getting our memory limit out of whack. 

The fix applied should mitigate this issue optimizing the keyframes for all animations being loaded. This is done by pruning all consecutive curve values with same-ish values.

https://play.decentraland.zone/branch/fix/big-gltf-crash/index.html?ENV=org&position=25%2C-126

### Before
A bunch of animations like this. In the SoYou case, we have 120 animations with this data style:
image

A complex one:
image

### After
image

The same complex one:
image

Regression tests 
- [x] Genesis plaza: baloon & sky train
- [x] Dinosaur Hunt:   -10,-37
- [x] Pixels for madagascar: -35,37


unity-client/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -1111,23 +1111,50 @@ void ProcessCurves(Transform root, GameObject[] nodes, AnimationClip clip, GLTFA
...
+            {
+                float valueDelta = Mathf.Abs(rawKeyframes[i].value - lastKf.value);
+                float deviation = valueDelta;
+
+                if (deviation > 0.001f)
+                {
+                    result.Add(rawKeyframes[i]);
+                }
+
+                lastKf = rawKeyframes[i];
```
menduz: Why isn't this part of the previous IF?

BrianAmadori: Hi @menduz !

This is done in order to don't mess with the original data and add the correct ending keyframe on the `if (result.Count == 1)` section of this method. As keyframes have a `time` field, if I don't do that we'd end with `time == 0` for the keyframes in both sides of the curve. 

In any case I am still working on some tests that should confirm there isn't something that break the curves with this logic, because we could end with far away keyframes that have differing tangents and don't match the original animation closely. This is supposed to be a very lazy optimization, not a curve matching keyframe reduction that would require far more involved code.

sandrade-dcl (approved):
Awesome optimization!

PR #1268, opened 2020-09-07
feat: world blockers

pravusjif:
This PR implements **_world blockers_** around the world-explorable area instead of scene-blockers


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WorldBlockersController.cs
```diff
@@ -0,0 +1,46 @@
+﻿using UnityEngine;
+
+namespace DCL.Controllers
+{
+    public class WorldBlockersController : MonoBehaviour
```
pbosio: just out of curiosity, do we need this to be a `MonoBehaviour`? can't we implemented so it doesn't?

pbosio (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WorldBlockersController.cs
pravusjif: We need a gameobject to be the parent of the blockers in-world and be repositioned when the world repositions... we could avoid having this class as a Monobehaviuor but that would require to move some logic into an existent one, like the SceneController, but we are refactoring it to take away all these kind of responsibilities right? So I'm not sure where else could we put the logic that deals with the parent of the blockers

pbosio: Ok, didn't see the reposition there... makes sense then 👍

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -41,7 +41,7 @@ public enum State
         public ContentProvider contentProvider;
         public int disposableNotReadyCount => disposableNotReady.Count;
 
-        [System.NonSerialized] public bool useBlockers = true;
+        [System.NonSerialized] public bool useBlockers = false;
```
sandrade-dcl: Where is this param being used?

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
pravusjif: for example here: https://github.com/decentraland/explorer/blob/d9dad36411b16b238f2928f1d23d06d76b2b8fb5/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs#L143

That bool is to instantiate (or not) the scene blockers, since this PR puts blockers around the explorable area of the world (scenes that didn't finish loading don't count as explorable area), the scene blockers are not necessary anymore

sandrade-dcl: Oh, ok! I didn't see it in ParcelScene. Make sense. Thanks!

unity-client/Assets/Shaders/MaterialTransition/Unlit Hologram.shader
```diff
@@ -77,6 +79,19 @@ Shader "DCL/FX/Hologram"
...
+                if(_MaxRenderingDistance > 0)
+                {
+                    float renderingDistance = length(_WorldSpaceCameraPos.xyz - i.posWorld.xyz);
+
+                    if (renderingDistance > _MaxRenderingDistance) return (0,0,0,0);
+
+                    float halfMaxRenderingDistance = _MaxRenderingDistance / 2;
+                    if (renderingDistance > halfMaxRenderingDistance)
+                        distanceAlphaMultiplier = 1 - ((renderingDistance - halfMaxRenderingDistance) / halfMaxRenderingDistance);
+                }
```
BrianAmadori: Why are we using the `_MaxRenderingDistance` if we just halve it later? I think its a bit better to let the frustum culling do the far away culling, and just do the `distanceAlphaMultiplier` bit with a `saturate` to clamp it to 0--1. This way we can save a couple of branching instructions that are bad for shader performance.

unity-client/Assets/Shaders/MaterialTransition/Unlit Hologram.shader
```diff
@@ -85,15 +100,16 @@ Shader "DCL/FX/Hologram"
 
                 float finalRimFactor = rim * animatedThrobbing;
                 float3 finalRimColor = _RimColor.rgb * (lambertAtten * finalRimFactor);
-				
+
                 float4 finalColor = float4( _Color.rgb + finalRimColor, _Color.a + finalRimFactor );
+                finalColor.a = clamp(finalColor.a * distanceAlphaMultiplier, 0, 1);
```
BrianAmadori:
```suggestion
                finalColor.a = saturate(finalColor.a * distanceAlphaMultiplier);
```

BrianAmadori (changes requested):

unity-client/Assets/Shaders/MaterialTransition/Unlit Hologram.shader
pravusjif: The current logic does the follwing:
* has a max rendering distance for the hologram
* In the farer half-distance, the alpha multiplier is lerped/applied
* In the nearest half-distance the alpha multiplier is always 1

`i.pos.z` didn't work out-of-the-box for me but I can try again, and also I would prefer to keep this separated from the frustrum culling in case we some day change the frustrum culling, but if you insist I can try changing that to use what you mentioned

BrianAmadori: I just meant that we can save the branching that's a bit slow on older GPUs. The distance logic is fine!

I suggest that we remove all the if statements with the exception of the early exit one:
```
                    float renderingDistance = length(_WorldSpaceCameraPos.xyz - i.posWorld.xyz);

                    if (renderingDistance > _MaxRenderingDistance) return (0,0,0,0);

                    float halfMaxRenderingDistance = _MaxRenderingDistance / 2;
                    distanceAlphaMultiplier = 1 - ((renderingDistance - halfMaxRenderingDistance) / halfMaxRenderingDistance);
```

This said, having the `halfMaxRenderingDistance` as the fade distance threshold irks me a bit, but its fine for our purposes.

pravusjif: Oh I understand now, great, yes, I'll make that change

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WorldBlockersController.cs
BrianAmadori: If that is the only reason, It isn't better to just make a `new GameObject` on the constructor and use that as the movable parent? This way we can mock tests by just doing `new WorldBlockersController()` and save the Unity specific overhead.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/WorldBlockerHandler.cs
```diff
@@ -0,0 +1,98 @@
+﻿using DCL.Configuration;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace DCL.Controllers
+{
+    public class WorldBlockerHandler : BlockerHandler
```
BrianAmadori: Why do we need the `WorldBlockerHandler` and `BlockerHandler`? Why don't use just a single class?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/WorldBlockerHandler.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        HashSet<Vector2Int> allLoadedParcelCoords = new HashSet<Vector2Int>();
+        HashSet<Vector2Int> blockersToRemove = new HashSet<Vector2Int>();
+        HashSet<Vector2Int> blockersToAdd = new HashSet<Vector2Int>();
+
+        static float milisecondsUsed = 0f;
```
BrianAmadori: I think this is not used anywhere.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/WorldBlockerHandler.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public void SetupGlobalBlockers(Dictionary<string, ParcelScene> loadedScenes, float height, Transform parent)
```
BrianAmadori: I think we can decouple this system from the `ParcelScene` class by just passing the parcels of the ready scenes from the caller, as I only we only are using `isReady` and `parcels` from it.

unity-client/Assets/Shaders/MaterialTransition/Unlit Hologram.shader
pravusjif: done

unity-client/Assets/Shaders/MaterialTransition/Unlit Hologram.shader
pravusjif: Done

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/WorldBlockerHandler.cs
pravusjif: Done

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/WorldBlockerHandler.cs
pravusjif: Done

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WorldBlockersController.cs
pravusjif: OK I'll do that and I'll add the WorldBlockersController initialization in the SceneControlle then.

pravusjif: Done, now the WorldBlockersConttoller is created with dependency injection (and the parent GO created on initialization).

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/WorldBlockerHandler.cs
pravusjif: Done, removed old blockers code and transformed the BlockerHandler with the new world blockers logic.

sandrade-dcl (changes requested):
The solution to the **unity-tests** timeout on CI should not be on increase the `no_output_timeout` to 30min. Between the commits `4eba0da` and `133a9b5` there have been any change that have caused the tests to take much longer (about 2 minutes more than before), so I think we should find the real problem before merge this PR to `master`.

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Tests/BlockerHandlerShould.cs
```diff
@@ -5,42 +5,167 @@
...
+            blockersHandler = Reflection_GetField<BlockerHandler>(worldBlockersController, "blockerHandler");
+            blockers = Reflection_GetField<Dictionary<Vector2Int, PoolableObject>>(blockersHandler, "blockers");
+
+            if (!PoolManager.i.ContainsPool(PARCEL_BLOCKER_POOL_NAME))
+            {
+                GameObject go = Object.Instantiate(Reflection_GetStaticField<GameObject>(typeof(BlockerHandler), "blockerPrefab"));
+                Pool pool = PoolManager.i.AddPool(PARCEL_BLOCKER_POOL_NAME, go);
+                pool.persistent = true;
+                pool.ForcePrewarm();
+            }
```
BrianAmadori: The tests of this suite are setting up everything using the SDK from the ground up when we should just try to test only the `BlockerHandler` and `BlockerHandlerController` classes. I think this is very related to the stuff we were discussing today. If we do this correctly I think we can manage to write tests that are a LOT shorter, and less vulnerable to CI false negatives.

BrianAmadori (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BlockerController/Tests/BlockerHandlerShould.cs
```diff
@@ -0,0 +1,162 @@
...
+            Assert.IsTrue(blockers.ContainsKey(new Vector2Int(-2, 2)));
+        }
+
+        [Test]
+        public void ClearOnlyChangedBlockers()
+        {
+            blockerHandler.SetupGlobalBlockers(sceneHandler.GetAllLoadedScenesCoords(), 10, blockersParent.transform);
+            var blockers = blockerHandler.GetBlockers();
+
+            // Save instante id of some blockers that shouldn't change on the next scene load
```
BrianAmadori:
```suggestion
            // Save instance id of some blockers that shouldn't change on the next scene load
```

BrianAmadori (approved):
Amazing 🚀 
Approving this without smoke-testing it.

PR #1270, opened 2020-09-07
fix: enable web3 when exporting scenes

nchamo:
During the last changes to the initial loading screen, the export page was not updated. Therefore, when exporting scenes, it wasn't possible to enable web3.

In order for it to work, we copied the current `preview.html`, and made a few little changes for it to be compatible with exporting scenes


pablitar (approved):
I would say I love this, but I would be lying.

PR #1271, opened 2020-09-07
feat: changes to attachable based on feedback

nchamo:
We received feedback from the community, regarding the new `Attachable` feature. Based on that feedback, we will be making the following changes:
1. We will rename `Attachable.PLAYER` to `Attachable.FIRST_PERSON_CAMERA`
1. We will rename `Attachable.AVATAR_POSITION` to `Attachable.AVATAR`
1. We will move the root point of the `FIRST_PERSON_CAMERA` entity reference, to the position of the first person camera. By doing this, the pivot point will match the camera

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
@@ -69,10 +69,10 @@ public class DCLCharacterController : MonoBehaviour
 
 
     // Will allow the game objects to be set, and create the DecentralandEntity manually during the Awake 
-    [HideInInspector] public DCL.Models.DecentralandEntity avatarPositionReference;
-    [HideInInspector] public DCL.Models.DecentralandEntity playerReference;
-    [SerializeField] private GameObject avatarPositionGameObject;
-    [SerializeField] private GameObject playerGameObject;
+    [HideInInspector] public DCL.Models.DecentralandEntity avatarReference;
+    [HideInInspector] public DCL.Models.DecentralandEntity firstPersonCameraReference;
+    [SerializeField] private GameObject avatarGameObject;
+    [SerializeField] private GameObject firstPersonCameraGameObject;
 
     [SerializeField] private InputAction_Measurable characterYAxis;
     [SerializeField] private InputAction_Measurable characterXAxis;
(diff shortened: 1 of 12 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -469,21 +469,21 @@ public void SetEntityParent(string entityId, string parentId)
 
             if (me != null)
             {
-                if (parentId == "PlayerEntityReference")
+                if (parentId == "FirstPersonCameraEntityReference")
                 {
-                    me.SetParent(DCLCharacterController.i.playerReference);
+                    me.SetParent(DCLCharacterController.i.firstPersonCameraReference);
```
BrianAmadori, line 476: I remember that we defined that the `Player` would represent the Avatar in third person but the camera in first person. This applies here as well? So attaching something to `FirstPersonCameraEntityReference` would attach to the camera in FPS mode but the avatar in TPS?

nchamo: Hey Brian!

Yes, the functionality is still the same. When you do `entity.setParent(Attachable.FIRST_PERSON_CAMERA)`, then the entity will be attached to the avatar, and rotate with the camera.

If the user then switches to third person camera, it will still be attached to the avatar. It was just a name change, because the community found `Player` vs `Avatar Position` confusing

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
```diff
@@ -69,10 +69,10 @@ public class DCLCharacterController : MonoBehaviour
 
 
     // Will allow the game objects to be set, and create the DecentralandEntity manually during the Awake 
-    [HideInInspector] public DCL.Models.DecentralandEntity avatarPositionReference;
-    [HideInInspector] public DCL.Models.DecentralandEntity playerReference;
-    [SerializeField] private GameObject avatarPositionGameObject;
-    [SerializeField] private GameObject playerGameObject;
+    [HideInInspector] public DCL.Models.DecentralandEntity avatarReference;
+    [HideInInspector] public DCL.Models.DecentralandEntity firstPersonCameraReference;
```
pbosio: Can we change these so can't be assigned publicly?
```suggestion
    public DCL.Models.DecentralandEntity avatarReference { get; private set; }
    public DCL.Models.DecentralandEntity firstPersonCameraReference  { get; private set; }
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -469,21 +469,21 @@ public void SetEntityParent(string entityId, string parentId)
...
                     SceneController.i.physicsSyncController.MarkDirty();
                 }
-                else if (parentId == "AvatarPositionEntityReference")
+                else if (parentId == "AvatarEntityReference")
                 {
-                    me.SetParent(DCLCharacterController.i.avatarPositionReference);
+                    me.SetParent(DCLCharacterController.i.avatarReference);
                     SceneController.i.boundariesChecker.AddPersistent(me);
                     SceneController.i.physicsSyncController.MarkDirty();
                 }
```
pbosio: Can we add here a couple of little comments about what is the supposed behavior when setting as parent any of those entities? So anybody can understand without needing to navigate through the code

pbosio (approved):

PR #1272, opened 2020-09-07
chore: AvatarHUD as legacy

pbosio:
set current AvatarHUD as legacy to start working in the new version


sandrade-dcl (approved):
I understand that this change is only for

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarHUD/Legacy/AvatarHUDController.cs
```diff
@@ -0,0 +1,95 @@
+namespace Legacy
+{
+    public class AvatarHUDController : IHUD
```
BrianAmadori, line 3: Maybe we can take advantage of this opportunity to change the name to something less confusing, like `ProfileHud`, `IdentityToastHud` or whatever (not sure if this is in the scope of this PR, just general feedback).

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarHUD/Legacy/AvatarHUDView.cs
```diff
@@ -0,0 +1,105 @@
+﻿using System.Runtime.CompilerServices;
+using TMPro;
+using UnityEngine;
+using UnityEngine.UI;
+using DCL.Interface;
+using Image = UnityEngine.UI.Image;
+
+[assembly: InternalsVisibleTo("AvatarHUDTests")]
```
BrianAmadori, line 8: We should move this to `AssemblyInfo.cs` (again, general feedback).

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarHUD/Legacy/AvatarHUDController.cs
pbosio: You are right, I will use the suggested names for the new HUD

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarHUD/Legacy/AvatarHUDView.cs
pbosio: All that code will end in the trash after the new HUDs go live, so I don't really see the point of creating a new file to delete it later.
but I will do what you said for the new implementation

PR #1277, opened 2020-09-08
feat: Global voice chat Experimental Feature

pablitar:
This PR adds functionality for a global voice chat implementation that uses the existing P2P network to send audio recorded in the browser, and then encoded using an Opus Codec. On the receiving end, each audio stream is decoded and played with spatial audio parameters for an immersive experience.

### To test this PR:
- Ensure you have the `VOICE_CHAT_ENABLED` parameter added in the URL. For example: https://play.decentraland.zone/branch/feat/global-voice-chat-poc/index.html?VOICE_CHAT_ENABLED&ENV=org
- Press T to talk. Voice chat is implemented as a "push to talk" feature so you'll need to keep T pressed for as long as you are talking.

### Review hints
The PR is long. Sorry about that :sweat_smile:. Some files like those related to libopusjs and the resampler are imported from external sources. They were no NPM packages. But you can think of them as external dependencies.

### What's Next (For future PRs)
- Usability and UI need to be reworked. We may want to add a recording toggle feature to work alongside the push to talk feature. 
- Mute support and voice chat volume control. Currently you can mute a person by blocking them
- More audio settings like device selection and all that
- Voice may not be as clear as it needs if network is not relatively stable or there is congestion. We may have to implement QoS protocol in the peer network
- We may want to implement a dynamic buffer that adjust to network jitter
- Currently, packets that arrive late (after a newer packet) are discarded. We may want to process them anyway and take advantage of the buffer to reorder the packages to improve reliability
- Spatial parameters need to be tuned. Currently, you can hear people that are quite far. We may want more attenuation or an ad-hoc mechanism to ignore audio from a certain distance. This distance may be dynamic, and depend on your friend list or other variables.


unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarName.cs
```diff
@@ -32,6 +33,10 @@ public void SetName(string name)
         }
     }
 
+    public void SetTalking(bool talking) {
```
AjimenezDCL: Style!

Braces goes on next line =D

pravusjif: Just in case we should have a 
```
    void OnDisable()
    {
        SetTalking(false);
    }
```

Since the avatars are pooled now, we may have the case of an avatar returning to the pool with the talkingIconContainer turned ON, and when recycling from the pool, the avatar may have it turned ON, so let's turn it off every time the avatar name is disabled (this happens for example when the avatar is disabled to return to the pool)

pravusjif (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarName.cs
pablitar: I don't think we need that, since SetTalking is not triggered once. As the name and other properties, they are updated constantly since these properties are synced with the Avatar Shape representation in kernel. See here: https://github.com/decentraland/explorer/pull/1277/files/50ee70993ad10a563f266e9f3f5debb4d0fbcab5#diff-3f1d47987ad32229bc1c30015b196428R105

But I'll add it nonetheless. There may be a small window of inconsistency.

kernel/packages/shared/comms/index.ts
```diff
@@ -198,6 +206,60 @@ function getParcelSceneSubscriptions(): string[] {
   return ids
 }
 
+let audioRequestPending = false
+
+export function updateVoiceRecordingStatus(recording: boolean) {
+  if (recording && voiceCommunicator) {
```
BrianAmadori: Maybe we could use a few early exits here to avoid nesting.

kernel/packages/voice-chat-codec/worker.ts
```diff
@@ -0,0 +1,157 @@
...
+      setTimeout(doWork, 0)
+    } else {
+      worklet.working = false
+    }
+  }
+
+  setTimeout(doWork, 0)
+}
+
+//Encoder(channels, samplerate, bitrate, frame_size, voice_optimization)
```
BrianAmadori: Maybe we can remove this comment or at least explain what is this supposed to help with.

kernel/packages/voice-chat-codec/worker.ts
```diff
@@ -0,0 +1,157 @@
...
+        e.data.streamId,
+        decoderWorklet,
+        (worklet) => worklet.decoder.output(),
+        (output, streamId) => ({
+          topic: ResponseTopic.DECODE,
+          streamId,
+          samples: toFloat32Samples(output)
+        })
+      )
+    }
```
BrianAmadori: It wouldn't hurt to extract the `encoder` and `decoder` bits into separate functions, so the code becomes a bit more readable.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarName.cs
```diff
@@ -32,6 +33,16 @@ public void SetName(string name)
...
     }
 
+    void OnDisable()
+    {
+        SetTalking(false);
+    }
+
+    public void SetTalking(bool talking)
+    {
+        talkingIconContainer.SetActive(talking);
```
BrianAmadori: As this is called from `OnDisable` that can be called in weird contexts (i.e. when the object is being destroyed), I'd add a null check just in case.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/VoiceChat/DCLVoiceChatController.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace DCL
+{
+    public class DCLVoiceChatController : MonoBehaviour
```
BrianAmadori, line 7: I think we can safely use a standard C# class here and initialize it from `SceneController.Awake()` .

pablitar: But wouldn't that make necessary to have the `voiceChatAction` as a reference in `SceneController`? Wouldn't that go against splitting `SceneController` into more cohesive chunks?

kernel/packages/shared/comms/index.ts
pablitar: Done!

kernel/packages/voice-chat-codec/worker.ts
pablitar: Done!

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarName.cs
pablitar: Done!

pravusjif (approved):

Issue #1241, opened 2020-09-01
Avatar head stopped hiding for some wearables


pravusjif:
**Ideally this should be fixed before the week of September the 7th, there's a wearables creator contest**

for example the skeleton pirate wearable collectible should turn off the avatar head and it's not happening.

![Screen Shot 2020-08-31 at 10.06.11.png](https://images.zenhubusercontent.com/5d9200f4f58e25000108abac/3fcfc721-d8df-4ade-9dfb-8f055810588c)

To see all wearables you should use **ZONE** environment and 
`ALL_WEARABLES&TEST_WEARABLES` URL PARAMS

PR #1280, opened 2020-09-09
fix: hidden list for BodyShapeController is not being refreshed properly.

AjimenezDCL:
#1241 

Hidden matrix was not working for bodyshapes

-----------------
Test instructions:
To see all wearables you should use `ZONE` environment and `ALL_WEARABLES&TEST_WEARABLES` URL PARAMS, for example the pirate skeleton head was one of the collectible wearables that wasn't hiding the avatar head correctly:

https://explorer.decentraland.zone/branch/fix/1241_head_not_hidden/index.html?ALL_WEARABLES&TEST_WEARABLES

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/BodyShapeController.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/BodyShapeController.cs
@@ -12,8 +12,15 @@ public class BodyShapeController : WearableController
     {
     }
 
-    protected BodyShapeController(WearableController original) : base(original)
+    protected BodyShapeController(BodyShapeController original) : base(original)
     {
+        headRenderer = original.headRenderer;
+        eyebrowsRenderer = original.eyebrowsRenderer;
+        eyesRenderer = original.eyesRenderer;
+        mouthRenderer = original.mouthRenderer;
+        feetRenderer = original.feetRenderer;
+        upperBodyRenderer = original.upperBodyRenderer;
+        lowerBodyRenderer = original.lowerBodyRenderer;
     }
 
     public SkinnedMeshRenderer skinnedMeshRenderer { get; private set; }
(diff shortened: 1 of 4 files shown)


pravusjif (approved):
Bug fixed, welcome back Alex!

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/BodyShapeController.cs
```diff
@@ -140,16 +147,34 @@ protected override void PrepareWearable(GameObject assetContainer)
...
             else if (parentName.Contains("feet"))
                 feetRenderer = r;
+            else if (parentName.Contains("head"))
+                headRenderer = r;
+            else if (parentName.Contains("eyebrows"))
+                eyebrowsRenderer = r;
+            else if (parentName.Contains("eyes"))
+                eyesRenderer = r;
+            else if (parentName.Contains("mouth"))
+                mouthRenderer = r;
```
BrianAmadori, line 157: We shouldn't be using `WearableLiterals` here?

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/BodyShapeController.cs
```diff
@@ -140,16 +147,34 @@ protected override void PrepareWearable(GameObject assetContainer)
...
+    public SkinnedMeshRenderer eyesRenderer { get; private set; }
+    public SkinnedMeshRenderer mouthRenderer { get; private set; }
     public SkinnedMeshRenderer feetRenderer { get; private set; }
     public SkinnedMeshRenderer upperBodyRenderer { get; private set; }
     public SkinnedMeshRenderer lowerBodyRenderer { get; private set; }
 
     public override void UpdateVisibility()
     {
-        SetAssetRenderersEnabled(!hiddenList.Contains(WearableLiterals.Misc.HEAD));
+        bool headisVisible = !hiddenList.Contains(WearableLiterals.Misc.HEAD);
```
BrianAmadori:
```suggestion
        bool headIsVisible = !hiddenList.Contains(WearableLiterals.Misc.HEAD);
```

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/BodyShapeController.cs
BrianAmadori: On a second thought, maybe not because we are handling renderers here, not the wearable model. And the `ubody` `lbody` don't match the literals. Maybe we can use some constants.

AjimenezDCL: We could and it would work, but I dont know... 

In `WearableLiterals` we have well defined literals like Categories, BodyShapeIds, Rarities and such. In this case we are just (in a really ugly way xD) checking the renderer's parent name to now what part of the bodyshape it belongs to. It's just a sort of convention we arranged with the artists that is not actually defined anywhere else like the rest of literals.

We could benefit for having consts but maybe not in `WearableLiterals` to avoid confusion? I dont have strong feelings against the proposed solution though

BrianAmadori: >We could benefit for having consts but maybe not in WearableLiterals to avoid confusion? I dont have strong feelings against the proposed solution though

Maybe this could work, but it'd only help if we need to make this ugly check somewhere else.

PR #1282, opened 2020-09-10
chore: decouple navmap from minimap

AjimenezDCL:
During the onboarding I came across some static function in `MinimapHUDView` to show the `Navmap` and also to update the current scene. This couples together those systems when it's not needed.

Luckily the Navmap was using the good ol' Input system so trigerring the action in the Minimap without coupling the Navmap was easy. Now Navmap reacts to player coords changes directly without having to piggyback into Minimap events.

Tests has been updated accordingly


pbosio (approved):

BrianAmadori (approved):
Didn't test it thoroughly but it looks good codewise!

PR #1288, opened 2020-09-10
feat: add animations to explore

pbosio:
Animation states were added to the UI.
Test link: https://play.decentraland.zone/branch/feat/exploreanimations/index.html?ENV=org&ENABLE_EXPLORE_HUD


sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ExploreHUD/Scripts/Handlers/ThumbnailHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+        if (texture)
+        {
+            GameObject.Destroy(texture);
+        }
+
+        if (texturePromise != null)
+        {
+            AssetPromiseKeeper_Texture.i.Forget(texturePromise);
+            texturePromise = null;
+        }
```
pravusjif: we should do either one of these 2 things, not both. In the current state of this code we could be destroying a texture that is still referenced through the promise.

So I'd suggest checking for the texture promise existence first, and doing the `Forget` in that case, and if the promise is null, then in that case we can destroy the texture in this script

pravusjif (changes requested):
Works fine, just found a potential problem when disposing the thumbnail texture

pravusjif (approved):

PR #1291, opened 2020-09-11
fix: passport buttons are unresponsive

AjimenezDCL:
Chat window was catching mouse buttons from the bottom side of the passport card and therefore hiding the card.


sandrade-dcl (approved):
Tested! It is working again :)

pbosio (approved):

PR #1292, opened 2020-09-11
test: static workers termination

pravusjif:
**THIS IS A HACKY APPROACH TO SEE HOW MUCH PERFORMANCE IS AFFECTED IF WE TERMINATE STATIC WORKERS**

* flagging scenes as "static" based on their minified code usage of `ISystem` and `OnPointer`, as we checked that detects at least empty parcels and street parcels.
* termination of webworkers for static scenes, as they are accumulated as idle workers otherwise
* feature toggle URL param `STATIC_WORKERS_TERMINATION`

Doc with profiling info here: https://docs.google.com/document/d/1alngddAcBxyOHQQ3ieuEGv2H7Y1GXR1LLxje05406sY/edit

Test link: https://play.decentraland.zone/branch/test/StaticSceneWorkersTermination/index.html?ENV=org&STATIC_WORKERS_TERMINATION&position=-72%2C72

You can do your own comparison by using (or not) the `STATIC_WORKERS_TERMINATION` URL param

### DIFF
--- a/kernel/packages/scene-system/scene.system.ts
+++ b/kernel/packages/scene-system/scene.system.ts
@@ -111,6 +111,7 @@ export default class GamekitScene extends Script {
 
   scenePosition: Vector2 = new Vector2()
   parcels: Array<{ x: number; y: number }> = []
+  allowedFirstStaticSceneWorker = false
 
   private allowOpenExternalUrl: boolean = false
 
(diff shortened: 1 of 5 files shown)


kernel/packages/scene-system/scene.system.ts
```diff
@@ -516,6 +518,9 @@ export default class GamekitScene extends Script {
       }
 
       try {
+        sceneIsStatic = !(source.indexOf('ISystem') !== -1 || source.indexOf('OnPointer') !== -1)
```
pablitar, line 521: This is super hacky hahaha.

I know this is a test, but if we decide to try and detect static scenes, let's see if we can do better. I'm not sure this is a reliable test to ensure a scene is static or not.

Issue #852, opened 2020-04-28
Fix shadow artifact in conference center


BrianAmadori:
The best way to fix this is: 

- Add a shadow distance setting to quality settings 
- Increase the default value for the default preset.

This is because this artifact is fixed by increasing shadow distance, BUT in doing this we may diminish the overall shadow quality. This fixes the issue and adds the option to preserve the shadow quality if desired.

![Annotation 2020-04-28 121505.jpg](https://images.zenhubusercontent.com/5d9756603fe18d00011cfcac/a5515bcb-1982-46a7-a3f6-c3e9132ea208)

PR #1296, opened 2020-09-14
Shadow distance menu in settings

AjimenezDCL:
#852 

A new slider to tweak the shadow distance has been added. Some settings have been redistributed to make room for it.

image


sandrade-dcl (changes requested):
Code reviewed and it's fine! But testing the feature in the build I have noticed that we have not associated predefined "Shadow Distance" values for each "Graphics Quality" preset (as we do with other settings, for example with "Shadow Resolution"). I think we should to add it.

AjimenezDCL:
> Code reviewed and it's fine! But testing the feature in the build I have noticed that we have not associated predefined "Shadow Distance" values for each "Graphics Quality" preset (as we do with other settings, for example with "Shadow Resolution"). I think we should to add it.

That is strange, I swear I changed it. Maybe I forgot to commit it with the rest. Thanks!!

pravusjif (approved):

PR #1300, opened 2020-09-14
feat: clipboard access

pbosio:
* enable access to store text in the clipboard
* read text from the clipboard (firefox not supported)
* fix copy & paste in webgl build


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/Implementation/Web/Plugins/ClipboardWebGL.jslib
```diff
@@ -0,0 +1,41 @@
+var ClipboardWebGL = {
```
BrianAmadori: As we already know, implementing plugins like this is a bit clunky because you have to:

- Compile to see results
- Forced to use an outdated JS version.

I think we can safely move this code to kernel and save a couple of future headaches, like we did with the native messages and GIF code.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -75,6 +75,9 @@ void Awake()
 
 #if !UNITY_EDITOR
             worldEntryPoint = new EntryPoint_World(this); // Es un subsistema independiente => Se pone en el entrypoint pero no en el enviroment.
```
BrianAmadori: Please can you do me a favor and remove/translate this comment? 🤣

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/Implementation/Web/ClipboardWebGL.cs
```diff
@@ -0,0 +1,83 @@
+using System;
+using System.Runtime.InteropServices;
+using AOT;
+using UnityEngine;
+
+internal class ClipboardWebGL : IClipboardImplementation, IDisposable
+{
+    private static ClipboardWebGL i;
```
BrianAmadori: If we are 100% sure we need a singleton, we should use `Singleton<T>`. 

In this case, however, creating a new instance from `SceneController` could make the class more test-friendly/less coupled.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/Implementation/Web/ClipboardWebGL.cs
```diff
@@ -0,0 +1,83 @@
...
+
+    private Action<string, bool> onRead;
```
BrianAmadori:
```suggestion
    private Action<string, bool> OnRead;
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/ClipboardReadPromise.cs
```diff
@@ -0,0 +1,65 @@
+using System;
+using UnityEngine;
+
+public class ClipboardReadPromise : CustomYieldInstruction
```
BrianAmadori: Name and scope of this class doesn't seem appropiate. I propose a rename and to move it to `CoroutineHelpers`, `Utils` or such.
```suggestion
public class Promise : CustomYieldInstruction
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/Implementation/IClipboardImplementation.cs
```diff
@@ -0,0 +1,8 @@
+using System;
+
+internal interface IClipboardImplementation
```
BrianAmadori: Its hard to me to see the word `Implementation` in an interface. The `handler` name maybe its more appropiate because it's what we were using for similar interfaces?
```suggestion
internal interface IClipboardHandler
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/Clipboard.cs
```diff
@@ -0,0 +1,50 @@
+using System;
+using System.Collections.Generic;
+
+public static class Clipboard
```
BrianAmadori: I think we can add some test coverage to at least cover if the `ClipboardReadPromise` queue logic works as intended.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/Clipboard.cs
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+#if UNITY_WEBGL && !UNITY_EDITOR
+    public static void HookBrowserCopyPasteInput()
+    {
+        // NOTE: this does nothing but we'll use it to force the instantiation of this static class
+        // cause you need to call something for a static class to be instantiated.
+        // all the hooking is actually done inside ClipboardWebGL class.
+    }
+#endif
```
BrianAmadori: Why not just create an instance?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
pbosio: LOL I did not add that, but I'm going to modify it, didn't see it in there 😅

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/Implementation/Web/ClipboardWebGL.cs
pbosio: Needs to be a singleton cause it need the static reference for the javascript callbacks.
It's instantiated inside the static class Clipboard

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/ClipboardReadPromise.cs
pbosio: Do you mean creating a Generic Promise? that's a good idea

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/Implementation/IClipboardImplementation.cs
pbosio: I will rename it, but these are more platform wise implementations than handlers

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/Clipboard.cs
pbosio: I think initializing or setting something up here will end in a unnecessary messy code when all we just need is for the static class to be created

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/Implementation/Web/Plugins/ClipboardWebGL.jslib
pbosio: That's true, but this is not something that needs maintaining or debugging, it's mainly hooking to callbacks from the browser

BrianAmadori: 1) All code needs maintaining and debugging down the road.

2) This code is interacting with `unityInterface` (kernel specific object) and relying on `eval` because we can't use promises here. This looks OK to you? You'd allow other devs to take this approach down the line?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/Clipboard.cs
BrianAmadori: Here the code is relying on a language side-effect to get things ready, this is BAD. There has to be a better way to design this feature. If you can't come up with a better approach, let's talk this for a bit to see if we can rubber-duck a bit and improve it.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/Tests/ClipboardTests.cs
```diff
@@ -0,0 +1,45 @@
...
+    protected override IEnumerator SetUp()
+    {
+        yield break;
+    }
+
+    [UnityTearDown]
+    protected override IEnumerator TearDown()
+    {
+        yield break;
+    }
```
BrianAmadori: I think we can safely remove the `TestsBase` inheritance here, and thus don't need the `SetUp` and `TearDown` methods.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Environment.cs
```diff
@@ -22,6 +23,7 @@ private Environment()
             messagingControllersManager = new MessagingControllersManager();
             pointerEventsController = new PointerEventsController();
             memoryManager = new MemoryManager();
+            clipboard = Clipboard.Create();
```
BrianAmadori, line 26: ❤️

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Promise.cs
```diff
@@ -0,0 +1,65 @@
+using System;
+using UnityEngine;
+
+namespace DCL.Helpers
+{
+    public class Promise<T> : CustomYieldInstruction
```
BrianAmadori, line 6: ❤️

BrianAmadori (approved):
Thanks for applying the requests 😍 

Left a mini-comment.

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Promise.cs
```diff
@@ -0,0 +1,65 @@
...
+            }
+            else if (failed)
+            {
+                errorCallback?.Invoke(this.error);
+            }
+        }
+
+        public Promise()
+        {
+        }
```
BrianAmadori:
```suggestion
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Clipboard/Tests/ClipboardTests.cs
pbosio: done 👍

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Promise.cs
pbosio: removed 😅

sandrade-dcl (approved):
Code reviewed! Only a thing: I have tested this build activating the new ProfileHUD (using `ENABLE_NEW_TASKBAR` url param) but I notice that the copy/paste behaviour is still not working with the wallet address... is it normal? do you have to apply some integration in the ProfileHUD after commit this PR to make it works?

Issue #937, opened 2020-05-26
Cylinders look like cones


nearnshaw:
The CylinderShape has a radiusTop property that starts as 0 by default, this makes it look like a cone, identical to ConeShape.
The default value should be equal to the radius at the bottom

This issue has been around and reported many times since we moved to Unity.

PR #1301, opened 2020-09-15
chore: CylinderShape now defaults its top radius to 1, matching the bottom one.

AjimenezDCL:
#937


sandrade-dcl (approved):

pravusjif (approved):
Just as a reminder, this will keep being broken in already-deployed scenes because of the SDK code they carry

AjimenezDCL:
It will affect only those cases were people used a CylinderShape wanting a ConeShape and left the value as default. It's a really edge case =D.

Issue #1072, opened 2020-06-30
transparencyMode issues in materials


nearnshaw:
Here we see a planeshape that appears to have z-fighting issues with what's behind it, even though there's a significant distance in between

![image.png](https://images.zenhubusercontent.com/5d9b940e491c060001c8647b/0ce71f83-2775-4836-9d39-5c8912f83fee)

[https://app.zenhub.com/files/220994476/b53ddd64-9638-41d5-ba3a-979778dc26fc/download](https://app.zenhub.com/files/220994476/b53ddd64-9638-41d5-ba3a-979778dc26fc/download)

Here the material is using a same image as the alphaTexture and the albedoTexture and transparencyMode is not defined

spriteMaterial.albedoTexture = spriteTexture;
spriteMaterial.alphaTexture = spriteTexture;

If I add an extra line to this scene with a transparencyMode = 1, it fixes it

spriteMaterial.transparencyMode = 1

But if I set that to 2 or 3, if fails again. It fails if undefined because 2 is the default value of transparencyMode.

   * | Value | Type                                           |
   * | ----- | ---------------------------------------------- |
   * | 0     | OPAQUE  (default)                              |
   * | 1     | ALPHATEST                                      |
   * | 2     | ALPHABLEND                                     |
   * | 3     | ALPHATESTANDBLEND                              |
   * | 4     | AUTO (ALPHABLEND if alpha OPAQUE otherwise     |
   */

So ALPHATEST works, ALPHABLEND and ALPHATESTANDBLEND fail.

Here's a second example where an alphaTexture is not working.

https://github.com/MetaZoneio/AlphaDemo

In this case, we're applying a different texture in alphaTexture from the one in albedoTexture. We should be seeing that a couple of these plains have a round shape being cut out, but the alphaTexture is not affecting what we see in any of the cases.

![Screen Shot 2020-06-30 at 10.15.23.png](https://images.zenhubusercontent.com/5d9b940e491c060001c8647b/d253c4c8-f9f8-4149-9f82-8856ddd27d48)

We can edit the transparencyMode property for all of them, and set any value we want there, but nothing changes with any value we set.

pabloes:
Recently in discord, similar issue with transparency was exposed, but in this case, it was modeled from blender:
https://discordapp.com/channels/417796904760639509/595984432096346142/736602898292670525

BrianAmadori:
Transparency order issues were already resolved. I leave this issue open until we fix the alphaTexture one.

PR #1302, opened 2020-09-15
fix: added AlphaTexture

AjimenezDCL:
#1072 
Alpha texture is now implemented properly in Unity. Said texture can be single-channel and otherwise it will read the first component (R in a standard RGBA).


BrianAmadori (commented):
Neat!. Maybe we can add a visual test to make sure this feature keeps working? The PBRMaterial/BasicMaterial classes are pretty ugly so having the tests up to date can be very helpful with a refactor.

sandrade-dcl (approved):

pravusjif (approved):

PR #1304, opened 2020-09-15
feat: avatar modifier area

nchamo:
We are now adding the concept of an avatar modifier area. These are areas that when modify all avatars that enter it.  In this first iteration, we are supporting two modifiers:
* HIDE_AVATARS: all avatars that enter such area are hidden
* DISABLE_PASSPORTS: passports can't be opened for avatars that enter such area

Also, the only type of area we support right now, is a box shaped one. In the future, we might want to support other types of areas


unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+
+public class AvatarModifierArea : BaseComponent
+{
+    [Serializable]
+    public class Model
+    {
+        // TODO: Change to TriggerArea and handle deserialization with subclasses
+        public BoxTriggerArea area;
```
nchamo, line 14: JSON deserialization doesn't detect subclasses correctly. So I set the only subclass we have right now. If/when we decide to support other areas, we will need to change this

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/TriggerArea.cs
```diff
@@ -0,0 +1,27 @@
+using UnityEngine;
+
+[System.Serializable]
+public abstract class TriggerArea
+{
+    public abstract void AddCollider(GameObject avatarModifierArea);
```
nchamo: Areas add the collider that matches their shape

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarVisibility.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using UnityEngine;
+using UnityEngine.Assertions;
+
+namespace DCL
+{
+    public class AvatarVisibility : MonoBehaviour
```
nchamo, line 11: We added this new behaviour for a number of reasons:
1. Other avatars behave differently than the own avatar. We wanted to handle that difference in one unique place
1. Hiding other avatar requieres to hide other game objects. For example the canvas & the passport collider
1. Previously, avatars were hidden by deactivating the avatar gameobject. However, we need when this happened, the position wasn't modified and the collider stopped triggering events. So we hide the new complexity in this behavour
1. Changing camera mode, would switch the own avatar on and off. Now, if we are in an area that hides the avatar, this needs to change. So we add the ability to "lock" whether the avatar is displayed or not

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PhysicsSyncController/PhysicsSyncController.cs
```diff
@@ -24,5 +24,6 @@ public void Sync()
 
         isDirty = false;
         Physics.SyncTransforms();
+        Physics.Simulate(Time.fixedDeltaTime);
```
nchamo, line 27: IMPORTANT!!!

Changes to the physics controller

pravusjif: I don't understand why other physics collisions were working if that line was missing XD, since we aren't doing the manual `Physics.Simulate` anywhere else, I'd expect the character to go through every collider, prior to this PR

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
```diff
@@ -499,13 +499,23 @@ public void SetVisibility(bool newVisibility)
                 gameObject.SetActive(newVisibility);
         }
 
-        private void HideAll()
+        public void HideAllRenderers()
         {
             Renderer[] renderers = gameObject.GetComponentsInChildren<Renderer>();
 
-            for (int i = 0; i < renderers.Length; i++)
+            foreach (Renderer renderer in renderers)
```
BrianAmadori: We generally avoid using `foreach` when working with arrays because its a bit slower and more GC trash is created. In this case we can live with this because I don't think this is a performance critical code, but take this into account for the future.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarVisibility.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+
+        public AvatarRenderer renderer;
+        [Tooltip("Game objects that should be hidden/shown together with the renderer")]
+        public GameObject[] gameObjectsToToggle;
+        // When this property is true, then the visibility can't be modified
+        private Boolean locked = false;
```
BrianAmadori: Why we are using `System.Boolean` here instead of just `bool`?

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarVisibility.cs
```diff
@@ -0,0 +1,49 @@
...
+        private Boolean? isVisible;
+
+        public void SetVisibility(Boolean visibility)
+        {
+            if (locked == false && isVisible != visibility)
```
BrianAmadori: This lock logic works fine here, but may produce some issues with avatars that are mid-loading and enter the area, as the renderers `enabled` state is enabled in `WearableController.SetAssetRenderersEnabled` (and potentially elsewhere). 

I think we should hide the avatars in a way that doesn't overlap the `renderers.enabled` value or at the very least wrap all the state changes with this approach.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
```diff
@@ -0,0 +1,81 @@
...
+            // Add all listeners
+            foreach (string modifierKey in model.modifiers)
+            {
+                AvatarModifier modifier = this.modifiers[modifierKey];
+                OnAvatarEnter += modifier.ApplyModifier;
+                OnAvatarExit += modifier.RemoveModifier;
+            }
+        }
+
+        yield return null;
```
BrianAmadori: We don't need to skip frames in this call. This is a performance critical path so its important:
```suggestion
        yield break;
```

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private void OnDestroy()
+    {
+        Collider collider = model?.area?.GetCollider(gameObject);
+        if (collider != null)
+        {
+            Destroy(collider);
+        }
```
BrianAmadori: The null check is not needed, if you pass `null` to `Object.Destroy` it will behave just fine.

BrianAmadori (changes requested):
Good job! 👍 . Aside of all the comments, please remember to add tests coverage.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PhysicsSyncController/PhysicsSyncController.cs
BrianAmadori: Nice catch. I believe we aren't using the OnTrigger events elsewhere? This may be the reason this wasn't needed.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
nchamo: Reverted the change 😄

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarVisibility.cs
nchamo: Changed to `bool`

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
nchamo: Done :)

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
nchamo: Done :)

unity-client/Assets/Scenes/InitialScene.unity
```diff
@@ -38,7 +38,7 @@ RenderSettings:
   m_ReflectionIntensity: 1
   m_CustomReflection: {fileID: 8900000, guid: ade92094798774df4a51b327374145b8, type: 3}
   m_Sun: {fileID: 943100645024490816}
-  m_IndirectSpecularColor: {r: 0.27771077, g: 0.65681255, b: 0.9724352, a: 1}
+  m_IndirectSpecularColor: {r: 0.27765352, g: 0.65676206, b: 0.96994627, a: 1}
   m_UseRadianceAmbientProbe: 0
 --- !u!157 &3
 LightmapSettings:
```
pravusjif: Do we really need to modify the `InitialScene.unity`? remember to modify prefabs instead of scenes whenever you can, so that we avoid:
- conflicts with the scene and other branches (and possibly losing changes when merging)
- losing changes when we are in test scenes or in another possible scene (test and visual test use 2 different unity scenes, if the changes weren't done on prefabs those changes won't be there)

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class AvatarModifierArea : BaseComponent
+{
+    [Serializable]
+    public class Model
+    {
+        // TODO: Change to TriggerArea and handle deserialization with subclasses
```
pravusjif, line 13: This will be tackled in another PR or can we remove the comment?

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/TriggerArea.cs
```diff
@@ -0,0 +1,27 @@
...
+    public override void AddCollider(GameObject avatarModifierArea)
+    {
+        BoxCollider boxCollider = avatarModifierArea.AddComponent<BoxCollider>();
+        boxCollider.isTrigger = true;
+        boxCollider.size = box;
+    }
+
+    public override Collider GetCollider(GameObject avatarModifierArea)
+    {
+        return avatarModifierArea.GetComponent<BoxCollider>();
```
pravusjif: We should caché the `boxCollider` in the class (at `AddCollider`), as every GetComponent call is costly and should be avoided whenever we can

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
```diff
@@ -509,6 +509,16 @@ private void HideAll()
...
+        public void ShowAllRenderers()
+        {
+            Renderer[] renderers = gameObject.GetComponentsInChildren<Renderer>();
+
+            for (int i = 0; i < renderers.Length; i++)
+            {
+                renderers[i].enabled = true;
+            }
+        }
+
```
pravusjif: every `WearableController` already has a list of all the renderers involved, so we could traverse the `wearableControllers` collection and call their `SetAssetRenderersEnabled(false/true)` to toggle the renderers.

If that works, we avoid calling GetComponentsInChildren which is costly. If it doesn't work out of the box like that, you can disregard this comment

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarVisibility.cs
```diff
@@ -0,0 +1,49 @@
...
+
+                isVisible = visibility;
+            }
+
+        }
+
+        public void SetLock(bool newLockValue)
+        {
+            locked = newLockValue;
+        }
```
pravusjif: We should probably call `if(!locked) SetVisibility(isVisible);` after this lone line in `SetLock()` to resume this class's function once it is unlocked. 

What do you think?

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarVisibility.cs
pravusjif: I think it's probably better if we move this `SetVisibility` (and its locking) into the `AvatarRenderer` itself, I understand that the behaviour wouldn't be the same for other avatars and the player avatar, but the `CharacterController` has its own `AvatarRenderer` (in a child GameObject), can't we use that in the same manner as we use the AvatarRenderer of the other players?

As Brian says we could unify the logic there, and instead of toggling the gameobject in `AvatarRenderer.SetVisibility()` we can use the `WearableController.SetAssetRenderersEnabled()` of each item in `AvatarRenderer.wearableControllers`

pravusjif (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarVisibility.cs
pravusjif: I think it's probably better if we move this `SetVisibility` (and its locking) into the `AvatarRenderer` itself, I understand that the behaviour wouldn't be the same for other avatars and the player avatar, but the `CharacterController` has its own `AvatarRenderer` (in a child GameObject), can't we use that in the same manner as we use the AvatarRenderer of the other players?

As Brian says we could unify the logic there, and instead of toggling the gameobject in `AvatarRenderer.SetVisibility()` we can use the `WearableController.SetAssetRenderersEnabled()` of each item in `AvatarRenderer.wearableControllers`

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
nchamo: This is related to the comment below:

> JSON deserialization doesn't detect subclasses correctly. So I set the only subclass we have right now. If/when we decide to support other areas, we will need to change this

If we add a new type of area, it won't work directly. That comment is there so that people understand what needs to be done

unity-client/Assets/Scenes/InitialScene.unity
nchamo: Reverted! Not sure why that change happened. Wasn't intended

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PlayerAvatarController/PlayerAvatarController.cs
```diff
@@ -53,15 +53,10 @@ private void Update()
 
     private void SetVisibility(bool shouldBeVisible)
     {
-        AvatarVisibility avatarVisibility = avatarRenderer.gameObject.GetComponent<AvatarVisibility>();
+        AvatarVisibility avatarVisibility = gameObject.GetComponent<AvatarVisibility>();
```
pravusjif: We should caché this `avatarVisibility` at class-level, `GetComponent()` calls are costly.

You can save that component inside the `Start()`, that is only called once.

pravusjif (changes requested):
last mini request

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/Tests/AvatarVisibilityTests.cs
```diff
@@ -0,0 +1,56 @@
...
+using DCL;
+using DCL.Helpers;
+using NUnit.Framework;
+using System.Collections;
+using UnityEngine;
+using UnityEngine.TestTools;
+
+namespace Tests
+{
+    public class AvatarVisibilityTests
```
BrianAmadori: We are starting to use the `Should` naming scheme for tests, as it leads for shorter test names that are a bit more clear.

```suggestion
    public class AvatarVisibilityShould
```

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/Tests/AvatarVisibilityTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        [TearDown]
+        protected void TearDown()
+        {
+            Object.Destroy(parentGameObject);
+        }
+
+        [Test]
+        public void WhenOneCallerSetsTheVisibilityToFalseThenTheGameObjectsAreDeactivated()
```
BrianAmadori:
```suggestion
        public void DeactivateGameObjectsWhenVisibilityIsSetToFalse()
```

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/Tests/AvatarVisibilityTests.cs
```diff
@@ -0,0 +1,56 @@
...
+
+            Assert.IsTrue(toggledGameObject.activeSelf);
+
+            visibility.SetVisibility("Caller1", true);
+            visibility.SetVisibility("Caller2", false);
+            Assert.IsFalse(toggledGameObject.activeSelf);
+        }
+
+        [Test]
+        public void WhenVisibilityIsSetToTrueThenObjectsAreReactivated()
```
BrianAmadori:
```suggestion
        public void ReactivateGameObjectsWhenVisibilityIsSetToTrue()
```

BrianAmadori (approved):

PR #1310, opened 2020-09-16
fix: navmap toast show default thumbnail

pbosio:


sandrade-dcl (approved):
For me... much better now!

BrianAmadori (approved):

PR #1314, opened 2020-09-16
fix: asset bundle conversion dependency issues

BrianAmadori:
In the asset bundle conversor context, GLBs with buffers as external dependencies (used mainly for animations) were injected in the `PersistentAssetCache` with a wrong ID, leading to conversion failure. 

I added some fixes so the tests pass, but another refactor PR will be created to make the tests internals more mockeable in order to expand the coverage.


unity-client/Assets/Batch/AssetBundleMenuItems.cs
```diff
@@ -55,6 +57,15 @@ static void DumpAreaToMax(AssetBundleBuilder builder, int x, int y)
             builder.DumpArea(new Vector2Int(x, y), new Vector2Int(10, 10), (error) => DumpAreaToMax(builder, nextX, nextY));
         }
 
+        [MenuItem("Decentraland/Asset Bundle Builder/Dump Tomb Chaser")]
```
AjimenezDCL, line 60: I guess this is debug code to be able to quickly check that the fix worked with the affected scene. Is there any reason to keep it?

unity-client/Assets/Batch/AssetBundleBuilder.cs
```diff
@@ -55,7 +56,7 @@ public enum ErrorCodes
 
         public Dictionary<string, string> hashLowercaseToHashProper = new Dictionary<string, string>();
 
-        internal ContentServerUtils.ApiEnvironment environment = ContentServerUtils.ApiEnvironment.ORG;
+        internal ContentServerUtils.ApiTLD tld = ContentServerUtils.ApiTLD.ORG;
```
AjimenezDCL, line 59: What does TLD stand for? Not a big fan of acronyms unless they are plain obvious

AjimenezDCL (approved):

pravusjif (approved):

unity-client/Assets/Batch/AssetBundleBuilder.cs
BrianAmadori: TLD = Top level domain, the reason for the change is that same term is used for this purpose at kernel, so I thought it'd be better to improve terms ubiquity between both projects.

unity-client/Assets/Batch/AssetBundleMenuItems.cs
BrianAmadori: If you look at this file, there are several similar menu items that dump several locations. I use them from time to time for testing and deleting it'd imply having to rewrite it each time I want to test this scene.

A better approach'd be to build a custom editor window for this purpose, but for now this is the only way I have to test important scenes.

PR #1315, opened 2020-09-16
fix: equirectangular skybox

BrianAmadori:
Why?:
This reduces the skybox GPU draw calls from 6 to 1, and this translates in increased rendering performance.

What:
The skybox is now rendered using a single equirectangular texture instead of sampling the 6 faces of a cubemap.


pravusjif (approved):

sandrade-dcl (approved):

BrianAmadori:
Merge pending until we fix art issues (skybox doesn't match exactly the one in prod)

BrianAmadori:
Going to re-implement this change with our new cubemap based skybox shader. Closing it...

PR #1317, opened 2020-09-17
fix: Remove TutorialController from MainTest scene

sandrade-dcl:
# What? 
Remove TutorialController instance from the MainTest scene.

# Why? 
All the Unity tests were being executed with the tutorial opened and it could affect to the tests performance. After refactorize the tutorial tests, it is no longer necessary to have the instance in the MainTest scene.


AjimenezDCL (approved):

AjimenezDCL:
Im leaving an approve already but I will test if when CI is done

PR #1325, opened 2020-09-17
fix: Audio settings now affect voice chat too

pablitar:
Connected Audio Settings volume with Voice Chat implementation


nchamo (approved):
Great work!

pravusjif (approved):

PR #1327, opened 2020-09-17
fix: camera jitter in moving platforms

pravusjif:
**WHY?**
right now the **third person camera** tries to accomodate its vertical position always and that gets clunky when the character is **on a vertical moving platform**, leading to a bad UX

**WHAT?**
We are disabling the camera positioning auto-damping when the chara is **on a moving platform** and resetting those values when he **gets out of the moving platform**

**TESTING**
You can test this in Genesis Plaza's air balloon moving platform, **specially when moving up/down and having the camera pointing down to the character** (instead of having the camera at the character-level or below), by comparing the camera jittering/stuttering in prod and in this branch:

prod: http://play.decentraland.org/
this branch: https://explorer.decentraland.zone/branch/fix/CameraJitterInMovingPlatforms/index.html?ENV=org


sandrade-dcl (approved):
Code reviewed and tested. The UX now is much better!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -544,16 +544,25 @@ public BaseComponent EntityComponentCreateOrUpdate(string entityId, CLASS_ID_COM
             {
                 MessageDecoder.DecodeTransform(data, ref DCLTransform.model);
 
+                TransformLerpController lerpController = null;
                 if (!entity.components.ContainsKey(classId))
+                {
                     entity.components.Add(classId, null);
 
+                    lerpController = entity.gameObject.AddComponent<TransformLerpController>();
```
BrianAmadori: Caution: enabling lerping for all the transforms will cause a lot of unwanted behaviour. We should at least enable this only on transforms that we are landed to. Remember the air balloon is animation based, this could break other system-based movements.

BrianAmadori (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
pravusjif: Right now the `TransformLerpController` is doing nothing different than the behaviour we already had, because we are not using its `lerpingSpeed`.

I just left that component there so that when we introduce transform lerping we could just start using that speed property (be it by calculating the speed before or reading it from a property of the SDK component model), and we could also add a different lerping speed for rotation and scaling.

AjimenezDCL (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
pravusjif: I removed the TransformLerpController usage

BrianAmadori (approved):

PR #1344, opened 2020-09-21
feat: restore decentra libs

menduz:
Restores #1313 

New changes:
- Changed build scripts to ensure amd.js is included in `decentraland-ecs/artifacts` folder. That was causing a regression.
- Enabled scene logs in `preview` mode
- typescript property `useDefineForClassFields` must be false to work with the ECS decorators [more info here](https://github.com/arduz-io/ulla/commit/8ccf73f34de07c96a2ea9fe86d97cd24e628dd63)

For future PR: when Typescript 4.1 reaches GA, we should implement this automatic configuration for libraries only:
- [Blog post](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1-beta/#declaration-and-outfile-requires-a-package-name-root)
- https://github.com/microsoft/TypeScript/pull/39277


kernel/packages/build-ecs/tsconfig.json
```diff
@@ -10,7 +10,13 @@
     "emitDecoratorMetadata": true,
     "declaration": false,
     "downlevelIteration": true,
-    "keyofStringsOnly": true
+    "keyofStringsOnly": true,
```
pablitar, line 13: I think we should leave out this checks, or at least make them optional somehow (have a clear and easy way to remove them).

Otherwise, scene developers are going to see their code not compiling after an update and are going to freak out:
image

pablitar (approved):
LGTM! Let us know when this is ready to merge so we can time it with the release schedule :)

menduz:
The missing change will be deferred until ts4.1 gets launched in GA and it will be a separate PR. 

This PR is ready to merge now.

PR #1350, opened 2020-09-22
fix: wss mode basecatalogue

pravusjif:
modified basecatalog urls when wss is enabled


sandrade-dcl (approved):

kernel/static/default-profile/download_all.js
```diff
@@ -19,7 +21,7 @@ const downloadFile = async (url, path) => {
   })
 }
 
-const catalog = require('./basecatalog.json')
+const catalog = WSS_ENABLED ? require('./basecatalog-wss.json') : require('./basecatalog.json')
```
pbosio: Shouldn't must we also do something like that inside the profile saga for this to work?
`const response = yield fetch(getResourcesURL() + '/default-profile/basecatalog.json')`

kernel/packages/shared/profiles/sagas.ts
```diff
@@ -274,7 +274,8 @@ export function* initialLoad() {
   } else {
     let baseCatalog = []
     try {
-      const response = yield fetch(getResourcesURL() + '/default-profile/basecatalog.json')
+      const catalogPath = WSS_ENABLED ? '/default-profile/basecatalog.json' : '/default-profile/basecatalog-wss.json'
```
pbosio: I think you have this backwards :P
```suggestion
      const catalogPath = WSS_ENABLED ? '/default-profile/basecatalog-wss.json' : '/default-profile/basecatalog.json'
```

nchamo (approved):

pravusjif (commented):
I just tested it with a preview mode scene and it worked like a charm, thx Pato!

PR #1353, opened 2020-09-23
fix: replace trigger with OverlapBox on AvatarModifierArea

nchamo:
When we implemented the feature `AvatarModifierArea`, we used triggers to detect when avatars entered the area. However, for these triggers to work, we had to use `Physics.Simulate(Time.fixedDeltaTime);`.

This caused a performance regression, so we will use `OverlapBox` instead.


unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
```diff
@@ -61,19 +61,58 @@ public override IEnumerator ApplyChanges(string newJson)
...
+        foreach (GameObject avatarThatEntered in newAvatarsInArea.Except(avatarsInArea))
+        {
+            OnAvatarEnter?.Invoke(avatarThatEntered);
+        }
+
+        // Call events for avatars that just exited the area
+        foreach (GameObject avatarThatExited in avatarsInArea.Except(newAvatarsInArea))
+        {
+            OnAvatarExit?.Invoke(avatarThatExited);
+        }
```
pravusjif: If there is no change in the avatars contained inside the hashset/area, we should escape before traversing all the elements, to avoid that unnecessary performance cost

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
```diff
@@ -61,19 +61,58 @@ public override IEnumerator ApplyChanges(string newJson)
...
     }
 
-    private void OnTriggerEnter(Collider other)
+    private GameObject[] DetectAllAvatarsInArea()
     {
-        Transform parent = other.transform.parent;
-        OnAvatarEnter?.Invoke(parent.gameObject);
+        if (entity?.gameObject == null)
+        {
+            return new GameObject[0];
```
pravusjif: I think we could return null here instead of generating a new empty array, I just looked it up and it seems an empty array actually allocates memory in .NET, so to be on the safe side let's return null here.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/TriggerArea.cs
```diff
@@ -12,11 +13,10 @@ public class BoxTriggerArea : TriggerArea
...
-    public override Collider AddCollider(GameObject avatarModifierArea)
+    public override GameObject[] DetectAvatars(Vector3 center, Quaternion rotation)
     {
-        BoxCollider boxCollider = avatarModifierArea.AddComponent<BoxCollider>();
-        boxCollider.isTrigger = true;
-        boxCollider.size = box;
-        return boxCollider;
+        Collider[] colliders = Physics.OverlapBox(center, box / 2, rotation, LayerMask.GetMask("AvatarTriggerDetection"), QueryTriggerInteraction.Collide);
+        return colliders.Select(collider => collider.transform.parent.gameObject)
+            .ToArray();
```
pravusjif, line 20: I think this `TriggerArea` class is no longer needed and we could do this detection in the `AvatarModifierArea` instead, what do you think?

Besides that, I try to avoid using Linq as it comes with a performance cost, in this case the query will be called on every frame (and every avatar modifier area) so it would be better if we can avoid that usage here. If this class disappears and we move this OverlapBox to the `AvatarModifierArea`, we could just populate the hashset with the corresponding colliders instead of dealing with an array with Linq.

pravusjif (changes requested):

pravusjif (approved):
Beautiful, also checked the unity performance stats (in editor) in ZONE 64,-64 and performance is back to normal regarding physics

Issue #828, opened 2020-04-20
Shadeless materials not supported


nearnshaw:
The `BasicMaterial` should not have any shadows, but here I see it affected, both by the light direction and by the presence of other entities casting a shadow.

The `Material` has a `disableLighting` property, but it's not working. If I set it, the material becomes bright pink.

```
let basicM = new BasicMaterial()

let noLightMaterial = new Material()
noLightMaterial.disableLighting = true
noLightMaterial.albedoColor = Color4.White()

const myBox = new Entity()
const box = new BoxShape()
myBox.addComponent(box)
myBox.addComponent(new Transform({ position: new Vector3(3, 1, 3) }))
engine.addEntity(myBox)
myBox.addComponent(basicM)

const myBox2 = new Entity()
const box2 = new BoxShape()
myBox2.addComponent(box)
myBox2.addComponent(
  new Transform({ position: new Vector3(1, 1, 1), scale: new Vector3(2, 6, 2) })
)
engine.addEntity(myBox2)
myBox2.addComponent(basicM)

const myBox3 = new Entity()
const box3 = new BoxShape()
myBox3.addComponent(box)
myBox3.addComponent(new Transform({ position: new Vector3(2.8, 1, 1.8) }))
engine.addEntity(myBox3)
myBox3.addComponent(noLightMaterial)
```

![materials.jpg](https://images.zenhubusercontent.com/5d9b940e491c060001c8647b/b7c31876-152f-4a62-b7d5-1149f2f86e09)
Here the two white cubes have a basic material (but notice the shadows), and the pink one has a normal Material w disableLighting

BrianAmadori:
@nearnshaw  As you suspected, the shader used for this material is the same as the PBR one. If we add a toggle to enable the lighting this material would be a poor's man PBR. Maybe it's just wiser to remove the lighting altogether assigning a different shader to the BasicMaterial and adding a toggle for shadows to both material components.

To summarize, with proposal of the description of this issue, we'd have:
- BasicMaterial with toggleable shading and shadow. If the shading is ON, it would look the same as PBRMaterial but with less properties. There isn't any way of having no shading but enabled shadows or viceversa.
- PBRMaterial as PBR, without toggleable shadow

With the suggested proposal, we'd have:
- BasicMaterial as shadeless, with toggleable shadow
- PBRMaterial as PBR, with toggleable shadow

With the new proposal, If anyone want to switch from a lighting enabled to a shadeless, all it'd take is to change the material.

[...]

PR #1356, opened 2020-09-24
fix: basic material + shadow support

AjimenezDCL:
#828


sandrade-dcl (approved):
Code reviewed, but I would like that someone with more experience in shaders also review it :)

pravusjif (approved):
The code looks good, didn't see any problem testing the build

BrianAmadori (commented):
We should at least add a few unit test/visual tests to ensure this feature keeps working in the future.

PR #1365, opened 2020-09-25
fix: asset bundle texture issues

BrianAmadori:
This PR includes the following changes:

- Now, missing external texture errors that prevented conversion are handled gracefully. This is done by using an 1x1 white texture on the affected cases.
- Fix for the `DXTnm` issue with albedo textures (red textures issue). Now textures will not be tagged as normal map if at least one gltf is using it as a color map.
- All external textures weren't being configured at all when they should.
- Removed stacktrace for non-error log types to prevent log spam.
- Forced compression to high quality with crunch enabled to see if this can bail us out of compression artifacts while keeping the size reduction.

Tested converting at: 
- [x] -119, 135 (baus casino) 
- [x] 12, 45 (tomb chaser)

Added local tests to cover all the issues.


AjimenezDCL (approved):

pravusjif (approved):

pravusjif:
Also tested the build with the `NO_ASSET_BUNDLES` URL param and all good

PR #1366, opened 2020-09-28
fix: Asset_Cache mismatch on textureIds

AjimenezDCL:
GLTFImporter was not setting the ID of the RefCountedTextureData properly. This ended up on some textures marked as cached but destroyed.


pravusjif (approved):

unity-client/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -2432,11 +2432,12 @@ public virtual Texture GetTexture(int textureIndex)
...
                     yield return ConstructImage(image, sourceId, markGpuOnly, isLinear);
-                    source = new RefCountedTextureData(image.Uri, _assetCache.ImageCache[sourceId]);
+                    string fullId = PersistentAssetCache.GetCacheId(image.Uri, id);
+                    source = new RefCountedTextureData(fullId, _assetCache.ImageCache[sourceId]);
 
-                    if (image.Uri != null && addImagesToPersistentCaching)
+                    if (fullId != null && addImagesToPersistentCaching)
                     {
-                        PersistentAssetCache.AddImage(image.Uri, id, source);
+                        PersistentAssetCache.AddImage(fullId, source);
```
BrianAmadori: This is a nitpick, but oh well: On the base code, `GetCacheId` would only get called if the `if` succeeds, so I don't see this change as an improvement.

BrianAmadori (approved):
🙌 Remember to add a small test (or a few asserts to existing `AssetPromise_Texture` tests) to ensure the cache is cleaned up correctly when any texture is being cleaned.

unity-client/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
AjimenezDCL: In fact this is actually the fix to the issue, the other `RemoveImage` is something I found along the way.

Here: https://github.com/decentraland/explorer/blob/a71559e2a9f0c198fa93770205dadb5dfcebf269/unity-client/Assets/UnityGLTF/Scripts/Cache/RefCountedTextureData.cs#L19
Before destroying the texture we are attempting to Remove the image `id` from the cache.
But here: https://github.com/decentraland/explorer/blob/a71559e2a9f0c198fa93770205dadb5dfcebf269/unity-client/Assets/UnityGLTF/Scripts/Cache/AssetCache.cs#L27
The Texture is cached with the full id.

AjimenezDCL: Maybe I didnt express myself clearly but basically the id stored in the `RefCountedTextureData` was different than the one cached, so the `Texture` was being destroyed but not removed from the cache and therefore throwing a bunch of `You are trying to use a destroyed Texture2D` errors.

BrianAmadori: Oh, now I understand!

Then, maybe this is a shallow fix - the real issue here is that `PersistentAssetCache.AddImage` doesn't ensure the underlying id in the `RefCountedTextureData` matches the used one. 

I think we have a design flaw here... maybe a better fix is to just set the id when using `AddImage`?

```
        public static void AddImage(string uri, string idSuffix, RefCountedTextureData refCountedTexture)
        {
            var key = GetCacheId(uri, idSuffix);
            refCountedTexture.id = key;
            ImageCacheByUri[key] = refCountedTexture;
        }
```

The issue with this, is that `id` is private. So maybe creating the `RefCountedTextureData` from outside isn't needed and we can do this?

```
        public static void AddImage(string uri, string idSuffix, Texture2D texture)
        {
            var key = GetCacheId(uri, idSuffix);
            var refCountedData = new RefCountedTextureData( key, texture );
            ImageCacheByUri[key] = refCountedData;
        }
```

On top of this, I believe we also have the `AddBuffer` tidbit with this bug.

https://github.com/decentraland/explorer/blob/da0089b497fa03a5e6b6898722812589907958ac/unity-client/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs#L679

So you've effectively opened a small can of worms here, congratz :P

BrianAmadori (approved):

Issue #285, opened 2019-12-30
Animations play while loading model


nearnshaw:
When an entity's GLTF or GLB model includes animations, the first one of these will always be played on a loop. That's the desired behavior if no Animator is preset. But if there is one, then animations shouldn't play until instructed. The problem is that while the entity is loading, the animation might play for a second before the Animator is recognized. This is especially a problem with things like doors and chests, where playing that animation might mean that the door starts open or partly open.
The (horrible) workaround we found for smart items was to add an idle animation that does nothing, and put that one first in the model. So that when the animation is played there are no consequences. This of course is a pretty bad experience for artists and developers, and we should find ways to avoid it.

PR #1373, opened 2020-09-29
fix: Animation plays on awake with an Animator presented

AjimenezDCL:
#285 
The animator, on intializing, stops any on-going animation. It was working but those few frames the animation was running left the model in a non-default state. That could ruin stuff like doors.


pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/Animator/DCLAnimator.cs
```diff
@@ -71,33 +71,31 @@ private void OnComponentUpdated(DecentralandEntity e)
...
 
-                animComponent.playAutomatically = true;
-                animComponent.enabled = true;
-                animComponent.Stop(); //NOTE(Brian): When the GLTF is created by GLTFSceneImporter a frame may be elapsed, 
-                                      //putting the component in play state if playAutomatically was true at that point.
+            animComponent.playAutomatically = true;
+            animComponent.enabled = true;
+            animComponent.Stop(); //NOTE(Brian): When the GLTF is created by GLTFSceneImporter a frame may be elapsed,
+            //putting the component in play state if playAutomatically was true at that point.
+            animComponent.clip?.SampleAnimation(animComponent.gameObject, 0);
```
BrianAmadori, line 89: I don't get this code. Why using `SampleAnimation` here fixes the bug? 

I understand that this issue is coming from `GLTFSceneImporter` setting `playAutomatically` as true before this component comes along to stop it. What i'm missing?

https://github.com/decentraland/explorer/blob/da0089b497fa03a5e6b6898722812589907958ac/unity-client/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs#L1310

AjimenezDCL: From the ticket: `When an entity's GLTF or GLB model includes animations, the first one of these will always be played on a loop. That's the desired behavior if no Animator is preset`

The problem is we dont know in advance if an animation will have an Animator or not so for a few frames the animation is playing and then is stopped (which leaves the object in the middle of the animation). That was ruining stuff like doors and such.

So sampling the first frame of the animation will leave it as it is.

BrianAmadori: Let's look at what's happening behind this bug (at least from my understanding):

- GLTF starts loading
- GLTF loads the animation, and sets `playAutomatically` to `true`
- Animation starts playing in loop
- A few frames pass by <- This is where the bug happens, few frames with the improper animation play
- The `DCLAnimator` component is created
- `Awake()` gets called, effectively "stopping" the animation

Then, how sampling the animation in the `Awake()` fixes it? In this case, the stop will come too late.

If all of this is true, the fix should be along the lines of:

- GLTF starts loading
- GLTF loads the animation, and sets `playAutomatically` to `false` <- To prevent the early play
- GLTF finishes loading

- Now we should 
      (A) add a `DCLAnimator` component to ALL gltfs or 
      (B) wait a grace time to see if a `DCLAnimator` component exists, if not, we should `Play` the animation to conform with the default behaviour
      (C) ???

- The `DCLAnimator` component is created
- `Awake()` gets called, and animation should be effectively stopped when reached this stage

BrianAmadori (approved):

PR #1376, opened 2020-09-29
chore: allows TEST_WEARABLES as an url

caminoezequiel:
# What?
Allows change default test url for catalog setting TEST_WEARABLES as url

### DIFF
--- a/kernel/packages/config/index.ts
+++ b/kernel/packages/config/index.ts
@@ -243,9 +243,8 @@ export function getExclusiveServer() {
     try {
       return new URL(url.searchParams.get('TEST_WEARABLES') as string).toString()
     } catch (e) {
-      // ignore error -> return default test url
+      return 'https://dcl-wearables-dev.now.sh/index.json'
     }
-    return 'https://dcl-wearables-dev.now.sh/index.json'
   }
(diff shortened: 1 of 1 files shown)


kernel/packages/config/index.ts
```diff
@@ -238,7 +238,13 @@ export function getDefaultTLD() {
 }
 
 export function getExclusiveServer() {
-  if (window.location.search.match(/TEST_WEARABLES/)) {
+  const url = new URL(window.location)
+  if (url.searchParams.has('TEST_WEARABLES')) {
+    try {
+      return new URL(url.searchParams.get('TEST_WEARABLES') as string).toString()
+    } catch (e) {
+      // ignore error -> return default test url
```
nchamo: Super nit: Maybe we could return `return 'https://dcl-wearables-dev.now.sh/index.json'` here, instead of doing it down below?

nchamo (approved):
👍

kernel/packages/config/index.ts
caminoezequiel: it's true! thanks

marcosnc (approved):

Issue #177, opened 2019-12-02
Removing entity from engine removes its components from memory, even if shared


BrianAmadori:
This code doesn't work. Once the box is removed from the engine, adding it again doesn't render it:

```
const box = new Entity();
box.addComponent(new BoxShape());
box.addComponent(new Transform({position:new Vector3(8,0.11,8)}));
engine.addEntity(box);
engine.removeEntity(box);
engine.addEntity(box);
```

Something similar happens if I tried to use the shape that was previously used by a removed entity on a new entity. It seems like when removing from engine, the shape component is no longer available.

Here we should see the box2 entity, but it can't find a shape to be rendered with.

```
const shape = new BoxShape();

const box1 = new Entity();
box1.addComponent(shape);

engine.addEntity(box1);
engine.removeEntity(box1);

const box2 = new Entity();
box2.addComponent(shape);
engine.addEntity(box2);
```

In this other scene we see something that's surely related, and might throw more light into this:
When clicking box 1 you toggle box 2 on and off, but at the 3rd attempt, all boxes disappear.

[add-remove-bug.zip](https://github.com/decentraland/unity-client/files/3498909/add-remove-bug.zip)

Weird fact:  If I add two more cubes to the scene (that don't do anything), I have to toggle the cube two more times to have it all break. It seems like we're losing one of the BoxShape components every time we remove the cube, and as long as at least one exists in memory the scene can work things out, until we run out of them.

nearnshaw:
It looks like the second example, the one in the attached zip, is no longer happening, that's fixed.

But the first problem, is still happening:
```
const box = new Entity();
box.addComponent(new BoxShape());
box.addComponent(new Transform({position:new Vector3(8,0.11,8)}));
engine.addEntity(box);
engine.removeEntity(box);
engine.addEntity(box);
```

PR #1378, opened 2020-09-30
fix: remove entity now clears lossy messages to be processed

AjimenezDCL:
#177 

Removing an entity and adding it back had some unexpected behaviour whith lossy messages (like TransformUpdate) being ignored after the entity was actually removed.


BrianAmadori (approved):
🎉

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingBus.cs
```diff
@@ -139,6 +140,28 @@ public void Enqueue(QueuedSceneMessage message, QueueMode queueMode = QueueMode.
         {
             bool enqueued = true;
 
+            // When removing an entity we have to ensure that the enqueued lossy messages after it are processed and not replaced
```
pravusjif, line 143: This comment is a bit confusing, you mean the lossy messges should be **removed and not processed** instead of **processed and not replaced** right?

AjimenezDCL: Nope.

The issue goes as follow:
You get a batch of messages in this order:
A. Create Entity `Ed` (Reliable)
B. Update Entity `Ed` Transform (0,0,0) (Lossy)
C. Remove Entity `Ed` (Reliable)
D. Create Entity `Ed` (Reliable)
E. Update Entity `Ed` Transform (1,1,1) (Lossy)

The bus will:
Enqueue A
Enqueue B
Enqueue C
Enqueue D

 _Bear in mind that we are just enqueueing nothing has been processed yet_
Now because E is a lossy message, we just care about the last value so instead of Enqueueing E, the bus will find another message of the same type (Message B) and modify the original content (0,0,0) for (1,1,1). so our bus will look as the messages will be processed in this order:
1. Create Entity `Ed` (Reliable)
2. Update Entity `Ed` Transform (1,1,1) (Lossy)
3. Remove Entity `Ed` (Reliable)
4. Create Entity `Ed` (Reliable)

This works great (and enhances performance a lot) but has a side effect. When you remove the entity and create it back, even though `kernel` sent you the messages that you needed (Transform), the transform is never set because it was a lossy message and it was consumed in the message before destroying the entity.

So the fix is detecting when we destroy Entities and unmark the lossy messages affecting said entity so the new Transform wont find a previous enqueued Transform message and instead of replacing it will be put in the queue and processed.

I hope I explained myself clearly, it's a bit confusing

pravusjif (approved):
Also tested the build teleporting around and didn't find any issue

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/MessagingBus.cs
pravusjif: Ohhh OK, much clearer now, thanks!

So in that case I think the comment should clarify that the removed entity's lossy messages are being **removed** instead of **processed and not replaced** because we aren't actually processing those lossy messages we are removing them, deleting them, right?

AjimenezDCL: We are just unmarking the previous messages to the DestroyEntity as Lossy. We are not removing anything.

pravusjif: So removing them from the collection at `unreliableMessages.Remove(key);` is not actually removing the message just kind of "unmarking it" ?

Also, shouldn't this be done every time we remove an entity instead of the moment we enqueue a message in the  `MessagingBus` ?

It doesn't make much sense to me that we keep messages for an entity that we are removing =S. Although in the code i understand we are removing them.

AjimenezDCL: Yeah, the messages are still enqueued, we just remove that message from the list of unreliables(lossy) so they are treated as reliable.

Bear in mind that this is just the messages queue, nothing has been processed yet, these messages has not been applied to the world yet. So the actual removal of the entity comes way after this is done, if we do it then it will be too late and the lossy messages will be already replaced.

Also, even if the entity is gonna be removed we still have to process its messages otherwise the world could have really unexpected behaviours. For example, if  your computer is super slow processing messages and you have a 5 mins queue of messages and you receive an EntityRemove message, you still have to process these 5mins of messages affecting your entity, because it might be affecting the world and it must be deterministic-ish.

pravusjif: I see, ok, that makes sense, thanks!

PR #1380, opened 2020-10-01
fix: AvatarShape calling OnPoolGet at start + update event + builder handling update event

pbosio:
# What? 
AvatarShape: Removed call for `OnPoolGet` from `Start` function.
AvatarShape: Added event triggered when AvatarShape is updated.
Builder: Handle AvatarShape's update.

# Why? 
Calling `OnPoolGet` on `Start` was causing an issue where AvatarShape's model was being overwritten by a new instance before being send to AvatarRenderer, resulting in an invisible Avatar.


BrianAmadori (approved):

pravusjif (approved):
tested in Unity editor and avatars still pool correctly.

PR #1383, opened 2020-10-05
chore: added DCL NSubstitute package + PrivateChat Implementation

AjimenezDCL:
Adding NSubstitue and converting one of our tests as an example.

The NSubstitute fork we are using is located here https://github.com/decentraland/Unity3D-NSubstitute. You can take a look at the `cdd11bb74398afae73e71b6fe01114f597a6ce08` commit for a more generic example of use.

Said generic example can also be checked within Unity:
image


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/ChatController/ChatMessage/ChatMessage.cs
```diff
@@ -11,6 +11,14 @@ public enum Type
             SYSTEM
         }
 
+        public ChatMessage() { }
```
sandrade-dcl, line 14: Why is it necessary to create an empty constructor? Is it only for standard conventions? or is it being used in some place?

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/ChatController/ChatMessage/ChatMessage.cs
AjimenezDCL: So we can use `object initializers` like ` new ChatMessage { messageType = "type" } ;`

pravusjif (approved):

Issue #179, opened 2019-12-02
UV mapping of some models getting messed up


BrianAmadori:
Some models, specifically by Cybermike and Renderdog, used to work well (not sure how long ago), but now have the uv mapping different from what it used to be.

See attached a models that suffered this.
If you importing into the Babylon sandbox in https://sandbox.babylonjs.com/ you'll see that the texture looks different than if you use it in a scene

*Renderdog*:

[https://app.zenhub.com/files/154520831/93e29268-25d1-4df2-a556-83f16e95ca22/download](https://app.zenhub.com/files/154520831/93e29268-25d1-4df2-a556-83f16e95ca22/download)

from RenderDog

SDK has a bug and UV tiling does not work. Is that still the case? I am using blender and I am finding it does not work in DCL but it does on other platforms. In DCL I have been getting around this by having a ground terrain UV scaled beyond the 0-1 coordinates. 

*CyberMike*:

[Models-2.zip](https://github.com/decentraland/unity-client/files/3655200/Models-2.zip)

From Cybermike:

I think it was in version 6.1 onwards, somewhere inbetween 6.1 and current 6.4 (i think) is when this problem showed itself. Also when i do manage to do a work around of moving uv's back to the default channel of 1 (which have not been collapsed and lost) they appear SO DARK in the sdk, you cant see the texture, i had to over expose the texture in ps and it looks bad...  so the problem is uv mapping channel can not go greater then default 1, i think

pravusjif:
I just tested the "RenderDog" glb in the UnityGLTF project (not our modified version) and its UVs are loaded correctly so we should check what we are doing differently or update a part of our GLTFSceneImporter based on their version (as we did with the animations parsing section).

Of course as we have heavily modified the importer, any update there is no simple task.

AjimenezDCL:
We have two different issues here, this ticket will be closed and these two issues will me moved to new tickets:

_RenderDog_  #1389: I dont think this ever worked before, we have never supported UV scale/offset, _RenderDog_'s statement kinda confirms it.

[...]

PR #1388, opened 2020-10-06
feat: implement texture scale/offset

AjimenezDCL:
~#179~ (the issue has been splitted into 2 different issues)

#1389


AjimenezDCL:
This branch must be tested with the `NO_ASSET_BUNDLES` flag. Otherwise the GLTFs wont be imported and the previous materials will load.

unity-client/Assets/UnityGLTF/Scripts/UniformMaps/SpecGloss2StandardMap.cs
```diff
@@ -45,6 +70,30 @@ public virtual int SpecularGlossinessTexCoord
             set { return; }
         }
 
+        private Vector2 specGlossOffset = Vector2.zero;
```
BrianAmadori: All offsets that aren't directed to `_BaseMap` will not work because we aren't declaring the `_ST` uniforms for each one and implementing the corresponding offset UV transform. You can check that the only declared `_ST` is the `_BaseMap_ST` in `LitInput.hlsl`.

The UV transform for `_BaseMap` is computed here: https://github.com/decentraland/explorer/blob/e40d56c939db891b53342822a29da5adf7363d7d/unity-client/Assets/Rendering/Shaders/URP/LitForwardPass.hlsl#L95

I'm afraid that we might not have enough TEXCOORD semantics to support this extension entirely without removing lightmapping support and doing some gymnastics.

BrianAmadori: Also, we should add at least a test with a GLTF cube/plane + offseted material to ensure this is working properly. We are supporting a new extension here, so it deserves the extra mile.

AjimenezDCL: Ah dammit... I mixed `_ST` with samplers and I thought it would work since said samplers are actually declared per texture.

Then I will remove the feature for anything that is not a basemap and talk with @nearnshaw so the documentation reflects it properly. I was also planning of doing a bunch of visual tests as well, I know they are disabled in the CI but they are more reliable to changes on the shaders.

sandrade-dcl (approved):

AjimenezDCL:
Documentation for creators:

Now tiling and offset in textures are supported. Bear in mind that only the one applied to the Albedo/BaseMap is taking into account and it will be used in the Normal or any other PBR texture.

pravusjif (approved):

PR #1394, opened 2020-10-06
feat: world render profiles

BrianAmadori:
This PR includes:

- [X] `RenderProfile` asset, this asset will give us the means of switching visual presets from kernel, like i.e. day/night.
- [X] JSON meta support for `MOTD` and `RenderProfiles`
- [X] Add avatar `_LightColor` customization for profiles.
- [X] `HALLOWEEN` debug flag support that will set the halloween `RenderProfile` when its used.

This is how the asset looks like: 
image


AjimenezDCL (approved):
Looks good, great job.

How can I test it? everything looks normal in the published preview (maybe related to the CI failing to build kernel?).

caminoezequiel (approved):
look good!

kernel/packages/config/index.ts
```diff
@@ -146,6 +146,8 @@ export const ENABLE_NEW_TASKBAR =
   location.search.indexOf('ENABLE_NEW_TASKBAR') !==
   -1 /* NOTE(Santi): This is temporal, until we remove the old taskbar */
 
+export const HALLOWEEN = location.search.indexOf('HALLOWEEN') !== -1
```
pablitar, line 149: Knowing that we will probably have more events in the future that require content-specific changes I think we should explore alternatives to have this not be a particular flag, but rather a setting or group of settings we can toggle on or off without deploying code.

I'm not saying we should do that now, but rather that maybe we should think about how we are going to handle this in the future.

pablitar (approved):

kernel/packages/config/index.ts
BrianAmadori: If you look closely to this PR, we are adding the `WorldConfig` type to the `MetaConfiguration`:

```
export type MetaConfiguration = {
  explorer: {
    minBuildNumber: number
    useUnityIndexedDbCache: boolean
  }
  servers: {
    added: string[]
    denied: string[]
    contentWhitelist: string[]
  }
  world: WorldConfig
  comms: CommsConfig
}

export type WorldConfig = {
  pois: Vector2Component[]
  renderProfile?: RenderProfile
  motd?: MotdConfig
}
```

This means that we will be able to: 
- Turn halloween night mode on/off by setting the `renderProfile` field 
- Customize the message of the day (MOTD)

The `HALLOWEEN` url flag is just for internal testing. We will be able to toggle events entirely using this approach.

PR #1397, opened 2020-10-07
fix: sdk move player avatars lerping

pravusjif:
**WHY?**
Currently, if the SDK `movePlayerto()` is used, the experience is fine being the moved player, but all the other players instead of seeing the instant repositioning they see the avatar walking/flying towards the new position, which pretty much breaks the experience for those other players since they don't perceive that the first player has been respawned/repositioned.

**WHAT?**
Added an `immediate` bool to the position package shared by the comms peers. When the AvatarSystem updates a fellow avatar position, if that `immediate` bool is true, we remove the entity from the engine, update its transform position, and then add it back to the engine. That way we avoid all the trouble around adding a new property in the transform that would only be used for this case.

**TEST**
The only testing strategy right now is
1. checkout this branch locally
2. start unity in LOCAL 0,101
3. at the same time open an incognito window and enter the same scene at http://localhost:8080/?&DEBUG_MODE&position=0%2C101
4. while looking at the guest avatar inside unity editor, move the guest avatar towards the cube in the scene (by controlling it in its own incognito window) and interact with it to be teleported a few meters away from the cube


menduz:
The changes look good. I'd like to explore if we can produce the same effect by removing and re-adding the Transform component.

Adding a new (very specific) field to the Transform protocol may produce unwanted maintenance problems in the future.

```ts
function teleport(entity: Entity, newPosition: Vector3){
  const originalTransform = entity.getComponent(Transform)
  entity.removeComponent(Transform)
  originalTransform.position = newPosition
  entity.addComponent(originalTransform)
}
```

pravusjif:
> The changes look good. I'd like to explore if we can produce the same effect by removing and re-adding the Transform component.
> 
> Adding a new (very specific) field to the Transform protocol may produce unwanted maintenance problems in the future.
> 
> ```ts
> function teleport(entity: Entity, newPosition: Vector3){
>   const originalTransform = entity.getComponent(Transform)
>   entity.removeComponent(Transform)
>   originalTransform.position = newPosition
>   entity.addComponent(originalTransform)
> }
> ```

Good idea, I'll try that today

pravusjif:
> Good idea, I'll try that today

This worked and made the code more simple (no new property in the Transform component and everything involved in that).

Although the approach was changed to re-add the avatar entity to the engine instead of just the Transform Component.
As that's what happens when we remove the transform component in an AvatarEntity:
1. remove the Transform component from the AvatarEntity
2. Kernel removed the component on the entity, then kernel/unity positions the entity far away in space
3. The AvatarEntity has its `handleUserVisible()` method called with a false param and the entity ends up hiding itself (it removes its instance from the engine) and appearing again when the Transform component is added again.

So to make the fix clearer we just remove the entity, update the position, and put the entity back in the engine: this makes the reposition immediate in Unity.

pablitar (approved):
LGTM!

I don't see any logic for handling the possibility in `collectInfo` skips an immediate update. Did I miss something?

But I suppose that may be infrequent, so I guess is OK either way!

pbosio (approved):

menduz (approved):

PR #1402, opened 2020-10-07
fix: rendering activation on teleport

pravusjif:
fixed rendering activation on teleport bug introduced in the following PR: https://github.com/decentraland/explorer/pull/1173.

Removed `OnRenderingStateChange` usage in `SceneController` as it's no longer needed.

The SceneController was locking the rendering state and unlocking it when the new current scene reported it was ready, as the order in which we set the current scene changed, we no longer need this rendering state change event for the lock-unlock and we just use the current scene change event.


sandrade-dcl (approved):

AjimenezDCL (approved):

AjimenezDCL:
Approving since this is a hotfix, but we should have another PR with tests to ensure this wont happen again

pravusjif:
> Approving since this is a hotfix, but we should have another PR with tests to ensure this wont happen again

True that, I'll give it a try in this PR and if I run into too much trouble running the test we can put that in another PR as you say

PR #1404, opened 2020-10-07
feat: remote debugger

menduz:
# What? 
Generate a new runtime (scene system) to run scenes in node.js and enable debugging in VSCode

Changes:
* The `scene.system.ts` file content was moved almost entirely to `DecentralandRuntime.ts`
* Added `cli.scene.system.ts` to run in a [constrained node runtime](https://github.com/decentraland/node-runtime).
* Preview scenes will now work connected to the [CLI via WebSocket ](https://github.com/decentraland/cli/pull/523)when the `?WS_SCENE=1` is added to the URL.

_Important: This is an opt-in feature. This is NOT the default behavior._

# Why? 
Remove limitations of the web platform to develop and test scenes. Node.js will enable not only debugging but also testing tools and quicker development loops.

https://github.com/decentraland/cli/pull/523 <- cli counterpart
https://github.com/decentraland/node-runtime <- runtime for node


kernel/packages/entryPoints/preview.ts
```diff
@@ -25,12 +25,19 @@ if (!container) throw new Error('cannot find element #gameContainer')
...
+  wsScene = `${location.protocol === 'https:' ? 'wss' : 'ws'}://${document.location.host}/?scene`
+}
+
 function startPreviewWatcher() {
   // this is set to avoid double loading scenes due queued messages
   let isSceneLoading: boolean = true
 
   const loadScene = () => {
-    loadPreviewScene()
+    isSceneLoading = true
```
nchamo, line 39: Why was this change necessary? How does this work now?

nchamo (approved):
Looks great!

I love the idea of creating the concept of `SceneRuntime`

kernel/packages/entryPoints/preview.ts
menduz: Setting the flag was moved inside the function.

Before it was outside, right before of the function call, a dozen lines below.

```diff
-isSceneLoading = true
 loadScene()
```

Issue #1390, opened 2020-10-06
feature: support multiple UVs


AjimenezDCL:
This will solve _CyberMike_'s issue in #179

PR #1405, opened 2020-10-08
feat: uv multichannel support

AjimenezDCL:
#1390 
UV multichannel support for GLTF. 

If you test the preview dont forget to use the `NO_ASSET_BUNDLES` flag so the GLTFImporter does it's magic, but bear in mind that it's highly unlikely any object in the world is currently using the secondary UV channel.

Documentation: Due to the limited amout of TEXCOORDS and given that our shader/Unity use most of them we are facing some limitations. Currently we support 2 UV channels for the Albedo, Normal, Metallic and Emisive maps. Any other texture (Alpha Texture or Ambient Oclussion) will fallback to the UV selected by the Albedo.

We could increase the ammount of channels to 4 or 6 if the need arises but giving support to more textures is going to be tricky


AjimenezDCL:
~WARNING (mostly for my future self).~

~This feature is branching from PR: #1388 to avoid conflicts since both PRs modify the similar parts of the code. So once said PR is merged into master, we can change the base branch on this PR to `master` and process as normal.~

Done =)

pravusjif (changes requested):
Screen Shot 2020-10-08 at 09 11 55

There seems to be a bug affecting materials/textures

pravusjif (approved):

sandrade-dcl (approved):

unity-client/Assets/Shaders/MaterialTransition/LWRP/LitForwardPass.hlsl
```diff
@@ -92,7 +92,11 @@ Varyings LitPassVertex(Attributes input)
     half3 vertexLight = VertexLighting(vertexInput.positionWS, normalInput.normalWS);
     half fogFactor = ComputeFogFactor(vertexInput.positionCS.z);
 
-    output.uv = TRANSFORM_TEX(input.texcoord, _BaseMap);
+    float2 uvs[] = { TRANSFORM_TEX(input.texcoord, _BaseMap), TRANSFORM_TEX(input.texcoord1, _BaseMap)};
+    output.uvAlbedoNormal.xy = uvs[clamp(_BaseMapUVs, 0, 1)];
+    output.uvAlbedoNormal.zw = uvs[clamp(_NormalMapUVs, 0, 1)];
+    output.uvMetallicEmissive.xy = uvs[clamp(_MetallicMapUVs, 0, 1)];
+    output.uvMetallicEmissive.zw = uvs[clamp(_EmissiveMapUVs, 0, 1)];
```
BrianAmadori: we can't use `saturate` instead of `clamp` here?

AjimenezDCL: Isn't it the same? I thought `clamp` is just a wrap around `saturate`. We can use `saturate` if you like it more

BrianAmadori: I think `saturate` is more succinct. It's also supposed to be heavily optimized to deal with the 0,1 case.

AjimenezDCL: Fair enough, on it!

BrianAmadori (approved):
Looks good!

Issue #827, opened 2020-04-20
TextShape visible property not working


nearnshaw:
I can't make a TextShape invisible
If I set visible = false, it remains visible

This simple example shows a text that should be invisible, but can be seen:

```
const myEntity2 = new Entity()
const myText2 = new TextShape('Goodbye World!')
myText2.visible = false
myEntity2.addComponent(myText2)
myEntity2.addComponent(new Transform({ position: new Vector3(8, 1, 8) }))
engine.addEntity(myEntity2)
```

[https://app.zenhub.com/files/220994476/dabe984b-e9d0-41b7-803f-5a2c3b27e5d0/download](https://app.zenhub.com/files/220994476/dabe984b-e9d0-41b7-803f-5a2c3b27e5d0/download)

marcosnc:
@nchamo Can we fix this easily?

nchamo:
Hey @marcosnc !

So I just a little digging and the `visible` property is sent to Unity. The problem is that Unity doesn't know that the property exists. The json received is deserialized into a [class](https://github.com/decentraland/explorer/blob/master/unity-client/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs#L11) that doesn't have that propery.

That means that either:
1. The SDK is exposing a property that shouldn't be exposed, or
2. The client is incorrectly unaware of such property. 

Whatever we (I don't know who should make the decision) decide, be might consider it an SDK bug, or a client bug.

marcosnc:
I think we need to support the `visible` property, so it needs to be taken into account by Unity. @BrianAmadori what's the effort to process the property accordingly?

AjimenezDCL:
The problem is `TextShape` in Unity not extending `BaseShape` where this property exists. This is problematic for various reasons:
- `TextShape` currently is a `BaseComponent` in Unity, that means that changing the parent class to `BaseShape` (which is a Disposable component) can get really messy.
- Another side effect of making `TextShape` a `BaseShape` are the properties `withCollision` and `isPointerBlocker`. After discussing it with Nico, these two are enabled by default in kernel, which means that the current deployed scenes using `TextShape` has these properties already set by default to `true`. If we implement said properties in Unity we might be introducing a gamebreaking change.

So from now on `TextShape` wont have these `withCollision` `isPointerBlocker` 


On a side note, this ticket will also fix #1392

PR #1406, opened 2020-10-08
fix: TextShape now implements visible and opacity correctly

### LINKED ISSUE #1392: TextShape opacity property doesn't work
The TextShape component has an "opacity" property that should let you make a text semi-transparent. However, if you change that value nothing happens, and there is no other way to make a text semi-transparent.

The following code produces a text that is still 100% opaque
```ts
  const message = new Entity();
  message.addComponent(new TextShape("Test");
  message.addComponent(
    new Transform({ position: new Vector3(8, 1, 8) })
  );
  message.getComponent(TextShape).opacity = 0.5;
engine.addEntity(message)
```

AjimenezDCL:
#827

The problem is `TextShape` in Unity not extending `BaseShape` where this property exists. This is problematic for various reasons:
- `TextShape` currently is a `BaseComponent` in Unity, that means that changing the parent class to `BaseShape` (which is a Disposable component) can get really messy.
- Another side effect of making `TextShape` a `BaseShape` are the properties `withCollision` and `isPointerBlocker`. After discussing it with Nico, these two are enabled by default in kernel, which means that the current deployed scenes using `TextShape` has these properties already set by default to `true`. If we implement said properties in Unity we might be introducing a gamebreaking change.

On a side note, this ticket will also fix #1392

### DIFF
--- a/kernel/packages/decentraland-ecs/src/decentraland/Components.ts
+++ b/kernel/packages/decentraland-ecs/src/decentraland/Components.ts
@@ -640,7 +640,7 @@ export class TextShape extends ObservableComponent {
   billboard: boolean = false
 
   @ObservableComponent.field
-  visible: boolean = false
+  visible: boolean = true
 
   constructor(value?: string) {
     super()
--- a/kernel/packages/decentraland-ecs/src/decentraland/Components.ts
+++ b/kernel/packages/decentraland-ecs/src/decentraland/Components.ts
@@ -557,7 +557,7 @@ export enum Fonts {
  * @public
  */
 @Component('engine.text', CLASS_ID.TEXT_SHAPE)
-export class TextShape extends Shape {
+export class TextShape extends ObservableComponent {
   @ObservableComponent.field
   outlineWidth: number = 0
 
@@ -638,6 +638,9 @@ export class TextShape extends Shape {
 
   @ObservableComponent.field
   billboard: boolean = false
+
+  @ObservableComponent.field
+  visible: boolean = false
 
   constructor(value?: string) {
     super()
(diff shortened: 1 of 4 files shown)


BrianAmadori (approved):
Looks good!

unity-client/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
```diff
@@ -86,7 +88,7 @@ public static IEnumerator ApplyModelChanges(ParcelScene scene, TMP_Text text, Mo
 
             text.text = model.value;
 
-            text.color = new Color(model.color.r, model.color.g, model.color.b, model.color.a);
+            text.color = new Color(model.color.r, model.color.g, model.color.b, model.visible ? model.opacity : 0);
```
marcosnc, line 91: Why are we using `model.opacity` now instead of `model.color.a`?

kernel/packages/decentraland-ecs/src/decentraland/Components.ts
```diff
@@ -639,6 +639,9 @@ export class TextShape extends Shape {
   @ObservableComponent.field
   billboard: boolean = false
 
+  @ObservableComponent.field
+  visible: boolean = false
```
nchamo: Why is the default `false`, if it was previously `true`?

unity-client/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
AjimenezDCL: In `kernel` we are combining `Color3 color` and `float opacity` but Unity has no implementation of Color3, all of them has alpha (we could implement our own Color3, but it's a bit overkill) . So basically color.a does nothing because the value is defaulted.

We have two ways to solve this:
1) Remove opacity and use a Color4
2) Maintain opacity and use it in Unity instead of color.a.

After talking with Nico we decided that the second one was less disruptive for the creators.

kernel/packages/decentraland-ecs/src/decentraland/Components.ts
AjimenezDCL: Because I am stupid =D, nice catch

nchamo (approved):
Great work!

PR #1407, opened 2020-10-08
fix: Fixed a couple of bugs & tuned buffers

pablitar:
To improve voice chat quality, tuned the size of the frame and the jitter buffer.

Also, fixed a couple of buffers related to frame filling & frame ordering


caminoezequiel (approved):
👌

menduz:
I'm wondering if it is correct to have the generated `kernel/static/voice-chat-codec/worker.js` commited. Is that intentional?

pablitar:
@menduz hehehe. Is part of the dogma. I actually haven't tested not commiting it, because when I made the worker I based the changes in the gif worker which had the js commited.

I'll test ignoring the compiled js. I believe it should work.

PR #1415, opened 2020-10-10
feat: better green blockers

BrianAmadori:
Included in this PR:

Visual related changes:
- New, more configurable hologram shader
- Reduced blockers draw distance so they are more subtle
- Now blockers don't match the scene height limit, they have a limited height (i.e.: like fences)
- Now blockers lifecycle is handled with a simple fadeIn/fadeOut animation.

Code design changes:
- `BlockerHandler` is now called `BlockerInstanceHandler`. His main responsibility is being a composite of blockers, and handling the show/hide.
- `WorldBlockersController` is in charge of determining which blockers should be shown or not, depending on the scenes loading state. Before, this responsibility was segregated between this one and `BlockerHandler`.
- Added `BlockerAnimationHandler`. This class has the sole responsibility of handling the fadeIn/out blocker animation.

Misc changes:
- Blocker prefab is now taken from the render profile, so we can have different types of blockers depending on day/night, etc.


unity-client/Assets/Rendering/Profiles/RenderProfileWorld.cs
```diff
@@ -14,6 +14,7 @@
 [CreateAssetMenu(menuName = "DCL/Rendering/Create World Profile", fileName = "RenderProfileWorld", order = 0)]
 public class RenderProfileWorld : ScriptableObject
 {
+    [SerializeField] public GameObject loadingBlockerPrefab;
```
sandrade-dcl: If you don't need the variable as public:
```suggestion
    [SerializeField] private GameObject loadingBlockerPrefab;
```
If you need the variable as public:
```suggestion
    public GameObject loadingBlockerPrefab;
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BlockerController/BlockerAnimationHandler.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        void FadeIn(GameObject go);
+        void FadeOut(GameObject go, System.Action OnFinish);
+    }
+
+    /// <summary>
+    /// This class is in charge of handling the FadeIn/FadeOut animation
+    /// of blockers.
+    /// </summary>
+    public class BlockerAnimationHandler : IBlockerAnimationHandler
```
sandrade-dcl, line 17: I have seen that this class could be used for any game object with a `Renderer`. Maybe we could separate it from the blockers folders, rename it and set it as a general tool for any time where we need to apply fade in/out to any other game object? What do you think?

sandrade-dcl (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BlockerController/Tests/BlockerHandlerCan.cs
```diff
@@ -0,0 +1,119 @@
...
+        sceneHandler.GetAllLoadedScenesCoords().Returns(allLoadedParcelCoords);
+
+        blockerAnimationHandler = Substitute.For<IBlockerAnimationHandler>();
+
+        //NOTE(Brian): Call OnFinish() when blockerAnimationHandler.FadeOut is called. 
+        blockerAnimationHandler.FadeOut(Arg.Any<GameObject>(), Arg.Invoke());
+
+        blockerInstanceHandler = new BlockerInstanceHandler(new DCLCharacterPosition(), blockerAnimationHandler);
+
+        blockersParent = new GameObject();
```
AjimenezDCL, line 40: You can use `TestBase.CreateTestGameObject(name)` and it will create an object that automatically will be destroyed so you dont have to teardown them. It's a matter of opinion though, if you prefer it explicitly in the test itself, that's fine.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BlockerController/Tests/BlockerHandlerCan.cs
```diff
@@ -0,0 +1,119 @@
...
+        blockerInstanceHandler.ShowBlocker(new Vector2Int(0, 0), true);
+        blockerInstanceHandler.HideBlocker(new Vector2Int(0, 0), true);
+        blockerAnimationHandler.DidNotReceiveWithAnyArgs().FadeOut(default, default);
+        Assert.AreEqual(0, blockerInstanceHandler.GetBlockers().Count);
+    }
+
+    [Test]
+    public void SetParent()
+    {
+        // Prepare
```
AjimenezDCL: I can relate to this so much hahaha, the convention is `Arrange` instead of `Prepare` (ArrangeActAssert). I always forget what the first A stands for.

AjimenezDCL (approved):
Great job!

I love the new tests, the smart use of DI and these new TestUtils to check messages to kernel.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BlockerController/BlockerAnimationHandler.cs
BrianAmadori: Mmmh the issue I see with this approach is that this class is expecting the Blocker prefab arrangement/shader, so we can end up re-using it and then when we need to modify the code to change blocker-specific stuff, the re-used instances will break. For instance, we may add a new specific property to the blocker shader and we might want to animate it here.

I could've used animations for this, but decided for this lightweight approach because the blockers are widely used.

I propose to keep this as it is and if in the future we need to apply FadeIn/FadeOut to another non-ui object we can make this more generic, or maybe reuse the interface and change its name.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BlockerController/Tests/BlockerHandlerCan.cs
BrianAmadori: Interesting, actually I didn't know we had that method.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BlockerController/BlockerAnimationHandler.cs
sandrade-dcl: Ok! I agree with that!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BlockerController/Tests/BlockerHandlerCan.cs
BrianAmadori: I can't use this because this class doesn't inherit from `TestsBase`. Going to resolve this and maybe we can put this later un `TestHelpers` somehow.

unity-client/Assets/Rendering/Profiles/RenderProfileWorld.cs
BrianAmadori: Solved making the field public as we need it for initializing the blockers pool.

sandrade-dcl (approved):

PR #1416, opened 2020-10-10
chore: Updating peer library to fix a small issue with relay suspension

pablitar:


kernel/package.json
```diff
@@ -100,7 +100,7 @@
     "dcl-catalyst-client": "^2.1.5",
     "dcl-crypto": "^2.2.0",
     "dcl-social-client": "^1.3.10",
-    "decentraland-katalyst-peer": "0.2.15",
+    "decentraland-katalyst-peer": "0.2.16",
```
nchamo, line 103: Damn, I hate that we still have `katalyst` on some places

nchamo (approved):

kernel/package.json
pablitar: Haha. I don't know who has permissions, but we need to create the `decentraland-catalyst-peer` package on NPM and then we can change the package name in Catalyst.

nchamo: We now have the `@dcl` organization in npm. Maybe we could move the package there? Not necessary for this merge, of course. Just an idea for the future

pablitar: In that case, we could just call this `catalyst-peer` and then we could have this as `@dcl/catalyst-peer`. That would be nice.

nchamo: Could I ask you to create an issue on the catalyst repo? So don't forget 😄

pablitar: https://wiki.uqbar.org/wiki/articles/lombardizacion.html

Issue #1419, opened 2020-10-12
Disable entitiy's sounds from AudioStreaming when it leaves the scene


AjimenezDCL:
~It follows the same principle as in #1276 but to avoid creating monolithics PRs this one has a different ticket.~ 

Nevermind it could be solved in #1276 while keeping the PR in a healthy size.

Issue #1276, opened 2020-09-08
Disable entity's sounds, when it leaves the scene


nchamo:
Right now, when an entity goes outside of the scene's boundaries, its mesh is turned off. This means that it is no longer visible. However, all sounds continue to play. This is particularly problematic with the new `Attachable` features, since it's possible to attach an object with sound to the player, making it follow the user everywhere.

Possible solution: make the `SceneBoundariesChecker` disable all sounds, in addition to the mesh.

nchamo:
As of https://github.com/decentraland/explorer/issues/1352, if the player goes outside the scene, the sounds will stop. However, if the player is still on the scene and the entity with the sound attached goes out of bounds, it will continue to play.

PR #1420, opened 2020-10-12
chore: now AudioSource gets disabled when out of Scene Bounds

AjimenezDCL:
#1276 

Now AudioSource is part of the scenery boundaries checks. If the entity has no mesh a simple check against the entity position is made.

~The case for AudioStreaming/VideoStreaming is different since it's done in `kernel` and it will be tackled-down in #1419~ Nevermind, it will be done here.


sandrade-dcl (approved):

AjimenezDCL:
Tests for DCLAudioStream won't be included here. The tests (along the refactor needed to make them doable) will be tackled down here #1421

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneBoundariesController/SceneBoundariesChecker.cs
```diff
@@ -202,6 +210,15 @@ protected virtual void UpdateEntityCollidersValidState(DecentralandEntity entity
             }
         }
 
+        protected virtual void UpdateComponents(DecentralandEntity entity, bool isInsideBoundaries)
+        {
+            IOutOfSceneBoundariesHandler[] audioSources = entity.gameObject.GetComponentsInChildren<IOutOfSceneBoundariesHandler>();
```
```suggestion
            IOutOfSceneBoundariesHandler[] components = entity.gameObject.GetComponentsInChildren<IOutOfSceneBoundariesHandler>();
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -566,6 +566,11 @@ public BaseComponent EntityComponentCreateOrUpdate(string entityId, CLASS_ID_COM
                 return null;
             }
 
+            if (classId == CLASS_ID_COMPONENT.AUDIO_SOURCE || classId == CLASS_ID_COMPONENT.AUDIO_STREAM)
+            {
+                SceneController.i.boundariesChecker?.AddEntityToBeChecked(entity);
+            }
```
BrianAmadori: Not for this PR, but we should be configuring these aspect of components on some sort of manifest to avoid all the ifs like this. Take into account we already have this anti-pattern in messaging system -- the conditional voiding the lossy messages depending on message type.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioStream.cs
```diff
@@ -117,5 +118,20 @@ private void StartStreaming()
...
+            if (!isPlaying)
+                return;
+
+            if (enable)
+            {
+                StartStreaming();
+            }
+            else
+            {
+                Interface.WebInterface.SendAudioStreamEvent(model.url, true, 0);
```
BrianAmadori, line 134: This is supposed to stop the streaming? A comment or wrapper method would be useful here.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneBoundariesController/SceneBoundariesChecker.cs
```diff
@@ -6,6 +6,11 @@
 
 namespace DCL.Controllers
 {
+    public interface IOutOfSceneBoundariesHandler
+    {
+        void UpdateOutOfBoundariesState(bool enable);
+    }
```
BrianAmadori, line 12: ❤️ 🏗️

BrianAmadori (approved):
Looks good!. Left some small comments.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
AjimenezDCL: I think we can use the same interface and simply check if the `Monobehaviour` retrieved (either by the componentFactory or the entity components list is `IOutOfSceneBoundariesHandler`.

I'm gonna give it a call in this PR and if it gets messy I will leave it and create a ticket.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioStream.cs
AjimenezDCL: Yeah, the last param is the volume, I will add a comment!

PR #1450, opened 2020-10-16
chore: update canvas mask bottom limit

pravusjif:
Updated scenes ui canvas:
- Replaced mask-middle-gameobject with a new canvas renderer with no mask and the bottom limit lifted according to the new taskbar

### **BEFORE**
Screen Shot 2020-10-19 at 13 47 57
Screen Shot 2020-10-19 at 13 49 21
Screen Shot 2020-10-19 at 13 48 22

### **AFTER**
Screen Shot 2020-10-16 at 16 48 40
Screen Shot 2020-10-16 at 16 48 07
Screen Shot 2020-10-16 at 16 49 02

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Components/UI/UIContainer/UISizeFitter.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Components/UI/UIContainer/UISizeFitter.cs
@@ -21,11 +21,10 @@ public class UISizeFitter : MonoBehaviour
     {
         if (canvasChildHookRT == null)
         {
-            RectMask2D constrainedPanelMask = GetComponentInParent<RectMask2D>();
-
-            if(constrainedPanelMask != null)
+            CanvasRenderer resizedCanvasPanel = GetComponentInParent<CanvasRenderer>();
+            if (resizedCanvasPanel != null)
             {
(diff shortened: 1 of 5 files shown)


sandrade-dcl (approved):
Reviewed and tested... Good work!

BrianAmadori (approved):
Results look pretty good. 

To confirm I understand the solution, what you are doing is:
- Raise the canvas UI so its rendered above the taskbar
- Scale it down preserving the aspect ratio so the top part isn't occluded by the profile stuff

I am getting it right?

pravusjif:
> Results look pretty good.
> 
> To confirm I understand the solution, what you are doing is:
> 
> * Raise the canvas UI so its rendered above the taskbar
> * Scale it down preserving the aspect ratio so the top part isn't occluded by the profile stuff
> 
> I am getting it right?

Just to clarify it would be:
* Raise the UI Canvas **Bottom** to be just above the taskbar
* Removed mask component (otherwise the UI objects positioned relative to the bottom, that are getting pushed a bit up, because of the previous bulletpoint, would get cut by the top mask), also all the Decentraland UI renders on top of scenes UI so there's no problem there.
* The UI Canvas **Top** is still 10% below the true sceen top, to preserve compatibility with already-deployed scenes (and to avoid putting top-relative ui rendering below DCL UI now) <- this is probably what you mean by your 2nd point

PR #1453, opened 2020-10-16
chore: enable exception logging

BrianAmadori:
# What? 
...

# Why? 
...


BrianAmadori:
Closing it because we still need the symbol mapping and development mode on to make it useful.

Issue #1433, opened 2020-10-14
Implement configurable MOTD for events


BrianAmadori:
This task includes:

- [X] Adding kernel <> unity pipeline for sending over the MOTD data. This should be already in place in `unity.ts:149` in `feat/halloween-2020`
- [x] Creating a PR that comes from `feat/halloween-2020`
- [x] Setting up the MOTD (Message of the day) HUD to load the custom image, show the custom text, etc.
- [x] Syncing up with @RominaMarchetti for visual polish

PR #1456, opened 2020-10-19
feat: configurable MOTD

AjimenezDCL:
#1433 
Now MOTD (Message of the Day) is configurable. 

This PR ships the feature with placeholder art. That will be taken down later.


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WelcomeHUD/WelcomeHUDController.cs
```diff
@@ -1,20 +1,39 @@
+using System;
 using DCL.Helpers;
 using DCL.Interface;
 using UnityEngine;
 
+[Serializable]
+public class MessageOfTheDayConfig
```
sandrade-dcl, line 10: Lately we are adding an overview description for all new classes we implement (also for the public methods that we consider necessary) in order to keep a basic documentation of each one. It would be fine to add it to this new HUD.

sandrade-dcl (changes requested):
Nice work! I left only a comment.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WelcomeHUD/WelcomeHUDController.cs
AjimenezDCL: Oh, I had no idea :o

menduz (approved):

sandrade-dcl (approved):
Thanks for the docu! I would also add some documentation to the `WelcomeHUDController` class, simply giving a basic explanation about how the MOTD feature works.

BrianAmadori (approved):
Looks good!

PR #1458, opened 2020-10-19
feat: add mask for tinting mouths in Avatar

AjimenezDCL:
#1432 

The shader has been updated to receive a mask where the tint should be performed.

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/Shaders/UnlitCutoutTinted.shader
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/Shaders/UnlitCutoutTinted.shader
@@ -1,6 +1,7 @@
 ﻿Shader "DCL/Unlit Cutout Tinted" {
 Properties {
     _BaseMap ("Base (RGB) Trans (A)", 2D) = "white" {}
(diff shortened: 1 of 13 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/Shaders/UnlitCutoutTinted.shader
```diff
@@ -50,7 +52,11 @@ SubShader {
  
             fixed4 frag (v2f i) : SV_Target
             {
-                fixed4 col = tex2D(_BaseMap, i.texcoord) * _BaseColor;
+                fixed4 col = tex2D(_BaseMap, i.texcoord);
+                fixed4 tintMask = tex2D(_TintMask, i.texcoord);
+
+                col *= lerp(float4(1,1,1,1), _BaseColor, tintMask.r);
```
BrianAmadori: In `EyesShader.shader` I see that the lerp function does `( 1.0 - tex2D( _IrisMask, uv_IrisMask ).r )` I think we should invert this one too so the pipeline keeps coherency.

pravusjif (approved):
LGTM, tested the build a bit and keep seeing other avatars faces with no problems

BrianAmadori (approved):
LGTM

PR #1462, opened 2020-10-19
fix: Don't send all avatar events

pablitar:
Changed Avatar Events tracking so it only tracks the specified events and not every new event added


caminoezequiel (approved):

nchamo (approved):

pravusjif (approved):
Also tested the build a bit with no problems.

Issue #1441, opened 2020-10-15
Fix misc nullrefs detected at Rarible event


pravusjif:
1. `Assets/Scripts/MainScripts/DCL/Components/Materials/PBRMaterial.cs:276`
```
MissingReferenceException: The object of type 'Material' has been destroyed but you are still trying to access it.
Your script should either check if it is null or you should not destroy the object.
UnityEngine.Material.SetTexture (System.Int32 nameID, UnityEngine.Texture value) (at /Users/builduser/buildslave/unity/build/Runtime/Export/Shaders/Shader.cs:503)
DCL.Components.PBRMaterial+<>c__DisplayClass23_0.<SetMaterialTexture>b__0 (DCL.DCLTexture fetchedDCLTexture) (at Assets/Scripts/MainScripts/DCL/Components/Materials/PBRMaterial.cs:276)
DCL.DCLTexture+<FetchTextureComponent>d__9.MoveNext () (at Assets/Scripts/MainScripts/DCL/Components/Textures/DCLTexture.cs:63)
UnityEngine.SetupCoroutine.InvokeMoveNext (System.Collections.IEnumerator enumerator, System.IntPtr returnValueAddress) (at /Users/builduser/buildslave/unity/build/Runtime/Export/Scripting/Coroutines.cs:17)
```

2. `Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetBundlesLoader.cs:165`
```
NullReferenceException: Object reference not set to an instance of an object
DCL.AssetBundlesLoader+<LoadAssetBundle>d__20.MoveNext () (at Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetBundlesLoader.cs:165)
UnityEngine.SetupCoroutine.InvokeMoveNext (System.Collections.IEnumerator enumerator, System.IntPtr returnValueAddress) (at /Users/builduser/buildslave/unity/build/Runtime/Export/Scripting/Coroutines.cs:17)
```

3. `Assets/Scripts/MainScripts/DCL/Components/Billboard/Billboard.cs:36`
```
NullReferenceException: Object reference not set to an instance of an object
DCL.Billboard+<ApplyChanges>d__6.MoveNext () (at Assets/Scripts/MainScripts/DCL/Components/Billboard/Billboard.cs:36)
[...]

PR #1461, opened 2020-10-19
fix: nullcheck in avatar

AjimenezDCL:
#1440 
AvatarRenderer system is throwing some uncatched null references. Let's avoid them until we discover the underlying cause.

PR #1463, opened 2020-10-19
fix: added nullsafe checks in misc files

AjimenezDCL:
#1441 
Some nullsafe checks with debug info in some files.

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Components/Billboard/Billboard.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Components/Billboard/Billboard.cs
@@ -33,6 +33,12 @@ namespace DCL
             {
                 //NOTE(Zak): We have to wait one frame because if not the entity will be null. (I'm Brian, but Zak wrote the code so read this in his voice)
                 yield return null;
+
+                if (entity == null)
+                {
+                    Debug.LogWarning("It seems skipping a frame didnt work, entity is still null");
+                    yield break;
+                }
                 entityTransform = entity.gameObject.transform;
             }
         }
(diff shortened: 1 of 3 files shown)


pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/Billboard/Billboard.cs
```diff
@@ -33,6 +33,12 @@ public override IEnumerator ApplyChanges(string newJson)
             {
                 //NOTE(Zak): We have to wait one frame because if not the entity will be null. (I'm Brian, but Zak wrote the code so read this in his voice)
                 yield return null;
+
+                if (entity == null)
+                {
+                    Debug.LogWarning("It seems skipping a frame didnt work, entity is still null");
+                    yield break;
+                }
```
pbosio, line 41: maybe, it could be a good idea to use a `while` here and retry like for 5 frames? just in case? what do you think?

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/BodyShapeController.cs
```diff
@@ -66,6 +72,12 @@ public void SetupEyes(Material material, Texture texture, Texture mask, Color co
 
     public void SetupEyebrows(Material material, Texture texture, Color color)
     {
+        if (assetContainer?.transform == null)
+        {
+            Debug.LogWarning("Tried to setup eyes when the asset not ready");
+            return;
+        }
```
pbosio: remember not to overwrite the changes in PR #1461 with these :P

pbosio (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/Billboard/Billboard.cs
AjimenezDCL: I would attack that stuff in the #1467 and just leave this PR for the null-checks.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/BodyShapeController.cs
AjimenezDCL: This shouldn't be here. We had a bit of a fiesta with the branches while doing tests hahaha. I force pushed putting the branch up to master, really nice catch!

PR #1465, opened 2020-10-19
fix: Block feature key triggers during the tutorial

sandrade-dcl:
Block feature key triggers during the tutorial.

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
@@ -74,6 +74,10 @@ public class InputController : MonoBehaviour
         for (var i = 0; i < triggerTimeActions.Length; i++)
         {
             var action = triggerTimeActions[i];
+
+            if (action.blockTrigger != null && action.blockTrigger)
+                continue;
+
             switch (action.GetDCLAction())
             {
                 case DCLAction_Trigger.CameraChange:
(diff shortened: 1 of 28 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
```diff
@@ -84,35 +84,38 @@ public void Update_Trigger()
                                        InputProcessor.Modifier.FocusNotInInput);
                     break;
                 case DCLAction_Trigger.ToggleNavMap:
-                    if (allUIHidden) break;
+                    if (allUIHidden || CommonScriptableObjects.featureKeyTriggersBlocked) break;
```
BrianAmadori: Let me suggest an alternative, less hacky approach:

Add a new `BoolVariable` field to the `ActionTrigger_Input` SO, so it gets enabled/disabled according to a custom `BoolVariable` object.

Just query that field value at the `for` start, like this:
```
if ( action.boolEnableObject != null && !action.boolEnableObject.Get() )
    continue;
```
With this in place, we can safely remove all the hardcoded conditions and make this feature data-driven.

sandrade-dcl: This a very elegant solution for this! Thanks for the suggestion :) I have pushed the change, please take a look!

BrianAmadori (approved):

pbosio (approved):

pravusjif (changes requested):
the blocked inputs are not getting released if you skip the tutorial

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
```diff
@@ -74,6 +74,10 @@ public void Update_Trigger()
         for (var i = 0; i < triggerTimeActions.Length; i++)
         {
             var action = triggerTimeActions[i];
+
+            if (action.blockTrigger != null && action.blockTrigger)
```
pravusjif: Just in case I would suggest we always use the `Get()` method of our scriptable object variables to avoid confusion on Unity's "not null" check, I'd change that line to:
`if (action.blockTrigger != null && action.blockTrigger.Get())`

sandrade-dcl: Thanks for the suggestion, you're right! I have done the change in the last commit.

sandrade-dcl:
> the blocked inputs are not getting released if you skip the tutorial

Oh! Good catch!! Fixed on the last commit :) Thanks!

pravusjif (approved):

PR #1466, opened 2020-10-19
chore: Halloween Pattern UI atlas updated

RominaMarchetti:
The new pattern was not visible on the private conversation and avatar editor background.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,7 +91,7 @@ jobs:
       - run:
           name: Get the hash of source files
           command: |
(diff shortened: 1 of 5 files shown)


pravusjif (approved):

BrianAmadori (approved):
Looks good! In the future we should aim to make this kind of changes toggleable via the new Render Profile feature. In any case that feature is only on the halloween branch for now, so let's talk about it for future events.

RominaMarchetti:
> Looks good! In the future we should aim to make this kind of changes toggleable via the new Render Profile feature. In any case that feature is only on the halloween branch for now, so let's talk about it for future events.

Great idea!

PR #1469, opened 2020-10-19
feat: users around player list. mute/unmute voice chat

pbosio:
https://explorer.decentraland.zone/branch/feat/users-around-list/index.html?ENV=org&VOICE_CHAT_ENABLED

picture for reference:
Screen Shot 2020-10-23 at 17 14 09


kernel/packages/shared/comms/actions.ts
```diff
@@ -29,3 +29,7 @@ export type VoiceRecordingUpdate = ReturnType<typeof voiceRecordingUpdate>
 export const SET_VOICE_VOLUME = 'Set Voice Volume'
 export const setVoiceVolume = (volume: number) => action(SET_VOICE_VOLUME, { volume })
 export type SetVoiceVolume = ReturnType<typeof setVoiceVolume>
+
+export const SET_VOICE_MUTE = 'Set Voice Mute'
+export const setVoiceMute = (mute: boolean) => action(SET_VOICE_MUTE, { mute })
```
pablitar, line 34: Could it be that this action is not dispatched anywhere? Is there something missing?

pbosio: Is not used... I added it in a first approach and then decided to keep it just in case, but I can remove it if you wish

sandrade-dcl (changes requested):
I think I have found a little issue:

1. Go to any place where there are several people.
2. Activate the MUTE ALL option --> Notice all the list below is marked as muted (with the X), this is OK.
3. Now teleport to other place where there are other people --> Notice the MUTE ALL option is turned off now but the list of users below continue marked as muted (with the X). I think this is not OK.

For me, the correct behaviour in the step 3 should be that the MUTE ALL option should continue activated and all the users list below marked as muted.

image

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Minimap/MinimapHUDController.cs
```diff
@@ -93,4 +93,10 @@ public void SetVisibility(bool visible)
     {
         view.SetVisibility(visible);
     }
+
+    public void AddUsersAroundIndicator(UsersAroundListHUDController controller)
```
sandrade-dcl, line 102: In order to follow the good practice of improve our documentation, from some time now we try to add a basic documentation to all the new classes and public methods (not including tests). Could you add a basic description as header about what this function do? It will help to anyone who need to call it, seeing the description throught the IDE intellisense.

It would be fine if you could apply this to all new classes / public methods you have created in this PR (at least in all those places where you think it would be useful or has sense).

sandrade-dcl (changes requested):

pbosio:
> I think I have found a little issue:
> 
>     1. Go to any place where there are several people.
> 
>     2. Activate the MUTE ALL option --> Notice all the list below is marked as muted (with the X), this is OK.
> 
>     3. Now teleport to other place where there are other people --> Notice the MUTE ALL option is turned off now but the list of users below continue marked as muted (with the X). I think this is not OK.
> 
> 
> For me, the correct behaviour in the step 3 should be that the MUTE ALL option should continue activated and all the users list below marked as muted.
> 
> image

Good catch Santi thanks. I didn't notice that 😁
Seems to be an issue with the switch animator resetting after the GO is disabled

pravusjif:
> > I think I have found a little issue:
> > ```
> > 1. Go to any place where there are several people.
> > 
> > 2. Activate the MUTE ALL option --> Notice all the list below is marked as muted (with the X), this is OK.
> > 
> > 3. Now teleport to other place where there are other people --> Notice the MUTE ALL option is turned off now but the list of users below continue marked as muted (with the X). I think this is not OK.
> > ```
> > 
> > 
> > For me, the correct behaviour in the step 3 should be that the MUTE ALL option should continue activated and all the users list below marked as muted.
> > image
> 
> Good catch Santi thanks. I didn't notice that 😁
> Seems to be an issue with the switch animator resetting after the GO is disabled

Yeah mecanim animators always reset when the GO is deactivated and re-activated =S

pravusjif (changes requested):
The "mute all" is actually a "mute state toggle" it doesn't make sure the users end up actually muted, it happened to me that I had 2 users muted, 2 not muted, and when I toggled the "mute all" it just reversed those 4 users (the ones originally muted got unmuted and the ones originally unmuted got muted)

pbosio:
> The "mute all" is actually a "mute state toggle" it doesn't make sure the users end up actually muted, it happened to me that I had 2 users muted, 2 not muted, and when I toggled the "mute all" it just reversed those 4 users (the ones originally muted got unmuted and the ones originally unmuted got muted)

this can happen in the following two scenarios:
1- you experiment the issue reported by Santi, where the toggle was not displayed OK because the animator was reset
2- there is an issue with the profile that Pablo is currently tackling that result in failing while trying to mute/unmute an user 
the first one I already fixed it in the latest commit.

pravusjif (approved):
My issue was fixed with the last commit, it works awsome now.

sandrade-dcl (approved):

pablitar (approved):
Great work!!

Issue #742, opened 2020-03-16
Some textures look worse in 6.4.7


nearnshaw:
A user reported that textures in his models used to look a lot better in 6.4.4 vs the current SDK version (6.4.7)

They used to look just like in his editing tools, but now they look "blotchy".

![image.png](https://images.zenhubusercontent.com/5d9b940e491c060001c8647b/5f9df71c-e54a-461c-8165-d641cc6876bc)

[https://app.zenhub.com/files/220994476/72ce19d1-c6b9-4b15-b8f8-84215d8119a7/download](https://app.zenhub.com/files/220994476/72ce19d1-c6b9-4b15-b8f8-84215d8119a7/download)

Did we maybe just change the default texture resolution in preview?

nearnshaw:
could this be related to this comment by @dzsunyec?

it seems setting the texture filtering mode to 0:NEAREST still doesn't show up after deployment

dzsunyec:
I am also experiencing the blotchiness/banding issue in 6.4.7 (see SDK preview comparison image below), but I'm assuming it has to do with texture bit depth and probably separate issue from the filtering. 

Setting the filtering mode to NEAREST works with both the SDK Materials and Blender exported materials **in preview**, but not working in the deployed scene with neither 6.4.4 or 6.4.7.

![texture_banding](https://user-images.githubusercontent.com/3963713/80274605-ef5c8800-86db-11ea-9a60-4720fca0c6b7.png)

BrianAmadori:
We think this issue is related to texture compression. We are going to check if there are better compression alternatives but removing compression is not an option for now.

BrianAmadori:
Just saw that we are using `Compress(false)` (low quality) on `GLTFSceneImporter`. Maybe this is related to that code.

AjimenezDCL:
HQ Compression: ![textures_compression_hq](https://user-images.githubusercontent.com/50231608/96886945-5c32be00-1484-11eb-9a51-523212267a03.png)
Normal Compression: ![textures_compression_normal](https://user-images.githubusercontent.com/50231608/96886949-5ccb5480-1484-11eb-8076-b93e8cb6af66.png)
No Compression: ![textures_no_compression](https://user-images.githubusercontent.com/50231608/96886952-5d63eb00-1484-11eb-9120-82fcc4927e4b.png)

I just tested with different settings for compresion and it's not helping that much. So we are closing the ticket as a not-an-issue. Sadly removing compression is not an option currently.

[...]

PR #1485, opened 2020-10-22
fix: set texture compression to HighQuality

AjimenezDCL:
#742 

This issue is likely caused by our new compression in textures. If that's the case we will have to decide whether or not the quality improvement worth the extra memory.

### DIFF
--- a/unity-client/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
+++ b/unity-client/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
@@ -750,7 +750,7 @@ namespace UnityGLTF
 
 #if !UNITY_EDITOR
             //NOTE(Brian): This breaks importing in editor mode
-            texture.Compress(false);
+            texture.Compress(true);
 #endif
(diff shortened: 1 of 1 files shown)


unity-client/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -750,7 +750,7 @@ protected virtual IEnumerator ConstructUnityTexture(byte[] buffer, bool markGpuO
 
 #if !UNITY_EDITOR
             //NOTE(Brian): This breaks importing in editor mode
-            texture.Compress(false);
```
BrianAmadori, line 753: Note that this is only to fix the issue for GLTF models, AB models with keep the compression set by the converter. This said, I updated the converter compression to HQ in the last AB converter update.

AjimenezDCL: I forgot to mention the branch should be tested with `NO_ASSET_BUNDLES`. In any case this PR will be discarded since the ticket is gonna be marked as not-an-issue.

PR #1488, opened 2020-10-22
 fix: ui container rect borders and canvas

pravusjif:
During last week's merges (onboarding, etc.) some changes were corrupted/lost.

* Re-connected UIContainerRect child panel image to the missing sprite (radius 4)
* Fixed child hook recttransform fetching in UISizeFitter
* Removed "persistent scene" branching in UIScreenSpace as we no longer use that for the global UI


BrianAmadori (approved):
Tests are failing somehow but the changes look solid to me ✔️

sandrade-dcl (approved):

PR #1489, opened 2020-10-22
feat: allow SDK to trigger avatar emote

nchamo:
We are now adding to the SDK the ability to trigger predefined emotes on the current avatar. The available emotes are:
`'WAVE' | 'FIST_PUMP' | 'ROBOT' | 'RAISE_HAND' | 'CLAP' | 'MONEY' | 'KISS'`

We also took this oportunity to make the existent `RestrictedActionModule` consistent with the other apis. This is actually a breaking change, even though it's pretty small. Content creators will have to now add the following import to their scenes to continue using the `movePlayerTo` action:
```import { movePlayerTo } from '@decentraland/RestrictedActions'```

The same will be necessary for the new `triggerEmote` action:
```
import { triggerEmote, PredefinedEmote } from '@decentraland/RestrictedActions'
triggerEmote({ predefined: PredefinedEmote.KISS })
```

As said by the api name, the trigger emote action will also requiere permissions from the user. Since the permission system is not yet defined, content creators will need to add the `ALLOW_TO_TRIGGER_AVATAR_EMOTE` flag to the [required permissions property](https://docs.decentraland.org/development-guide/scene-metadata/#required-permissions) on the `scene.json`

### CODE REVIEW

File: kernel/packages/decentraland-ecs/types/@decentraland/RestrictedActions/index.d.ts [hunk 1]
```diff
@@ -0,0 +1,24 @@
...
+   *
+   * @param emote the emote to perform
+   */
+  export function triggerEmote(emote: Emote): Promise<void>
+
+  export type Emote = {
+    predefined: PredefinedEmote
+  }
+
+  export type PredefinedEmote = 'WAVE' | 'FIST_PUMP' | 'ROBOT' | 'RAISE_HAND' | 'CLAP' | 'MONEY' | 'KISS'
```
nchamo: This can't be an enum, because we are using RPC here, and it is not supported

kernel/packages/shared/apis/RestrictedActions.ts
```diff
@@ -76,4 +47,60 @@ export class RestrictedActionModule extends ExposableAPI implements IRestrictedA
...
+    // checks permissions
+    if (!this.hasPermission(Permission.ALLOW_TO_TRIGGER_AVATAR_EMOTE)) {
+      defaultLogger.error(`Permission "${Permission.ALLOW_TO_TRIGGER_AVATAR_EMOTE}" is required`)
+      return
+    }
+
+    unityInterface.TriggerSelfUserExpression(this.translatePredefinedEmoteToId(emote.predefined))
+  }
+
+  private translatePredefinedEmoteToId(emoteName: string): string {
```
nchamo, line 62: All theses methods were just moved around the file

kernel/packages/decentraland-ecs/types/@decentraland/RestrictedActions/index.d.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  /**
+   * trigger an emote on the current player
+   *
+   * @param emote the emote to perform
+   */
+  export function triggerEmote(emote: Emote): Promise<void>
+
+  export type Emote = {
+    predefined: PredefinedEmote
```
nchamo, line 20: In the future, this should be something like:
```
export type Emote = { predefined: PredefinedEmote } | { custom: string }
```

kernel/test/unit/RestrictedActions.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+        .once()
+        .withExactArgs('Permission "ALLOW_TO_TRIGGER_AVATAR_EMOTE" is required')
+
+      const module = new RestrictedActions(options)
+      await module.triggerEmote({ predefined: 'FIST_PUMP' })
+      sinon.verify()
+    })
+  })
+
+  describe('MovePlayerTo tests', () => {
```
nchamo, line 53: All the tests below were not added. This file was just renamed

caminoezequiel (approved):

kernel/packages/shared/apis/RestrictedActions.ts
```diff
@@ -76,4 +47,60 @@ export class RestrictedActionModule extends ExposableAPI implements IRestrictedA
...
+      case 'KISS': return 'kiss'
+    }
+    throw new Error(`Unrecognized emote name ${emoteName}`)
+  }
+
+  private getSceneData() {
+    return this.parcelIdentity.land.sceneJsonData
+  }
+
+  private hasPermission(permission: Permission) {
```
nchamo, line 66: All theses methods below were just moved around the file

pablitar (approved):
LGTM! Great work!

kernel/packages/shared/apis/RestrictedActions.ts
```diff
@@ -9,43 +9,14 @@ import { lastPlayerPosition } from '../world/positionThings'
...
-  ALLOW_TO_MOVE_PLAYER_INSIDE_SCENE = 'ALLOW_TO_MOVE_PLAYER_INSIDE_SCENE'
+  ALLOW_TO_MOVE_PLAYER_INSIDE_SCENE = 'ALLOW_TO_MOVE_PLAYER_INSIDE_SCENE',
+  ALLOW_TO_TRIGGER_AVATAR_EMOTE = 'ALLOW_TO_TRIGGER_AVATAR_EMOTE'
 }
 
-export interface IRestrictedActionModule {
-  movePlayerTo(newPosition: Vector3, cameraTarget?: Vector3): Promise<void>
-}
-
-@registerAPI('RestrictedActionModule')
```
menduz, line 19: Renaming `RestrictedActions` to `RestrictedActionModule` will break every scene currently using `movePlayerTo(position: PositionType, cameraTarget?: PositionType)` (moved to the module)

menduz (changes requested):

kernel/packages/shared/apis/RestrictedActions.ts
nchamo: Good catch! Just pushed the fix

menduz (approved):

kernel/packages/decentraland-ecs/types/@decentraland/RestrictedActions/index.d.ts
pravusjif: but the enum could be used with the item indices as usual right? can't we use ints in the RPC?

nchamo: Regular enums won't work on RPS, regardless of whether they are mapped to ints or strings. But your comment gave the idea to try const enums, and they work!!

Thanks @pravusjif !

cc @menduz

PR #1493, opened 2020-10-23
feat: MOTD endpoint

AjimenezDCL:
MOTD now is retrieved from the proper endpoint

### DIFF
--- a/kernel/packages/entryPoints/unity.ts
+++ b/kernel/packages/entryPoints/unity.ts
@@ -6,7 +6,7 @@ global.enableWeb3 = true
 
 import { createLogger } from 'shared/logger'
 import { ReportFatalError } from 'shared/loading/ReportFatalError'
-import { experienceStarted, NOT_INVITED, AUTH_ERROR_LOGGED_OUT, FAILED_FETCHING_UNITY } from 'shared/loading/types'
+import { AUTH_ERROR_LOGGED_OUT, experienceStarted, FAILED_FETCHING_UNITY, NOT_INVITED } from 'shared/loading/types'
 import { worldToGrid } from '../atomicHelpers/parcelScenePositions'
 import {
   NO_MOTD,
@@ -63,7 +63,11 @@ initializeUnity(container)
     })
     i.ConfigureHUDElement(HUDElementID.AIRDROPPING, { active: true, visible: true })
     i.ConfigureHUDElement(HUDElementID.TERMS_OF_SERVICE, { active: true, visible: true })
-    i.ConfigureHUDElement(HUDElementID.TASKBAR, { active: true, visible: true }, { enableVoiceChat: VOICE_CHAT_ENABLED })
+    i.ConfigureHUDElement(
+      HUDElementID.TASKBAR,
+      { active: true, visible: true },
+      { enableVoiceChat: VOICE_CHAT_ENABLED }
+    )
     i.ConfigureHUDElement(HUDElementID.WORLD_CHAT_WINDOW, { active: true, visible: true })
     i.ConfigureHUDElement(HUDElementID.OPEN_EXTERNAL_URL_PROMPT, { active: true, visible: false })
     i.ConfigureHUDElement(HUDElementID.NFT_INFO_DIALOG, { active: true, visible: false })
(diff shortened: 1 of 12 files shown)


kernel/packages/shared/meta/types.ts
```diff
@@ -22,20 +22,21 @@ export type MetaConfiguration = {
...
   renderProfile?: RenderProfile
-  messageOfTheDay?: MessageOfTheDayConfig
+  messageOfTheDay?: MessageOfTheDayConfig | null
+  messageOfTheDayInit?: boolean
 }
 
 export type MessageOfTheDayConfig = {
-  imageUrl: string
-  endUnixTimestamp: number
+  background_banner: string
```
BrianAmadori, line 30: There's a real reason for using snake_case here? It looks inconsistent with the camelCase of the rest of the payload.

kernel/packages/shared/meta/types.ts
```diff
@@ -22,20 +22,21 @@ export type MetaConfiguration = {
...
+  endUnixTimestamp?: number
```
BrianAmadori, line 31: We are going to implement the timestamp logic?

kernel/packages/shared/meta/types.ts
AjimenezDCL: The endpoint is set like this and change it is a bit of a mess @marcosnc

kernel/packages/shared/meta/types.ts
AjimenezDCL: The endpoint doesn't provide any timestamp, we are leaving the field as optional for the future but it will likely go away.
Since the endpoint is customized by wallet I assume it only returns something when a MOTD is set.

BrianAmadori: Ok! thanks for the clarification.

kernel/packages/shared/meta/types.ts
BrianAmadori: Ok, If its that difficult to change we can review the naming later then.

BrianAmadori (approved):

AjimenezDCL:
https://play.decentraland.zone/branch/feat/halloween-2020_message/index.html?ENV=org&position=1%2C0&realm=fenrir-amber

kernel/packages/entryPoints/unity.ts
```diff
@@ -117,7 +121,12 @@ initializeUnity(container)
     }
 
     if (!NO_MOTD) {
-      i.ConfigureHUDElement(HUDElementID.MESSAGE_OF_THE_DAY, { active: false, visible: true }, worldConfig.messageOfTheDay)
+      const messageOfTheDay = await waitForMessageOfTheDay()
```
pablitar, line 124: Do we need to await here blocking the initialization below? Couldn't this execute in parallel, i.e with an anonymous async function?

kernel/packages/shared/meta/index.ts
```diff
@@ -20,3 +23,20 @@ export async function ensureMetaConfigurationInitialized(): Promise<void> {
     })
   })
 }
+
+export async function waitForMessageOfTheDay(): Promise<MessageOfTheDayConfig | null> {
+  const store: Store<RootState> = globalThis.globalStore
+  const worldConfig = store.getState().meta.config.world
```
pablitar, line 29: The best practice would be to implement a selector for this, if there is none already

pablitar (approved):
Approved! But please pay attention to this comment and see if you agree: https://github.com/decentraland/explorer/pull/1493/files#r512167569. I'm worried this change can impact initialization otherwise (for instance, block it in case the MOTD server takes too long to respond).

kernel/packages/shared/meta/index.ts
```diff
@@ -20,3 +23,20 @@ export async function ensureMetaConfigurationInitialized(): Promise<void> {
...
+  if (worldConfig && worldConfig.messageOfTheDayInit) {
+    return Promise.resolve(worldConfig.messageOfTheDay || null)
+  }
+  return new Promise<MessageOfTheDayConfig | null>((resolve) => {
+    const unsubscribe = globalThis.globalStore.subscribe(() => {
```
pablitar, line 34: What happens if the request for message of the day fails? Does this stay blocked forever?

kernel/packages/shared/meta/index.ts
caminoezequiel: yes, you're right I'll change it

PR #1502, opened 2020-10-23
feat: store commit hash to measure performance regressions

eordano:


nchamo (approved):

kernel/static/index.html
```diff
@@ -497,6 +497,7 @@
       document.getElementById('eth-login-confirm-button').disabled = !document.getElementById('agree-check').checked;
     }
   </script>
+  <script>window.VERSION = 'EXPLORER_VERSION';</script>
```
caminoezequiel, line 500: I think we can't use the word `EXPLORER_VERSION` here because it is replace with the commit hash. I'm not sure, I'll check it

caminoezequiel (approved):

pablitar (approved):

PR #1503, opened 2020-10-26
fix: ABConverter texture import issue

BrianAmadori:
This fixes the following: 

- Broken logic in the material/texture steps of `GLTFImporter`.
- Not taking the `TLD` string into account when fetching `sceneIds`.


AjimenezDCL (approved):
We really need both an autoformatting pre-commit hook and something for the IDEs. This PR was way more difficult to review than it should.

BTW, looks good =)

pbosio (approved):

pravusjif (approved):
LGTM, tested the build with `NO_ASSET_BUNDLES` with no problems

PR #1507, opened 2020-10-26
chore: move ABConverter to a namespace

eordano:
This PR changes ABConverter from a `partial abstract` class to a namespace.


BrianAmadori (approved):
Not sure about why are getting that weird diff, but after a quick read, I suppose the only relevant changes are the partial to namespace ones. 

✔️

PR #1515, opened 2020-10-28
fix: worker update error handling

pravusjif:
**WHY?**
Right now scenes code can throw exceptions in their systems `Update()` that can crash the world (for example Websocket exceptions in BattleRacer scene).

**WHAT?**
We added a bit of code for catching those worker-update exceptions that escape our sandboxing and terminating the worker at that point.

**TESTING**
This is happening right now in BattleRacer scene, the explorer crashes (kernel-side) a few seconds after the scene loads.

The crash happens in production at: 
https://play.decentraland.zone/?ENV=zone&LOS=0&position=68%2C-71
or
https://play.decentraland.zone/?ENV=org&LOS=0&position=61%2C-137

And is fixed in this build:
https://explorer.decentraland.zone/branch/fix/WorkerUpdateErrorHandling/index.html?ENV=org&LOS=0&position=61%2C-137
or
https://explorer.decentraland.zone/branch/fix/WorkerUpdateErrorHandling/index.html?ENV=zone&LOS=0&position=68%2C-71


kernel/packages/scene-system/scene.system.ts
```diff
@@ -15,6 +16,19 @@ class WebWorkerScene extends SceneRuntime {
...
+  updateEnabled: boolean = true
+
+  constructor(transport: ScriptingTransport) {
+    super(transport)
+
+    addEventListener('error', (e) => {
+      console.warn('🚨🚨🚨 Unhandled error in scene code. Disabling worker update loop 🚨🚨🚨')
+      console.error(e)
+      this.updateEnabled = false
+      debugger
```
BrianAmadori, line 28: What's the behaviour of `debugger` when dev tools is closed? This wouldn't pause the execution of the client?

menduz: > What's the behaviour of debugger when dev tools is closed?

It does nothing, that is the spec of JS, it is safe and sound

menduz (approved):
:shipit:

kernel/packages/scene-system/scene.system.ts
pravusjif: no, it works only with the dev tools open, that's why we left it there

pbosio: love the idea of putting `debugger` in this kind of errors and those emojis! ❤️

pbosio (approved):

PR #1516, opened 2020-10-28
chore: replace Sprite with Texture2D in avatar snapshot

AjimenezDCL:
Creating a runtime `Sprite` from a texture is slow and expensive and there are not advantages of using them unless you can pack them together in an atlas (which we cannot do) to reduce drawcalls.

Therefore all the use of `Sprite` has been replaced by `Texture2D`  and the prefabs readjusted to use `RawImage` instead of `Image`.


AjimenezDCL:
When testing pay attention to every use case of avatar snapshots. The top right HUD, Expressions, Friends, Chat, Passport...

sandrade-dcl (approved):
Great job! I remember the `Sprite.Create()` had some known performance spikes and this will improve them.

pravusjif (approved):
awsomeeee

PR #1517, opened 2020-10-28
feat: Profiles over comms and guest profiles + Refactor

pablitar:
This PR adds support for sending profiles through comms. This allows people without wallets to send their profiles to others, and be seen as they are seeing themselves.

Also, refactored profiles a bit.
- Separated catalogs to a different package
- Separated profile from identity
- Added support for multiple ephimeral keys per browser
- Added support for multiple local profiles


kernel/packages/shared/comms/index.ts
```diff
@@ -727,7 +824,7 @@ function subscribeToRealmChange(store: Store<RootState>) {
 
 export async function connect(userId: string) {
   try {
-    const user = getCurrentUser()
+    const user = getStoredSession(userId)
```
menduz, line 843: so nice

kernel/packages/shared/comms/index.ts
```diff
@@ -785,7 +882,7 @@ export async function connect(userId: string) {
           },
           authHandler: async (msg: string) => {
             try {
-              return Authenticator.signPayload(getIdentity(), msg)
+              return Authenticator.signPayload(getIdentity() as AuthIdentity, msg)
```
menduz, line 901: Why is that explicit `as` needed?

menduz: oh. null?

pablitar: So it seems. I don't actually remember. I think I may have been more focused on making it compile than on why it didn't.

kernel/packages/shared/catalogs/actions.ts
```diff
@@ -0,0 +1,11 @@
+import { action } from 'typesafe-actions'
```
pablitar, line 1: All this "catalogs" package has been moved from profiles.

kernel/packages/atomicHelpers/base64ToBlob.ts
```diff
@@ -0,0 +1,22 @@
+export function base64ToBlob(base64: string): Blob {
```
nchamo: Would it make sense to clarify in the name that the resulting blob is an image? Or maybe provide the ability to pass the type as a parameter?

kernel/packages/config/index.ts
```diff
@@ -336,3 +336,10 @@ export namespace ethereumConfigurations {
 }
 
 export const isRunningTest: boolean = (global as any)['isRunningTests'] === true
+
+export const genericAvatarSnapshots = {
```
nchamo: Could you add a small doc explaining when this will be used?

kernel/packages/shared/apis/UserIdentity.ts
```diff
@@ -24,24 +27,22 @@ export class UserIdentity extends ExposableAPI implements IUserIdentity {
   @exposeMethod
   async getUserPublicKey(): Promise<string | null> {
     const identity = getIdentity()
-    const user = getCurrentUser()
```
nchamo, line 27: Nice!

kernel/packages/shared/chat/sagas.ts
```diff
@@ -269,12 +269,13 @@ function initChatCommands() {
     const users = [...peerMap.entries()]
 
     const strings = users
-      .filter(([_, value]) => !!(value && value.user && value.user.profile && value.user.profile.name))
+      .filter(([_, value]) => !!(value && value.user && value.user.userId))
       .filter(([uuid]) => userPose[uuid])
       .map(function ([uuid, value]) {
+        const name = getProfile(getGlobalState(), value.user?.userId!)?.name ?? 'unknown'
```
nchamo, line 275: In which case could it be `unknown`?

kernel/packages/shared/chat/sagas.ts
```diff
@@ -407,13 +407,13 @@ function initChatCommands() {
     actionName: 'mute' | 'block' | 'unmute' | 'unblock'
   ) {
     let pastTense: string = actionName === 'mute' || actionName === 'unmute' ? actionName + 'd' : actionName + 'ed'
-    const currentUser = getCurrentUser()
-    if (!currentUser) throw new Error('cannotGetCurrentUser')
+    const currentUserId = getCurrentUserId(getGlobalState())
+    if (!currentUserId) throw new Error('cannotGetCurrentUser')
 
-    const user = findPeerByName(username)
+    const user = findProfileByName(globalThis.globalStore.getState(), username)
```
nchamo:
```suggestion
    const user = findProfileByName(getGlobalState(), username)
```

kernel/packages/shared/chat/sagas.ts
```diff
@@ -347,10 +347,10 @@ function initChatCommands() {
   let whisperFn = (expression: string) => {
     const [userName, message] = parseWhisperExpression(expression)
 
-    const currentUser = getCurrentUser()
-    if (!currentUser) throw new Error('cannotGetCurrentUser')
+    const currentUserId = getCurrentUserId(getGlobalState())
```
nchamo, line 350: I've seen the get or fail combo quite a few times. Would it make sense to create a `getCurrentUserIdOrFail`?

kernel/packages/shared/comms/v1/brokerWorldInstanceConnection.ts
```diff
@@ -255,6 +263,16 @@ export class BrokerWorldInstanceConnection implements WorldInstanceConnection {
     return Promise.resolve()
   }
 
+  sendProfileRequest(position: Position, userId: string, version: number | undefined): Promise<void> {
+    // To be implemented
```
nchamo, line 267: When would this be called? Preview mode?

kernel/packages/shared/comms/v1/brokerWorldInstanceConnection.ts
```diff
@@ -405,7 +423,14 @@ export class BrokerWorldInstanceConnection implements WorldInstanceConnection {
...
                 time: profileData.getTime(),
-                data: { user: userId, version: profileData.getProfileVersion() }
+                data: {
+                  user: userId,
+                  version: profileData.getProfileVersion(),
+                  type:
+                    profileData.getProfileType() === ProfileData.ProfileType.LOCAL
+                      ? ProfileType.LOCAL
+                      : ProfileType.DEPLOYED
+                } // We use deployed as default b
```
nchamo:
```suggestion
                } // We use deployed as default 
```

kernel/packages/config/index.ts
```diff
@@ -336,3 +336,10 @@ export namespace ethereumConfigurations {
 }
 
 export const isRunningTest: boolean = (global as any)['isRunningTests'] === true
+
+export const genericAvatarSnapshots: Record<string, string> = {
```
nchamo, line 340: Would it make sense to add a little doc explaining the use for these?

kernel/packages/shared/chat/sagas.ts
```diff
@@ -507,6 +507,10 @@ function initChatCommands() {
   })
 }
 
+function getGlobalState(): RootState {
+  return globalThis.globalStore.getState()
```
nchamo, line 511: I've seen this A LOT of times over the code. Would it make sense to make a global function out of it?

kernel/packages/shared/comms/index.ts
```diff
@@ -433,6 +483,59 @@ function processVoiceFragment(context: Context, fromAlias: string, message: Pack
   }
 }
 
+const TIME_BETWEEN_PROFILE_RESPONSES = 10000
```
nchamo: Could you add a comment explaining in human terms what this represents?

kernel/packages/shared/comms/index.ts
```diff
@@ -397,7 +447,7 @@ function processChatMessage(context: Context, fromAlias: string, message: Packag
           const messageEntry: InternalChatMessage = {
             messageType: ChatMessageType.PUBLIC,
             messageId: msgId,
-            sender: displayName || 'unknown',
+            sender: user.userId || 'unknown',
```
nchamo, line 450: I didn't understand this change, until I saw what `displayName` was haha

pravusjif (approved):
Only reviewed the build and profile propagation from the user experience side, works like a charm! thanks for this! I think it was really necessary

kernel/packages/shared/session/index.ts
```diff
@@ -1,16 +1,56 @@
...
+    if (oldSession && oldSession.userId === userId) {
+      setStoredSession(oldSession)
+      return oldSession
+    }
+  }
+
+  return null
+}
+
+export const removeStoredSession = () => removeFromLocalStorage('dcl-profile')
```
nchamo: Would it make sense to extract 'dcl-profile'? Like with `SESSION_KEY_PREFIX`

nchamo (approved):

sandrade-dcl:
@pablitar I have been testing and I have found the next weird scenario:

1. Be sure to stay offline in the Metamask plugin (Chrome).
2. Enter https://explorer.decentraland.zone/branch/feat/profiles-over-comms-and-guest-profiles/index.html?ENV=org
3. Notice that you will be able to enter the world even without log in Metamask (it is not happening in prod), loading correctly your avatar stored in localstorage.

**NOTE:** In prod, if you reproduce this scenario, it always open automatically the Metamask window prompting you for enter the credentials, and if you don't do it, it don't let you continue.

Apart from this, everything is working perfectly :) Good work!!

kernel/packages/atomicHelpers/base64ToBlob.ts
pablitar: I just moved this code. So I didn't even bother to see what was inside. But I'll make your change

kernel/packages/shared/chat/sagas.ts
pablitar: If you don't have information of the profile in the redux store, 'unknown' is used as its name for the user. This is rather defensive, but I think we have to consider that problem. Is either this or filtering the player altogether.

kernel/packages/shared/chat/sagas.ts
pablitar: Maybe later :)

kernel/packages/shared/session/index.ts
pablitar: I think this is not used anymore

kernel/packages/shared/comms/v1/brokerWorldInstanceConnection.ts
pablitar: Yep. But I found out that preview mode for the moment uses a default profile, so we don't need to implement this yet. We may want to change this in the future.

kernel/packages/shared/comms/index.ts
pablitar: :shrug:

kernel/packages/shared/comms/index.ts
pablitar: There is a comment below explaining its use (line 502). Don't you think that's enough?

nchamo: It's not necessary for the merge, but I think it would help to also have it right next to the variable

kernel/packages/shared/chat/sagas.ts
pablitar: Probably. But that would affect the whole explorer. We may want to do that in another PR, just for sanity

PR #1519, opened 2020-10-28
feat: enable stateful workers

nchamo:
## What are we doing?
Our current SDK has no concept of state. It simply generates messages that are sent to the renderer, so that it updates the scene in some way. This works, but it has some disadvantages. As part of the builder in-world feature, we want to start considering the scene's state as a first class citizen, and share it across players. By doing so, we will be able to support multiplayer world edit tools, such as the builder in-world.

### General idea
Right now, each scene runs in a different webworker. We decided to create a new webworker that runs completely independent from the others, so that we can switch between them when the user enters and leaves edit mode. The new webworker will contain the scene's state, and keep it in sync with the renderer. It will also be in charge of saving and retrieving the state on a semi-persistent storage.

### Implementation details
We took the current `SceneWorker` and created an abstraction for it. This implied hiding a lot of exposed properties behind class methods. Then, we took the current implementation and called it `SceneSystemWorker`. The new worker is called `StatefulWorker`.

When the user enters edit mode, it will send a message to kernel. Kernel will then kill the scene worker, and start a new stateful one. The opposite will happen when the users leave edit mode. 

We are now storing the state in local storage. This was mainly done to enable an easy integration with the builder in-world feature, without needing to have deploy access to a LAND. The state is stored in a json format, in an easy to read format. We could change this in the future if we want to, but until the feature is stable enough, it will make things easier.

## How to test it locally
1. Run kernel 
```make watch```
2. Go to http://localhost:8080/?DEBUG_MODE&position=-1%2C-51. You will see a regular scene being executed.
3. Open the console and execute
```var sceneId = 'Qm2uuMLmJKSZ1Z4PUnDgxqnvWnVTMGDaJmdke98cTKWHSxVgg649rTiMZojqcTj2ft1X4mb7uZSn11Xi7LBBMtUjTV'```
4. Finally enter stateful mode by executing on the console:
```browserInterface.ControlEvent({eventType: 'StartStateMode', payload: {sceneId}})```
You will see how the scene unloaded, and a new version of the scene loads. That version is defined in a descriptive way.
_Note: the event type was changed to 'StartStatefulMode' in https://github.com/decentraland/explorer/pull/1589_ 

From there on, you can trigger the following actions:
* Leave edit mode:
```browserInterface.ControlEvent({eventType: 'StopStateMode', payload: {sceneId}})```
_Note: the event type was changed to 'StopStatefulMode' in https://github.com/decentraland/explorer/pull/1589_ 
* Add entities:
```browserInterface.SceneEvent({sceneId, eventType:'stateEvent', payload: { type: 'AddEntity', payload: { entityId: 'E2', components: [{ id: 1, data: { position: { x: 8, y: 0, z: 15 }}}, { id: 54, data:{ src: 'models/BlockDog.glb' } }]}}})```
* Modify components:
```browserInterface.SceneEvent({sceneId, eventType:'stateEvent', payload: { type: 'SetComponent', payload: { entityId: 'E1', componentId: 1, componentData: {position: { x: 8, y: 0, z: 1 }}}}})```
* Remove entities:
```browserInterface.SceneEvent({sceneId, eventType:'stateEvent', payload: { type: 'RemoveEntity', payload: { entityId: 'E2' }}})```
* Remove components:
```browserInterface.SceneEvent({sceneId, eventType:'stateEvent', payload: { type: 'RemoveComponent', payload: { entityId: 'E1', componentId: 1 }}})```
* Store the current state in local storage:
```browserInterface.SceneEvent({sceneId, eventType:'stateEvent', payload: { type: 'StoreSceneState', payload: { }}})```

It is important to mention that whenever you execute one of the above, you won't see it reflected on the renderer. This is because we are simulating changes that happen on the renderer, so it makes no sense to send them back. If you want to see the changes reflected, then you have to:
1. Save the changes
2. Reload the scene and enter edit mode, or directly enter edit mode again

## Open questions
**Components support**
With this PR, we are only supporting two types of components: `Transform` and `GLTFShape`. The current builder also supports `Script` and `NFTShape`. Which ones do we want to support for the builder in-world feature?

**Files**
We need to figure out how to handle files, such as gtlfs. We are currently only syncing state data between kernel/worker & the renderer

**Component data**
We need to agree on what the renderer will send to the worker. Will it send the whole component, or will it send just the updates? We should use the same logic for syncing between players in the future.

## Next steps
* Agree on open questions and handle those changes in future PRs
* Start sending the scene's state to content server (when not in `DEBUG_MODE`)
* Generate the scene's code, so that it is compatible with the current way of executing a scene

### CODE REVIEW

File: kernel/packages/unity-interface/BrowserInterface.ts [hunk 1]
```diff
@@ -204,6 +205,18 @@ export class BrowserInterface {
         }
         break
       }
+      case 'StartStateMode': {
```
nchamo, line 212: I don't like this name so much. Maybe something like `StartStatefulMode`? What do you think?

BrianAmadori: Maybe something with the `sync` word in it? `StartSceneSync` `StartSceneSyncMode`?

menduz (approved):

kernel/packages/scene-system/stateful-scene/SceneStateDefinition.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  sendStateTo(container: StateContainer) {
+    for (const [entityId, components] of this.entities.entries()) {
+      const mappedComponents = Array.from(components.entries())
+        .map(([id, data]) => ({ id, data }))
+      container.addEntity(entityId, mappedComponents)
+    }
+  }
+
+  toStorableFormat(): any {
```
pablitar, line 33: Shouldn't we maybe separate this to another module? Something like `StateDefinitionSerializer` or similar.

I think it would make more sense, and would also enable us to separate the concerns of having an scene state and being able to store it somehow.

kernel/packages/scene-system/stateful-scene/SceneStateDefinition.ts
```diff
@@ -0,0 +1,84 @@
...
+      const components = []
+      for (const [componentId, componentData] of entityComponents.entries()) {
+        components.push({ type: idToHumanReadableType(componentId), value: componentData })
+      }
+      entities.push({ id: entityId, components })
+    }
+    return { entities }
+  }
+
+  static fromStorableFormat(data: any): SceneStateDefinition {
```
pablitar, line 45: This would be moved to the serializer too.

kernel/packages/scene-system/stateful-scene/SceneStateDefinition.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+}
+
+
+/**
+ * We are converting from numeric ids to a more human readable format. It might make sense to change this in the future,
+ * but until this feature is stable enough, it's better to store it in a way that it is easy to debug.
+ */
+
+const HUMAN_READABLE_TO_ID: Map<string, ComponentId> = new Map([['Transform', CLASS_ID.TRANSFORM], ['GLTFShape', CLASS_ID.GLTF_SHAPE]])
```
pablitar, line 66: And this too.

kernel/packages/scene-system/stateful.scene.system.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.eventSubscriber = new EventSubscriber(this.engine)
+
+    // Fetch stored scene
+    this.sceneState = SceneStateDefinition.fromStorableFormat(await this.sceneStateStorage.getStoredState(sceneId))
+
+    // Listen to the renderer and update the local scene state
+    this.renderer.forwardChangesTo(this.sceneState)
+
+    // Send the initial state ot the renderer
+    this.sceneState.sendStateTo(this.renderer)
```
pablitar, line 47: This to me seems like a "piping/stream" idea. We might want to explore the posibility of having an abstraction that "connects" different state containers and makes sure they are synced.

kernel/packages/unity-interface/BrowserInterface.ts
pablitar: I like `StartStatefulMode`. What I don't like so much is "start/stop". I think we should name this "SwitchToStatefulMode" and "SwitchToLegacyMode" or "SwitchToRunnerMode" or something like that.

pablitar (approved):
LGTM! Great work!

kernel/packages/scene-system/stateful-scene/RendererActor.ts
```diff
@@ -0,0 +1,168 @@
+import { CLASS_ID } from "decentraland-ecs/src";
+import { IEngineAPI } from "shared/apis/EngineAPI";
+import { AttachEntityComponentPayload, ComponentCreatedPayload, ComponentRemovedPayload, ComponentUpdatedPayload, CreateEntityPayload, EntityAction, RemoveEntityPayload, UpdateEntityComponentPayload } from "shared/types";
+import { Component, ComponentData, ComponentId, EntityId, StatefulActor } from "./types"
+import { EventSubscriber } from "decentraland-rpc";
+import { generatePBObjectJSON } from "scene-system/sdk/Utils";
+
+export class RendererActor extends StatefulActor {
```
pablitar, line 8: I think this should maybe be named `RendererStatefulActor`. `RendererActor` by itself doesn't convey that the state is important for this class

pravusjif (approved):
I like the aproach, didn't test it locally, had no issues with the current build in prod

pravusjif:
@nchamo I had to revert this PR merged in master (https://github.com/decentraland/explorer/commit/41070c3c8f65651c4d208657ae99d2797956bbdc) because it breaks something related to the **playerPosition** used from the SDK.

It is reproduced with 100% repro rate at Salmonomicon https://explorer.decentraland.zone/branch/feat/enable-stateful-workers/index.html?ENV=org&position=-53,2
you'll see that when the monster finishes spawning he always goes to the west (instead of going to the player, I assume it is going to the 0,0,0). Also, Mika's eyes that should point to the nearest page are always pointing to the east, so the problem is probably with the player position. You can see that it works well in prod https://play.decentraland.org/?position=-53,2

I took an hour to debug but didn't find what's provoking the bug.

I **restored this PRs branch** but **I can't re-open the PR** probably because I didn't create it, not sure why.

Issue #1520, opened 2020-10-28
[Regression] Avatar modifier areas not working properly


nchamo:
The SDK currently provides two different types of avatar modifiers: `HIDE_AVATARS` and `DISABLE_PASSPORTS`.

They are currently not working properly:
* `HIDE_AVATARS` hides avatars, but when the avatar leaves the area, it remains invisible
* `DISABLE_PASSPORTS` is not disabling passports

There is a test scene that covers both scenarios. It can be executed by running the kernel and entering: http://localhost:8080/?DEBUG_MODE&position=0%2C103

PR #1522, opened 2020-10-29
fix: exiting from an AvatarModifier volume is handled properly

AjimenezDCL:
#1520


sandrade-dcl (approved):

sandrade-dcl (approved):

nchamo (approved):
Great work!

Tested in zone (70,79)

PR #1523, opened 2020-10-29
chore: decouple usercontextmenu from friendlist

pbosio:
# What? 
*decoupled UserContextMenu from friendlist and others.
*modified it to use a single prefab for all use cases.
*context menu now has the responsibility to handle actions to avoid having code duplications.
*removed duplicated code

# Why? 
we had duplicated code and messy modifications on the prefab to support the context menu in a couple of places, we need to add more context menu support in others part of the UI so it's better to have a simple solution and remove the need to duplicate code everywhere.


sandrade-dcl (changes requested):
I found the already classic issue with some texts xD

image

sandrade-dcl (changes requested):
The same in the menu from workd chat:
image

pbosio:
> I found the already classic issue with some texts xD

classic! the funny things is that I never changed those fonts! I only moved the prefab and did some other modifications! 😭

sandrade-dcl (approved):
Good work and good code cleaning! ...and thank you for add documentation comments in new classes/methods :)

NOTE: I will test again once the build is ready.

sandrade-dcl (approved):
Now it is working fine! Texts issues fixed!

pravusjif (approved):
Awsomeeee

pravusjif (approved):

PR #1529, opened 2020-10-29
fix: optimizing rotation keyframes wasn't working correctly

BrianAmadori:
With the introduction of custom keyframe optimization code to avoid out of memory errors due to badly optimized animations in the SoYou scene (#1253),  some complex animations stopped working. This is particulary important for the halloween event party, where some skeletons dancing animations look like they are bending themselves backwards. (no pun intended)

This probably has to do with quaternion -> euler data mismatch after the optimization takes place. The keyframes come in quaternion values, but upon visualizing the animation curves in Unity's animation window, the resulting values are in euler. I didn't find a way to force Unity to use euler interpolation from the legacy animations scripting code, so we are disabling the keyframe optimizations for rotation curves.

Test the following with `NO_ASSET_BUNDLES` flag:

- [x] ORG 29,9: Inside the dome there's a dancing guy. He shouldn't be "bending himself backwards" [(Link)](https://explorer.decentraland.zone/branch/fix/complex-animation-import-issue/index.html?ENV=org&position=29,9&NO_ASSET_BUNDLES)
- [x] SoYou scene: shouldn't crash the client [(Link)](https://explorer.decentraland.zone/branch/fix/complex-animation-import-issue/index.html?ENV=org&position=24,-127&NO_ASSET_BUNDLES)


pbosio (approved):

pravusjif (approved):
Everything worked as expected

PR #1541, opened 2020-11-02
fix: simplify FPS metrics

BrianAmadori:
Including changes discussed in [this document](https://docs.google.com/document/d/1JPB2cgAx76Yvy6xdkMTS3OQSxlHox5KEYno441aPhPA/edit?ts=5fa04fe5)

Changes:

- PerformanceMetricsController is not a MonoBehaviour anymore
- some renaming kernel side to match the rendererState naming
- removed sending hiccup data
- only send FPS metrics when we are in the foreground and rendering is activated

TEST LINK: https://play.decentraland.zone/branch/fix/simplify-fps-metrics/index.html?ENV=org


kernel/packages/shared/world/worldState.ts
```diff
@@ -1,35 +1,50 @@
...
-export function onNextWorldRunning(callback: Function) {
-  const observer = worldRunningObservable.add((isRunning) => {
+export function onRendererEnabled(callback: Function) {
+  const observer = renderStateObservable.add((isRunning) => {
     if (isRunning) {
-      worldRunningObservable.remove(observer)
+      renderStateObservable.remove(observer)
       callback()
     }
   })
```
menduz, line 50: Line 47 removes the observer as soon as it is called the first time, in that case the function should be called `onNextRendererEnabled`. On the contrary, `onRendererEnabled` may suggest that it will be called every time the renderer gets re-enabled and raise confusions.

menduz (changes requested):

kernel/packages/shared/world/worldState.ts
BrianAmadori: Good catch. Applied.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -123,6 +125,8 @@ private void Update()
                 sceneSortDirty = false;
                 SortScenesByDistance();
             }
+
+            performanceMetricsController?.Update();
```
eordano, line 129: Would like to see this on top of the `Update` call. My thinking is: how does this behave, if we miss a frame?

BrianAmadori: `Update` is called all the frames. If a frame takes 100ms, this'd be called at the start of the next frame. 

This is just the result of collapsing `Update` calls after the `PerformanceMetricsController` lost its `MonoBehaviour` status. There should be not any meaningful difference in flow from the previous commits.

pravusjif (approved):
didn't have any problems with the build, the code looks good to me.

pablitar (approved):
LGTM!

menduz (approved):

Issue #936, opened 2020-05-25
[glTF] metallicFactor and metallicRoughness are always 1 when no texture is set


EibrielInv:
When using a texture only for `metallicRoughnessTexture` (and not `metallicFactorTexture`) the material ignores the value on `metallicFactor` (assuming always a value of 1)

glTF file to reproduce bug:
[MetalBug.zip](https://github.com/decentraland/explorer/files/4678716/MetalBug.zip)

Comparison between SDK Explorer and Babylon Sandbox:
![Selección_999(678)](https://user-images.githubusercontent.com/172656/82840821-51123c80-9eaa-11ea-850d-9effc2c8ed39.png)

Same behavior when using only `metallicFactorTexture` and not `metallicRoughnessTexture` is set.

![image](https://user-images.githubusercontent.com/172656/82841664-4a38f900-9ead-11ea-9f7c-cd50986abebe.png)

The workaround is easy, just use textures on both `metallicFactorTexture` and  `metallicRoughnessTexture`

EibrielInv:
Same behavior when using only `metallicFactorTexture` and not `metallicRoughnessTexture`

![image](https://user-images.githubusercontent.com/172656/82841664-4a38f900-9ead-11ea-9f7c-cd50986abebe.png)

PR #1542, opened 2020-11-02
fix: metallic factor is applied with textures as well

AjimenezDCL:
#936 Metallic factor is now applied along a metallic texture as well. This will bring Unity's render closer to Babylon's sandbox

Before:
image

After:
image


AjimenezDCL:
Remember to test thoroughly without Asset Bundles to find regressions

menduz:
looking good

pravusjif:
at this coordinates in ZONE we have a special test scene for material properties
https://play.decentraland.zone/branch/fix/936_metallic_factor_with_textures/index.html?ENV=zone&NO_ASSET_BUNDLES&position=58%2C-71

pravusjif (approved):
Didn't see anything strange, although maybe it would be wise to show this build to one of the Artists like Shibu to see if they catch something odd

BrianAmadori (approved):

ShibuyaMartin (approved):

Issue #1340, opened 2020-09-21
Scene performance drops after being open 20-30 minutes


nearnshaw:
Baus (Miles) reported this issue. He created a simple scene and it works fine if loaded freshly. But if he keeps the scene open for an hour and then tries to interact with it, the update function is very laggy. Even though the reported FPS are still good.

Here's his description:

Did some tests and there's definitely some weird stuff going on with the DCL engine's update() logic that's responsible for the stuttering issues that worsen over time in blackjack and poker

```ts
import utils from "../node_modules/decentraland-ecs-utils/index";
let box = new Entity();
box.addComponent(new BoxShape());
box.addComponent(
  new Transform({
    position: new Vector3(8, 0, 8),
  })
);
box.addComponent(
  new OnPointerDown((e) => {
    spawnCubes();
  })
);
engine.addEntity(box);
function spawnCubes() {
  let root = new Entity();
  engine.addEntity(root);
  const blue = new Material();
  blue.albedoColor = Color3.Blue();
  for (let i = 0; i < 5; i++) {
    let el = new Entity();
    el.addComponent(new BoxShape());
    el.addComponent(blue);
    let startPos = new Vector3(6, 0.3 + i * 0.5, 6);
    el.addComponent(
      new Transform({
        position: startPos,
        scale: new Vector3(0.3, 0.3, 0.3),
      })
    );
    el.setParent(root);
    el.addComponent(
      new utils.Delay(500, () => {
        el.addComponent(
          new utils.MoveTransformComponent(startPos, startPos.add(new Vector3(2, 0, 0)), 1.5, () =>
            el.addComponent(
              new utils.Delay(500, () => {
                engine.removeEntity(root);
              })
            )
          )
        );
      })
    );
  }
}
```

So I used this pretty basic code snipped that triggered an animation when the white button was pressed, this is using utils.MoveTransformComponent in DCL's utils library

Here's a vid of an instance where I let it sit for around an hour and triggered it and the animations are insanely laggy even though the client maintains >140 fps, then you can see right afterwards when I load a fresh instance it works properly (edited)

https://decentralandteam.slack.com/files/UNGGR372Q/F01BSU3HA2U/2020-09-20_13-50-10.mp4

This is with my desktop which is pretty well-specced so I have a feeling this could be happening even quicker on laptops (like ~20 mins like the blackjack feedback we got)

pravusjif:
This issue is solved in 2 fronts:

[...]

PR #1551, opened 2020-11-03
chore: engine getcomponent optimization

pravusjif:
**WHY?**
Currently, every time an `engine.getComponentGroup()` is called, a new component group is always created, leading to memory leaks when used aggresively (e.g. in a system update()) like the one reported at #1340 

**WHAT?**
Added a bit of code to traverse the already existent (we were aleady caching them) component groups and return the one with exactly the same requirements. Otherwise, we follow the normal flow and create and store it.

### DIFF
--- a/kernel/packages/decentraland-ecs/src/ecs/Engine.ts
+++ b/kernel/packages/decentraland-ecs/src/ecs/Engine.ts
@@ -264,7 +264,34 @@ export class Engine implements IEngine {
   }
 
   getComponentGroup(...requires: ComponentConstructor<any>[]) {
-    // TODO: memoize?
+
+    // Return an already created component-group if it already exists
+    if (requires.length > 0) {
+      // 1. get component groups for first require
+      let componentGroups = this._componentGroups[getComponentName(requires[0])]
+
+      if (componentGroups) {
+        const components = requires.slice()
+
+        // 2. search for a component group that has all the same requirements
+        componentGroups.forEach(componentGroup => {
+          if (components.length === componentGroup.requiresNames.length) {
+            for (let index = 0; index < components.length; index++) {
+              const componentName = getComponentName(components[index])
+
+              if (componentGroup.requiresNames.indexOf(componentName) === -1) break
+
+              if (index === (components.length - 1)) {
+                // 3. Found an existent component group with the exact same requirements
+                return componentGroup
+              }
+            }
+          }
+        })
+      }
+    }
+
+    // Otherwise create and store it
     const componentGroup = new ComponentGroup(...requires)
 
     componentGroup.active = true
--- a/kernel/packages/decentraland-ecs/src/ecs/Engine.ts
+++ b/kernel/packages/decentraland-ecs/src/ecs/Engine.ts
@@ -278,12 +278,9 @@ export class Engine implements IEngine {
         componentGroups.forEach(traversedComponentGroup => {
           if (components.length === traversedComponentGroup.requires.length) {
             for (let i = 0; i < components.length; i++) {
-              console.log("looking for " + getComponentName(components[i]) + " in " + traversedComponentGroup.requiresNames)
-
               if (traversedComponentGroup.requires.indexOf(components[i]) === -1) break
 
               if (i === (components.length - 1)) {
-                // 3. Found an existent component group with the exact same requirements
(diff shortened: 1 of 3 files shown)


sandrade-dcl (approved):

kernel/packages/decentraland-ecs/src/ecs/Engine.ts
```diff
@@ -264,7 +264,34 @@ export class Engine implements IEngine {
...
+      let componentGroups = this._componentGroups[getComponentName(requires[0])]
+
+      if (componentGroups) {
+        const components = requires.slice()
+
+        // 2. search for a component group that has all the same requirements
+        componentGroups.forEach(componentGroup => {
+          if (components.length === componentGroup.requiresNames.length) {
+            for (let index = 0; index < components.length; index++) {
+              const componentName = getComponentName(components[index])
```
menduz: Component class should be used instead of name (names can be repeated)

kernel/packages/decentraland-ecs/src/ecs/Engine.ts
```diff
@@ -264,7 +264,34 @@ export class Engine implements IEngine {
...
+
+              if (componentGroup.requiresNames.indexOf(componentName) === -1) break
+
+              if (index === (components.length - 1)) {
+                // 3. Found an existent component group with the exact same requirements
+                return componentGroup
```
menduz: That return exits the `componentGroups.forEach(componentGroup => {`, not the `function getComponentGroup`

menduz (changes requested):
Although the motivation is sound, it seems like the implementation doesn't do what is expected

kernel/packages/decentraland-ecs/src/ecs/Engine.ts
pravusjif: With component class this doesn't work, it never finds a previously created component group, that was my first aproach.

kernel/packages/decentraland-ecs/src/ecs/Engine.ts
pravusjif: Daaaaaaaaamn!

pravusjif: I'll re-check the whole implementation then

kernel/packages/decentraland-ecs/src/ecs/Engine.ts
pravusjif: tried again after refactoring and using plain components works fine now

kernel/packages/decentraland-ecs/src/ecs/Engine.ts
pravusjif: fixed

menduz (approved):
It looks good, we could use a test just in case. like these: https://github.com/decentraland/explorer/blob/77c64d1bea27225adaac3ccba30e08b87452959e/kernel/test/unityIntegration/ecs/math/vector3.test.ts

pravusjif:
Added a test although locally I see it is a "pending" test like many others we have that are not being run, hopefully it will run in the CI...

Screen Shot 2020-11-04 at 17 44 04

kernel/test/unit/engine.test.tsx
```diff
@@ -0,0 +1,28 @@
+import { expect } from 'chai'
+import { Entity, BoxShape, Transform, Vector3, engine } from 'decentraland-ecs/src';
+
+describe.skip('Engine tests', () => {
```
menduz:
```suggestion
describe('Engine tests', () => {
```

menduz (changes requested):

kernel/test/unit/engine.test.tsx
menduz: @pravusjif the test was skipped by code. please pull and run it in the UI to verify

menduz: It failed in the CI :( 
```
  1) Engine tests
       getComponentGroupReturnsCachedGroup:
     TypeError: c.expect(...).to.same is not a function
      at Context.<anonymous> (test/out/index.js:661:3011493)
```

menduz (approved):
Ship it

PR #1552, opened 2020-11-04
feat: autosettings based on performance

AjimenezDCL:
Quality settings will be adjusted automatically based on performance metrics.

It's yet to be defined how or if the user will have input on this. Probably we will merge it with a feature toggle first.

https://docs.google.com/document/d/1OMouC890S_Akw2SKvENE1UZ6qYY8FExic-G64pOTnwM


AjimenezDCL:
https://play.decentraland.zone/branch/feat/autosettings_by_performance/index.html?ENV=org&realm=fenrir-amber

menduz:
Could you explain how it works? I think this may be one of those things worth documenting

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/SettingsHUD/SettingsGeneralView.cs
```diff
@@ -162,6 +167,21 @@ void Awake()
...
+                advancedCanvasGroup.interactable = !value;
+                tempGeneralSetting.autoqualityOn = value;
+                advancedBlocker.SetActive(value);
+                if (value)
+                {
+                    QualitySettings.BaseResolution currentBaseResolution = tempQualitySetting.baseResolution;
+                    tempQualitySetting = Settings.i.lastValidAutoqualitySet;
+                    tempQualitySetting.baseResolution = currentBaseResolution;
+                    isDirty = true;
+                }
```
sandrade-dcl: In order to improve the code readability, I think it would be fine to extract this piece of code to a private function (something like `SetAutoSettingsActive(bool active)`). What do you think?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Settings/Settings.cs
```diff
@@ -76,12 +91,18 @@ private void LoadGeneralSettings()
             }
         }
 
-        public void ApplyQualitySettingsPreset(int index)
+        public void ApplyAutoQualitySettings(int index)
```
sandrade-dcl, line 98: Only as suggestion: As a public function that will be called from other entitie(s), it would be fine to add a header with some general documentation about what this function does.

sandrade-dcl (changes requested):
The code looks fine! I only left a pair of little suggestions :)

**Tests:** In my current PC is hard to test this feature for the "low-specs machines" use case, but I'm going to try on an old PC that I have (in some place xD).

sandrade-dcl:
Ok, I was able to test the Auto-Settings feature in a very old machine and I found some interesting things:

In this case where I'm using a very low-specs machine, I have noticed that the experience is not improving in a very noticiable way. In some ways this is normal because of the low-specs of the hardware, but I have noticed something that I think we should take into account on this feature:

- If I set manually the **SHADOWS** to **OFF**, my experience improves a lot.
- If I set manually the **RENDERING SCALE** to **0.5**, my experience improves a bit more.

However if I activate the auto-settings mode, these 2 parameters are never being changed, the SHADOWS is always set to ON and the RENDERING SCALE is always set to 0.7 or more.

**I think if we make this auto-settings feature takes into account these 2 params, we can improve a lot the experience for a lot of users.**

image

AjimenezDCL:
These values are in fact overrode by autosettings, bear in mind that the quality is being adjusted when the average FPS of the last 5 seconds (to have a big enough sample size) are under a threshold. Probably you were just traveling from autosetting to autosetting until the system finds the best one for you.

We can reduce that sample size needed though, but we could face some false positive with hiccups. It needs more fine-tunning.

pravusjif (approved):
It probably could use more tweaking but I think we can adjust it on the run.

I've been checking out a specific scene that I found out the FPS always drop in there, it's Neo Arcade, for testing this feature I recommend toggling on the auto-settings and walk towards that place, inside that place, the FPS always drop a lot 

Screen Shot 2020-11-12 at 01 48 46

sandrade-dcl (approved):

PR #1560, opened 2020-11-04
fix: nullreference exception when stopping editor

pbosio:

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/UsersAroundListHUD/UsersAroundListHUDListView.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/UsersAroundListHUD/UsersAroundListHUDListView.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using DCL.Helpers;
 using UnityEngine;
 using UnityEngine.UI;
 
@@ -156,7 +155,7 @@ internal class UsersAroundListHUDListView : MonoBehaviour, IUsersAroundListHUDLi
 
         if (!isGameObjectDestroyed)
         {
-            Utils.SafeDestroy(gameObject);
+            Destroy(gameObject);
         }
     }
 
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/UsersAroundListHUD/UsersAroundListHUDListView.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/UsersAroundListHUD/UsersAroundListHUDListView.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using DCL.Helpers;
 using UnityEngine;
 using UnityEngine.UI;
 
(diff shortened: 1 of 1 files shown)


sandrade-dcl (approved):

AjimenezDCL (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/UsersAroundListHUD/UsersAroundListHUDListView.cs
```diff
@@ -145,7 +153,11 @@ void IUsersAroundListHUDListView.Dispose()
 
         userElementDictionary.Clear();
         availableElements.Clear();
-        Destroy(gameObject);
+
+        if (!isGameObjectDestroyed)
+        {
+            Utils.SafeDestroy(gameObject);
```
BrianAmadori: This check shouldn't be necessary. 

`Object.Destroy` doesn't fail when a null object is trying to be destroyed. We can stop using `Utils.SafeDestroy`, I think it doesn't add anything in this case.

pbosio: Sorry, you are right...I tried using `SafeDestroy` at first and didn't work... I added the flag and forgot that there 😅

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/UsersAroundListHUD/UsersAroundListHUDListView.cs
BrianAmadori: I just saw the changes, we shouldn't remove the `isGameObjectDestroyed` flag as well?

pbosio: Nope, It seems that there is a strange condition where the MonoBehaviour is destroyed and not destroyed at the same time? 📦🐈 
So trying to access `gameObject` in this situation arise the exception. I don't know what the `gameObject` property is doing internally 🤷‍♂️
Anyway, that flag is the only thing preventing the null reference exception

AjimenezDCL: It's, actually, the other way around.

Unity's shutdown is destroying the `gameobject` holding `UsersAroundListHUDListView` and `UsersAroundListHUDListViewController.Dispose` is doing the same but later. The problem is there's no way for the controller to know that the view was destroyed so, indeed, the flag is needed.

BrianAmadori: Ugh, this `Destroy` thing on stop is really a nasty issue. Anyway, resolving the discussion.

BrianAmadori (approved):

PR #1563, opened 2020-11-04
fix: hover feedback when blocked by ui

pravusjif:
**WHY?**
right now whenever we click on an object that triggers a UI (like an NFTShape), we are still displaying the hover feedback of the behind-object that we just interacted with (and allowing its interaction as well)

**WHAT?**
This PR adds an extra graphics ray cast (only UI, and **only when the pointer event interaction ray already detected an interaction**) to detect if there is UI in the middle of the screen, where the pointer is. In that case, we consider the UI as blocker to the interaction of the object behind it.

**TEST**
You can easily test in Genesis Plaza's nft art gallery (-3, -2):

prod: https://play.decentraland.org/
this PR: https://explorer.decentraland.zone/branch/fix/HoverFeedbackWhenBlockedByUI/index.html?ENV=org


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PointerEventsController.cs
```diff
@@ -41,7 +44,22 @@ public void Update()
...
 
             // We use Physics.Raycast() instead of our raycastHandler.Raycast() as that one is slower, sometimes 2x, because it fetches info we don't need here
-            if (!Physics.Raycast(GetRayFromCamera(), out hitInfo, Mathf.Infinity, PhysicsLayers.physicsCastLayerMaskWithoutCharacter))
+            // TODO: POLISH AND OPTIMIZE
+            bool didHit = Physics.Raycast(GetRayFromCamera(), out hitInfo, Mathf.Infinity, PhysicsLayers.physicsCastLayerMaskWithoutCharacter);
+            bool UIIsBlocking = false;
+
+            if (didHit)
+            {
+                GraphicRaycaster raycaster = GameObject.FindObjectOfType<Canvas>().GetComponent<GraphicRaycaster>();
```
BrianAmadori: There's no way of having more than a single `GraphicRaycaster` in the scene? Also, this looks like a nasty approach that would give a performance hit on all detections. We should at least cache the object.

pravusjif: I was just testing the aproach, this is by no means the final implementation, as the PR is still in WIP =P

sandrade-dcl (approved):

BrianAmadori (approved):

PR #1566, opened 2020-11-05
fix: invalid fps stacking when out of focus

BrianAmadori:
With the current approach there was the possibility to send invalid FPS metrics within a single batch. 

Any frame metric is invalid when is stacked on these cases:
* Tab out of focus
* Rendering deactivated

### DIFF
--- a/kernel/packages/entryPoints/unity.ts
+++ b/kernel/packages/entryPoints/unity.ts
@@ -15,7 +15,7 @@ import { StoreContainer } from 'shared/store/rootTypes'
 import { startUnitySceneWorkers } from '../unity-interface/dcl'
 import { initializeUnity } from '../unity-interface/initializer'
 import { HUDElementID, RenderProfile } from 'shared/types'
-import { renderStateObservable, onNextRendererEnabled } from 'shared/world/worldState'
+import { renderStateObservable, onNextRendererEnabled, foregroundObservable } from 'shared/world/worldState'
 import { getCurrentIdentity } from 'shared/session/selectors'
 import { userAuthentified } from 'shared/session'
 import { realmInitialized } from 'shared/dao'
@@ -84,7 +84,7 @@ initializeUnity(container)
 
       const voiceChatEnabled = isVoiceChatEnabledFor(globalThis.globalStore.getState(), identity.address)
       configureTaskbarDependentHUD(i, voiceChatEnabled)
-      
+
       i.ConfigureHUDElement(HUDElementID.USERS_AROUND_LIST_HUD, { active: voiceChatEnabled, visible: false })
 
       i.ConfigureHUDElement(HUDElementID.FRIENDS, { active: identity.hasConnectedWeb3, visible: false })
@@ -121,6 +121,14 @@ initializeUnity(container)
     } else {
       i.SetRenderProfile(RenderProfile.DEFAULT)
     }
+
+    foregroundObservable.add((isForeground) => {
+      if (isForeground) {
+        i.ReportFocusOn()
+      } else {
+        i.ReportFocusOff()
+      }
+    })
 
     if (!NO_MOTD) {
       waitForMessageOfTheDay().then((messageOfTheDay) => {
--- a/kernel/packages/entryPoints/unity.ts
+++ b/kernel/packages/entryPoints/unity.ts
@@ -122,7 +122,7 @@ initializeUnity(container)
       i.SetRenderProfile(RenderProfile.DEFAULT)
     }
 
-    if ( isForeground() ) {
+    if (isForeground()) {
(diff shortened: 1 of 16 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/ScriptableObject/CommonScriptableObjects.cs
```diff
@@ -122,6 +122,10 @@ public static class CommonScriptableObjects
     public static RendererState rendererState => GetOrLoad(ref rendererStateValue, "ScriptableObjects/RendererState");
     private static RendererState rendererStateValue;
 
+    public static BooleanVariable focusState => GetOrLoad(ref focusStateValue, "ScriptableObjects/FocusState");
+    private static BooleanVariable focusStateValue;
```
eordano, line 126: `= true`?

kernel/packages/shared/world/worldState.ts
```diff
@@ -2,19 +2,32 @@ import { Observable } from '../../decentraland-ecs/src/ecs/Observable'
...
 
 export const renderStateObservable = new Observable<Readonly<boolean>>()
+export const foregroundObservable = new Observable<Readonly<boolean>>()
+
+function handleVisibilityChange() {
+  foregroundObservable.notifyObservers(isForeground())
+}
 
 renderStateObservable.add((state) => {
   rendererEnabled = state
```
eordano, line 33: Thinking about the initial load. Should we `foregroundObservable.notifyObservers()` here?

unity-client/Assets/Scripts/MainScripts/Debugging/PerformanceMetricsController.cs
```diff
@@ -1,3 +1,4 @@
+using System.Reflection.Emit;
```
AjimenezDCL: I think this is not needed

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/Debugging/WSS/WSSController.cs
```diff
@@ -302,11 +302,17 @@ void Update()
...
+                            case "ActivateRendering":
+                                renderingController.ActivateRendering();
+                                break;
                             case "DeactivateRendering":
                                 renderingController.DeactivateRendering();
                                 break;
-                            case "ActivateRendering":
-                                renderingController.ActivateRendering();
+                            case "ReportFocusOn":
+                                CommonScriptableObjects.focusState.Set(true);
```
AjimenezDCL: As a rule of thumb I try to avoid setting values of a SO from different places. Cannot we use the Bridge here so we centralize the setter of this value there?

AjimenezDCL (approved):

unity-client/Assets/Scripts/MainScripts/DCL/ScriptableObject/CommonScriptableObjects.cs
BrianAmadori: Just added the initialization here:
https://github.com/decentraland/explorer/blob/d1b391fafbdb2ff2c04f1e2f426a611167d2ccd8/unity-client/Assets/Scripts/MainScripts/DCL/Bridges/FocusStateBridge/FocusStateBridge.cs#L9

unity-client/Assets/Scripts/MainScripts/Debugging/PerformanceMetricsController.cs
BrianAmadori: good catch, fixed

unity-client/Assets/Scripts/MainScripts/Debugging/WSS/WSSController.cs
BrianAmadori: makes sense, done!

kernel/packages/shared/world/worldState.ts
BrianAmadori: To be 100% on the safe side I added an initial notification with the current focus state in `unity.ts`.

kernel/packages/entryPoints/unity.ts
```diff
@@ -122,6 +122,20 @@ initializeUnity(container)
       i.SetRenderProfile(RenderProfile.DEFAULT)
     }
 
+    if ( isForeground() ) {
```
eordano:
```suggestion
    if (isForeground()) {
```

eordano (approved):
Approved with minor styling issue.

kernel/packages/entryPoints/unity.ts
BrianAmadori: Done.

PR #1569, opened 2020-11-05
chore: Meta config URL change

pablitar:
Changed meta configuration url to config.decentraland.<env>.

Also, made it configurable using URL, so we can test against custom configurations before releasing them.


pbosio (approved):
This is great. Thanks!

caminoezequiel (approved):

PR #1579, opened 2020-11-06
feat: voicechat animation in avatar shape

pbosio:
Add animation to indicate that an avatar shape is talking
Screen Shot 2020-11-10 at 10 38 11
Screen Shot 2020-11-10 at 10 41 32


unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/Animations/VoiceChat/VoiceIconController.controller
```diff
@@ -0,0 +1,211 @@
+%YAML 1.1
```
BrianAmadori: We should put the voice chat animations in a folder containing all the `AvatarName` related stuff too, instead of the avatar `Animations` folder. This way all the assets related with the name UI are in the same place.

BrianAmadori (commented):
I'm very wary about the animation approach used here. In any concurred event, rendering 20+ UI animations that use scaling on 3 shapes just to give voice chat feedback may give us a noticeable performance degradation (scale is the most expensive attribute to animate). 

Also, remember that changes to any UI transform will re-trigger the canvas layout calculations, so I think this PR will potentially give us regressions and future PRs to address this issue. 

We definitely should test the performance impact with many AvatarNames present. If we find that this is a problem, I recommend trying out another approach by using sprite-sheet animations or not using animations at all.

pbosio:
After doing some tests it looks like adding this animations has no impact on the performance

BrianAmadori (approved):

pravusjif (approved):

PR #1588, opened 2020-11-10
[ProfileHUD] Fix LogOut flow

sandrade-dcl:
This PR contemplates these **2 use cases**:

- When a **GUEST USER** clicks on `ProfileHUD` -> `Sign In` button, the app should redirect him to the landing page, adding the url param `show_wallet=1` (this will make the wallet selector be opened automatically).
- When a **LOGGED USER** clicks on `ProfileHUD` -> `LogOut` button, the app should redirect him to the landing page, removing these 2 url params: `show_wallet=1` and `position`.

### DIFF
--- a/kernel/packages/shared/session/index.ts
+++ b/kernel/packages/shared/session/index.ts
@@ -92,10 +92,13 @@ export class Session {
     sendToMordor()
     disconnect()
     removeStoredSession(getIdentity()?.address)
+    removeUrlParam('position')
+    removeUrlParam('show_wallet')
     window.location.reload()
   }
 
   async redirectToSignUp() {
+    removeUrlParam('position')
     window.location.search += '&show_wallet=1'
   }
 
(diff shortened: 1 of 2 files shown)


kernel/packages/shared/session/index.ts
```diff
@@ -124,3 +126,18 @@ export async function userAuthentified(): Promise<void> {
...
+  let parameterName
+
+  for (let i = 0; i < urlVariables.length; i++) {
+    parameterName = urlVariables[i].split('=')
+    if (parameterName[0] !== paramToRemove) {
+      url = url + parameterName[0] + '=' + parameterName[1] + '&'
+    }
+  }
+  window.history.replaceState({}, document.title, url.substring(0, url.length - 1))
+}
```
caminoezequiel: I think we could use this function [Location search](https://developer.mozilla.org/en-US/docs/Web/API/Location/search) and then remove elements with `params.delete(paramToRemove)`

sandrade-dcl: You're right! I have applied a refactor of the `removeUrlParam()` function. Please take a look! Thanks! :)

caminoezequiel (approved):

PR #1594, opened 2020-11-10
Fix filters public_url for react app when branch is master, staging or release

caminoezequiel:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -294,14 +294,13 @@ jobs:
           name: "post install script"
           command: sh scripts/post-install.sh
       - run:
-          name: "creating .env (ignore master)"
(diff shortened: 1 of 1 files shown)


pablitar:
What's going on here? I assume this fixes something on the build, but I'd like to know more hahaha

caminoezequiel:
@pablitar 
TLDR We need to set `PUBLIC_URL` with the final url only when we are deploying on a sub directory. It allows to React resolve images, css, etc. But when we are deploying on a root (ex. master, staging and release) it's not necessary.

Here I am using bash conditions to know if the branch requires `PUBLIC_URL` because `circleci` doesn't allows skip step filtering by branch name

pablitar (approved):

PR #1595, opened 2020-11-10
fix: Don't use a hardcoded catalyst when signing up

pablitar:
Since there is a need for sign up to have a catalyst configured in order to function, a hardcoded catalyst was implemented.

But this is not necessary as we don't need to login to fetch the catalysts from the DAO.

Also: Added an http request to avoid the need to use the catalyst contract directly. This falls back to DAO if there is an issue.

### DIFF
--- a/kernel/packages/shared/dao/index.ts
+++ b/kernel/packages/shared/dao/index.ts
@@ -103,7 +103,7 @@ async function fetchCatalystNodes(endpoint: string | undefined) {
     }
   }
 
-  return await fetchCatalystNodesFromDAO()
+  return fetchCatalystNodesFromDAO()
 }
 
 export async function fetchCatalystRealms(nodesEndpoint: string | undefined): Promise<Candidate[]> {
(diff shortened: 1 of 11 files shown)


kernel/packages/shared/dao/index.ts
```diff
@@ -88,8 +88,26 @@ export function ping(url: string, timeoutMs: number = 5000): Promise<PingResult>
...
+      }
+    } catch (e) {
+      defaultLogger.warn(`Tried to fetch catalysts from ${endpoint} but failed. Falling back to DAO contract`, e)
+    }
+  }
+
+  return await fetchCatalystNodesFromDAO()
+}
+
+export async function fecthCatalystRealms(nodesEndpoint: string | undefined): Promise<Candidate[]> {
```
nchamo:
```suggestion
export async function fetchCatalystRealms(nodesEndpoint: string | undefined): Promise<Candidate[]> {
```

nchamo (approved):
Great work!

caminoezequiel (approved):

PR #1598, opened 2020-11-11
chore: add 30 fps cap in settings

pravusjif:
**WHY?**
We observed that by keeping the rendering usage of the CPU "at bay" by capping the FPS to 30, more CPU is freed and thus the performance improves in several situations.

Profiling info done with 30FPS cap (the old way, Application.targetFrameRate = 30, causing Unity to use setTimeout instead of the better RequestAnimationFrame()):
https://docs.google.com/document/d/12Fb3jwPv6-dv1rpCz2s2IkQ5uOQc6rWeBDfpYKxHPCY/edit#heading=h.qisid6wm6xt0

**WHAT?**
We implemented
* A tick scheduling algorithm for controlling the framerate of the RequestAnimationFrame().
* A toggle in the quality settings panel for this new 30 FPS cap (medium and low quality profiles set it to true)

**TEST**
You can test by toggling the 30 FPS cap toggle in the quality settings pannel
https://explorer.decentraland.zone/branch/test/fps-capped-30/index.html?ENV=org&LOCAL_COMMS


unity-client/Assets/Plugins/CustomRequestAnimationFrame/CustomRequestAnimationFrame.jspre
```diff
@@ -0,0 +1,19 @@
...
+    if (Browser.nextRAF === 0) {
+      Browser.nextRAF = now + (1000/targetFPS);
+    } else {
+      while (now + 2 >= Browser.nextRAF) { // fudge a little, to avoid timer jitter causing us to do lots of delay:0
+        Browser.nextRAF += (1000/targetFPS);
+      }
+    }
+
+    var delay = Math.max(Browser.nextRAF - now, 0);
+    setTimeout(rafCallback, delay);
```
pablitar: Warning: We shouldn't use setTimeout for frame execution.

See here: https://stackoverflow.com/questions/38709923/why-is-requestanimationframe-better-than-setinterval-or-settimeout.

If we want to cap requestAnimationFrame we should skip frames manually to produce the desired frame rate.

See here: https://stackoverflow.com/questions/19764018/controlling-fps-with-requestanimationframe (I suggest looking at all the answers. Not just the first one).

BrianAmadori: If we are going to match 30 fps, maybe the best way to do this is to skip every other rAF so the intermediate calls do nothing.

I think that's is what's being referred in the Unity thread when talking about "decimating calls".

pravusjif: I'll be fiddling a bit more with this today, I think i misread Unity's approach and they only use the "fake RAF" that has the setTimeout only the first time, or something like that (I interpreted this was used all the time, as my logs showed that, but probably because i had to modify the code a lot and something went wrong in the middle, and that's why I unified the background RAF that Pablo had implemented with this new override).

My objective is to have the browser native RAF at a frequency we can control (30FPS for this test, and should be able to set that framerate dynamically) in the .jspre file. And re-add the background RAF that uses setTimeout instead of the browser native RAF.

Let me try a few more things later today

menduz: I think we should handle the framerate in unity directly. Unity has something called cooperative scheduling, it is the formal name of what is happening behind the "coroutines". It works by setting an execution quota every frame, and then continuing with important jobs (like actually rendering) once either the quota is complete or the "coroutines" are done.

By modifying the framerate in unity you may be modifying (_increasing_) the quota, therefore, affecting positively the CPU processing power per frame. 
In the contrary, by manually changing the external timer, you have the same effect as executing something else in the same thread (emulating another CPU intensive task), which has the effect of a non-preemptive scheduling schema, and therefore, it is not an optimization. It is instead, idle time in the CPU

menduz: The comment may be hard to digest. We can schedule a call to better explain the concepts

pravusjif: After fighting the whole day with this I finally achieved the decimation to get 30FPS using RequestAnimationFrame instead of setTimeout.

Check out the description of this PR as I updated it with the pertinent info, you can also check out the new build.

And of course, let's talk about that other approach tomorrow.

BrianAmadori: @menduz I am quite sceptic of that quota handling you are mentioning. Coroutines just work by a simple nested `IEnumerator`  loop technique, and I don't think they throttle the processing of the `MoveNext()`. Already revised many many custom coroutine loop implementations and even implemented one myself, we even have a custom coroutine implementation that supports exceptions already in the project. 

In fact, all of our throttling code (and UnityGLTF and GLTFast throttling code) is laid under the assumption that the coroutines loop aren't being throttled automatically. So as far as I checked in all these implementations, there's no evidence of an internal quota. The Unity's inner coroutine loop will just iterate on all available `MoveNext()` calls once until a `yield return;` is reached before ending the loop.

This said, I think we should compare the current rAF based implementation against the `setTimeout` one, because I just tested it and it seems to perform a bit worse. So in this I'm giving you the right hand @menduz, just wary about the justification. 

In the previous implementation I noted that we got almost permanent 30 fps, and in the current build I can see a bit more of 25-30 fps oscillation. Maybe an anomaly but it'd be safe to check.

Edit to add some docs and even more context (sorry if this turns out as TL;DR): 

This is how the unity loop works under the hood: https://medium.com/@thebeardphantom/unity-2018-and-playerloop-5c46a12a677

Knowing this, we can see that a frame in Unity has **way** more going on that some coroutines. The loop systems responsible them are the ones that have `Delayed` in the name:
`ScriptRunDelayedFixedFrameRate`
`ScriptRunDelayedTasks`
`ScriptRunDelayedDynamicFrameRate`

Edit #2 to explain some technical debt that we are having that further evidences this built-in quota limiting is not happening, and also I'd take this opportunity to give more context (and maybe an ADR may come out of this):

We are currently having some coroutine concurrency issues, as we have a number of coroutines with independent throttling logic. This lack of a global throttling value is giving us hiccups in some situations, because the sum of all the `MoveNext()` ms in a single frame can exceed our frame target ms budget.

The proposed fix for this is to implement some kind of coroutine manager that give us that global quota that'd prevent hiccups. If Unity handled this internally, there'd no need for this.

Some of the coroutine-based systems that can technically cause this cost overflow when running on a single frame are:
- MessagingControllerManager
- AssetPromiseKeeper in all its flavors
- AssetBundleLoader
- ComponentUpdateHandler
- GLTFSceneImporter
- etc.

BrianAmadori: To add a final note, please let's be wary of getting too technical for this particular solution. We should go for what works regardless of the preached best practices.

menduz: Ok, cool, let's continue this on slack. FWIW:
 - Preemptive scheduling: it is what you think I said. 
 - [Cooperative scheduling](https://en.wikipedia.org/wiki/Cooperative_multitasking) (what i mentioned): it is specifically the `yield return` behavior. It is not specific to unity, probably you will find better results searching for .Net. By the quota I mean that not all the Update() run all the frames. Please validate that assumption.

unity-client/Assets/Plugins/CustomRequestAnimationFrame/CustomRequestAnimationFrame.jspre
```diff
@@ -0,0 +1,69 @@
...
+
+    function nonUnityRAFCallback(stamp) {
+      window.requestAnimationFrame(nonUnityRAFCallback); // ends up calling the normal RAF again
+    }
+
+    window.requestAnimationFrame = function (unityRAFCallback) { // Unity's rafCallback calls RAF inside
+      if (document.hidden) {
+        backgroundRAF(unityRAFCallback);
+      } else {
+        var rafId = __requestAnimationFrame(function(stamp) {
```
menduz: Please take into account that rAF is not only used by unity, it is a globally available function and if some other part of the code calls it, the flip-flop logic implemented here would be broken for both codes.

menduz (changes requested):
Based on my previous comment, we could think about something like this

```ts
function tick(highResolutionNow){
  if (highResolutionNow - prev > 1 / TARGET_FPS) {
    // call all the callbacks
    callbacks.forEach(cb => cb(highResolutionNow))
    callbacks.length = 0 
    prev = highResolutionNow
  } else {
    // do nothing
  }
  scheduleNextFrameTick()
}

function scheduleNextFrameTick(){ 
  if(document.hidden)
    setTimeout(() => tick(performance.now(), 1 / TARGET_FPS)
  else
    requestAnimationFrame(tick)
}

window.raf = function(cb) {
  callbacks.push(cb)
}


```

pablitar:
Are we losing the functionality of RAF executing in background? Could we check that doesn't produce a regression?

We can test if DCL loads in background which is the most important reason to have background RAF.

unity-client/Assets/Plugins/CustomRequestAnimationFrame/CustomRequestAnimationFrame.jspre
```diff
@@ -0,0 +1,44 @@
...
+
+      scheduleNext();
+    }
+
+    function scheduleNext() {
+      if (document.hidden) {
+        setTimeout(function() { tick(performance.now()); }, FRAME_MS);
+      } else {
+        originalRaf(tick);
+      }
```
menduz: @pablitar we are not loosing it

pablitar: Oops, sorry. Didn't see that.

unity-client/Assets/Plugins/RequestAnimationFrameBackground/RequestAnimationFrameBackground.jspre
```diff
@@ -1,41 +0,0 @@
...
-      } else {
-        var rafId = __requestAnimationFrame(function(stamp) {
-          rafCallback(stamp);
-          delete __CURRENT_RAFS[rafId];
-        });
-        __CURRENT_RAFS[rafId] = rafCallback;
-      }
-    };
-
-    function switchToBackground() {
```
pablitar, line 26: @menduz @pravusjif Still, I think we should test it thoroughly. I remember that it was trickier than just calling setTimeout when the document was hidden.

I may have been wrong and overcomplicated it. But if I recall correctly the problem was that if you switched to background in the exact moment a requestAnimationFrame was pending to be executed, it didn't execute until the focus came back to the tab, effectively pausing the tab in background (because it is the RAF callback that ends up calling RAF again).

The solution I came up was this subscription to visibility change, that canceled pending RAFs, and switched them to setTimeout.

I'll try to test if we are having the problem in this branch in a bit.

menduz: Hmm makes sense.. Upon your confirmation we will make the changes

sandrade-dcl (approved):
Reviewed and tested. Good work!

unity-client/Assets/Plugins/RequestAnimationFrameBackground/RequestAnimationFrameBackground.jspre
pablitar: Unfortunately, I can confirm the issue exists.

To reproduce it:
- Enter the explorer and authenticate to start the loading process
- Immediately change tab or minimize the browser (ensure the tab is hidden somehow)
- Wait. Normal loading shouldn't take more than a minute.
- Go back to the tab. You should see that the loading process is still going

You maaaay get lucky and change tabs in the right moment so that the change to setTimeout is successful. But 3/3 times I tested I could reproduce the issue.

If you need help implementing a fix for this let me know :)

menduz: Cool, thanks

pravusjif: Fuck, ok, good catch, tomorrow I'll re-implement that previous fix and add comments to clarify on this so that we don't remove it in the future as well

AjimenezDCL (approved):
Unity's side looks good. I'm approving but wait for some kernel confirmation

unity-client/Assets/Plugins/RequestAnimationFrameBackground/RequestAnimationFrameBackground.jspre
pravusjif: @pablitar can you confirm that with the last change the background running fix is back?

pablitar (approved):
LGTM! Great work!!!

unity-client/Assets/Plugins/RequestAnimationFrameBackground/RequestAnimationFrameBackground.jspre
pablitar: seems to work great!

pravusjif:
We have to check again if the background raf is working correctly

PR #1600, opened 2020-11-11
feat: graphic card warning feature

AjimenezDCL:
Before reviewing this PR bear in mind a couple of things.

This is a first iteration built to be extended once product decides the whole flow and visuals. In that regard, the feature is implemented as a HUD but with a mock view that calls the Notification system. This way we can just create a proper view in the future (once a decision has been made). If a notification was there to stay as a final UI a small refinement in the approach would be needed as well.

If you want to try this in editor you can either make the method `DCL.Interface.WebInterface.GetGraphicCard()` to return anything containing `intel` or you can just hardcode `GraphicCardWarningHUDController.IsIntegratedGraphicCard` to always `true`.

image


BrianAmadori (approved):
LGTM.

sandrade-dcl (approved):
Tested and reviewed! Looks good 👍

PR #1603, opened 2020-11-11
chore: change PIN_CATALYST for CATALYST

### LINKED ISSUE #1424: Support a new testing realm/server
As of now, working in preview/export mode has some limitations or differences with production:
* Multiplayer is not supported
* Asset bundles are not created or used
* Other scenes are not loaded

These differences can have a huge impact in the final user experience, so content creators are forced to push their changes to production and test them there. This is of course not ideal:
* Content creators lose the element of surprise
* Users may have a bad experience by interacting with a scene that is not really finished
* Content servers have to handle more load, since each deployment is propagated through all servers

A possible solution might be to have a separate Catalyst server to be used only for testing. We could:
* Deploy a new Catalyst server outside the DAO
* Configure it so that it uses a realm with the name `testing`
* Make sure it pushes to the Asset Bundle converter

If we go with this approach, then deployments made to this server would not be propagated to the other realms, but deployments made to other servers would also be published on this new testing server.

Then, on the explorer side, we would need to make sure that this realm is used only when the realm is specified in the url (ignore any caching). For this approach to work we would need to make some changes to the explorer, since only whitelisted & DAO-approved Catalysts are supported right now.

nchamo:
Solves https://github.com/decentraland/explorer/issues/1424

### DIFF
--- a/kernel/packages/config/index.ts
+++ b/kernel/packages/config/index.ts
@@ -143,7 +143,7 @@ export const NO_ASSET_BUNDLES = location.search.indexOf('NO_ASSET_BUNDLES') !== 
 export const WSS_ENABLED = qs.ws !== undefined
 export const FORCE_SEND_MESSAGE = location.search.indexOf('FORCE_SEND_MESSAGE') !== -1
 
-export const PIN_CATALYST = qs.PIN_CATALYST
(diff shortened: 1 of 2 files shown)


pablitar (approved):
LGTM!

PR #1604, opened 2020-11-11
fix: avatar walking animation stutter

### LINKED ISSUE #1536: Avatar walking animation stutter
When the FPS is not optimal and we play on third person mode, stutters on the walk/run animations can be observed. This is due to our blend tree implementation on the LegacyAnimationController. 

The proposed fix is to drop the blend tree implementation and just implement discrete walk/run states. 

This is currently a blocker for the footsteps sfx integration.

pbosio:
close #1536 
* removed use of blending trees for walking, running and idle animations
* use character controller's clamped delta time to avoid having a mismatch in the speed value


unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarAnimatorLegacy.cs
```diff
@@ -133,32 +153,28 @@ void State_Init(BlackBoard bb)
...
     void State_Ground(BlackBoard bb)
     {
-        float dt = Time.deltaTime;
-
-        animation[baseClipsIds.run].normalizedSpeed = bb.movementSpeed / dt * bb.runSpeedFactor;
-        animation[baseClipsIds.walk].normalizedSpeed = bb.movementSpeed / dt * bb.walkSpeedFactor;
-
-        float normalizedSpeed = bb.movementSpeed / dt / MAX_VELOCITY;
+        animation[baseClipsIds.run].normalizedSpeed = bb.movementSpeed / bb.deltaTime * bb.runSpeedFactor;
+        animation[baseClipsIds.walk].normalizedSpeed = bb.movementSpeed / bb.deltaTime * bb.walkSpeedFactor;
```
pravusjif, line 163: there's no risk of dividing by zero here? XD

pravusjif (approved):

sandrade-dcl (approved):
Tested and code reviewed.

There is only a thing that I think we should take into account and decide if we want it that way: Removing the blending in the avatar movement provoke that the animations endings are much more abrupt than before (I will share a video in Slack). From my point of view it looked better before than now, but it is something completelly subjective and it should be decided by someone (Product?).

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarAnimatorLegacy.cs
pbosio: It shouldn't unless someone force it to be zero :S
Added a check anyway, just in case, better be safe :P

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarAnimatorLegacy.cs
```diff
@@ -153,22 +153,28 @@ void State_Init(BlackBoard bb)
...
+
         animation[baseClipsIds.run].normalizedSpeed = bb.movementSpeed / bb.deltaTime * bb.runSpeedFactor;
         animation[baseClipsIds.walk].normalizedSpeed = bb.movementSpeed / bb.deltaTime * bb.walkSpeedFactor;
 
         float movementSpeed = bb.movementSpeed / bb.deltaTime;
 
         if (movementSpeed > runMinSpeed)
         {
-            animation.Play(baseClipsIds.run);
+            animation.CrossFade(baseClipsIds.run, RUN_TRANSITION_TIME);
```
BrianAmadori, line 169: Transition times still feel pretty abrupt. Specially the Walk -> Idle transition.

I see that the transition times are set to 0.01. We should try to aim for 0.1 or 0.15.

BrianAmadori (changes requested):

BrianAmadori (approved):
The only detail is that Walk/Run->Idle transitions now look a tad slow. But looks good overall!

PR #1605, opened 2020-11-11
fix: avatar draw call batch improvements

BrianAmadori:
### This PR adds:
- SRP Batching support to all avatar used shaders.
- Proper SRP friendly sorting for all avatar materials.

### Following modifications to the draw call sorting algorithm in `SRPBatchingHelper`:
- Force the following keyword to be always on: `_ALPHATEST_ON`  
- Force the following unused keyword to be always off: `VERTEX_COLOR_ON`
- Take shader name into account when sorting, so the eyes, mouth, wearables and the rest are grouped together.

### Misc changes:
- Remove unused instancing code from some shaders
- Renamed `DCL/Toon Shader` to `DCL/Toon Shader Legacy`

---
### Test instancing 7 avatars via `bots.create()` in genesis plaza spawn point:

Before (50 render opaques):
image

After (9 render opaques):
image


sandrade-dcl (approved):
Tested and it seems to work fine :)

About the code review, wait for have another 2 approves because I'm not very familiar with shaders programming.

AjimenezDCL (approved):

pravusjif (approved):

PR #1609, opened 2020-11-12
Chore: react updates

caminoezequiel:
# What? 
* Fix post-install script
* Fix discord link in navbar
* Changes hash version of static files, now it is a md5 of content
* Disabled Metamask on react when it's not available
* Disable WelcomeMessage for new user
* Fix passport to keep name and email
* Update readme with new port


pablitar (approved):
LGTM!

PR #1614, opened 2020-11-13
TEST: testing scroll event

caminoezequiel:
# What? 
Scroll event doesn't work well in some  browser.


caminoezequiel:
replaced with #1615

PR #1621, opened 2020-11-13
feat: renderer + shadows + animations lod culling

BrianAmadori:
This PR adds a new culling logic for renderers, shadows and animations.

### Status:
- [x] Basic feature implementation
- [x] Settings integration
- [x] Tests implementation
- [x] QA

#### What this feature does? 
- Hide rendering objects that occupy small viewport real state (detail objects) 
- Disable very small and unneeded shadows
- Enable/disable animation culling for skinned renderers and animation components. 

#### Added settings to the quality settings panel:
- A feature toggle
- A slider to control the size threshold of the detail objects to be culled. Bigger values will mean better performance gain, but more and bigger objects will be culled.

#### Implementation details
- The culling system will update when its dirty flag is toggled. If the player moves or a new renderer is added, a check will be performed.
- Renderers and animators will be iterated using time-slicing with a time budget of 4 ms per frame.
- Scene objects are gathered using the infamous `Object.FindObjectsOfType`. Generally is a bad practice, but here we are betting on a positive tradeoff between CPU and GPU.
- Size checks are determined using viewport size, not scaling. This way we can guarantee that we cull the correct objects.
- To disable renderers, we use [Renderer.forceRenderingOff](https://docs.unity3d.com/ScriptReference/Renderer-forceRenderingOff.html). This way we can ensure that if SDK code disables a renderer, it stays disabled even if its un-culled.
- When the settings slider is moved, we are lerping between two sets of culling size parameters. How those parameters work is going to be explained below.

### Next steps
The objects pop-in can bother in terms of visual quality, so a industry standard dithering fade-in effect can be added when the objects are culled in/out.

#### Renderers
Emissive renderers have different size values than normal renderers because the bloom effect makes them take
more screen real estate. Also, culling bright objects is way more noticeable than other types of objects, i.e. small leaves of grass.

- Is nearest than a given distance from player? -> should be visible
- Player is inside the renderer bounds? -> should be visible
- Is emissive and bigger than very small size? -> should be visible
- Is opaque and bigger than small size? -> should be visible

#### Shadows
A [big portion of our frame benchmarking](https://docs.google.com/spreadsheets/d/1tW38rNXWaXTnkxKUDtybqgEutiwPquaDlrwrCFUhTSo/edit#gid=0) is taken by shadow map rendering. Here we are going to try to turn off small meshes that are far enough, or small enough meshes that will take less than X texels of the shadowmap.

The full list of checks is:
 - Is less than a given distance from player? -> should have shadow
 - Is bigger than a medium size? -> should have shadow
 - shadowmap shadow size is less than X texels? -> shouldn't have shadow

#### Skinned renderers
Here we are going to try and limit animation processing and mesh skinning for meshes that are very small, or far away. This should have impact in animation heavy scenes like soho plaza.

- If they are far, `updateWhenOffscreen` is off. If they are near, they will always be updated. This is to prevent glitches with animated meshes on the edge of screen.
- `SkinnedMeshRenderers` have different size threshold values than the rest of the renderers. We want animated meshes to be culled less often than normal renderers.

### Benchmarking Results
Taken with `Intel(R) UHD Graphics 630, OpenGL 4.5 core + i7-9750H 2.6Ghz`, mid settings, match 1080p.

#### Sample 1
Before (master):
Genesis plaza #1, 22 FPS (45~ ms)
image

After (this PR, slider at 50%)
Genesis plaza #1, 29-30 FPS (34~ ms, 32% gain)
image

#### Sample 2
Before (master):
Genesis plaza #2, 22 FPS (45~ ms)
image

After (this PR, slider at 50%):
Genesis plaza #2, 28-30 FPS (34~ ms, 32% gain)
image

#### Sample 3
Before (master):
Genesis plaza #3, 25-29 FPS (37~ ms)
image

After (this PR, slider at 50%):
Genesis plaza #3, 29-31 FPS (33~ ms, 12% gain)
image

#### Sample 4
Before (master):
Momus park, 16 FPS (62.5~ ms)
image

After (this PR, slider at 50%):
Momus park, 25 FPS (40~ ms, 56% gain)
image

#### Sample 5
Before (master):
Tomb chaser, 15-17 FPS (66~ ms)
image

After (this PR, slider at 50%):
Tomb chaser, 15-17 FPS (tbd)
(image wip)


unity-client/Assets/Rendering/Culling/ICullingController.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace DCL.Rendering
+{
+    public interface ICullingController
```
AjimenezDCL: Unless the interface is gonna be implemented multiple times, for DI, I rather have it on top of the main class that implements it. Just a matter of opinions though, it looks good this way as well.

unity-client/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -0,0 +1,505 @@
...
+        {
+            return new CullingController(
+                GraphicsSettings.currentRenderPipeline as UniversalRenderPipelineAsset,
+                new CullingControllerSettings()
+            );
+        }
+
+        public CullingController(UniversalRenderPipelineAsset urpAsset, CullingControllerSettings settings)
+        {
+            sceneObjects = new CullingObjectsTracker();
```
AjimenezDCL: Could it be useful to have an `ICullingObjectsTracker` injection as well? It would  allow us to mock it for tests or have heavier/lighter tracking alternatives.

unity-client/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -0,0 +1,505 @@
...
+        /// <returns>IEnumerator to be yielded.</returns>
+        internal IEnumerator ProcessProfile(CullingControllerProfile profile)
+        {
+            Renderer[] renderers = null;
+
+            if (profile == settings.rendererProfile)
+                renderers = sceneObjects.renderers;
+            else
+                renderers = sceneObjects.skinnedRenderers;
+
```
AjimenezDCL, line 145: Would you mind to elaborate a bit on this piece?

unity-client/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -0,0 +1,505 @@
...
+                if (shouldBeVisible && !shouldHaveShadow && !shadowlessRenderers.Contains(r))
+                    shadowlessRenderers.Add(r);
+
+                var skmr = r as SkinnedMeshRenderer;
+
+                if (skmr != null)
+                {
+                    skmr.updateWhenOffscreen = ShouldUpdateSkinnedWhenOffscreen(settings, distance);
+                }
+#if UNITY_EDITOR
```
AjimenezDCL, line 195: We can use pattern matching here. `if(r is SkinnedMeshRenderer skmr)`

unity-client/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -0,0 +1,505 @@
...
+            RaiseDataReport();
+            profiles = new List<CullingControllerProfile> {settings.rendererProfile, settings.skinnedRendererProfile};
+            CommonScriptableObjects.rendererState.OnChange += (current, previous) => SetDirty();
+
+            while (true)
+            {
+                if (!CommonScriptableObjects.rendererState.Get())
+                {
+                    timeBudgetCount = 0;
+                    yield return null;
```
AjimenezDCL: If renderer is disabled we could wait a 1 second to avoid this while looping too much. Not a big deal because if the renderer is disabled we dont care that much about performance though.

unity-client/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -0,0 +1,505 @@
...
+        /// </summary>
+        private void RaiseDataReport()
+        {
+            if (OnDataReport == null)
+                return;
+
+            int rendererCount = (sceneObjects.renderers?.Length ?? 0) + (sceneObjects.skinnedRenderers?.Length ?? 0);
+
+            OnDataReport.Invoke(rendererCount, hiddenRenderers.Count, shadowlessRenderers.Count);
+        }
```
sandrade-dcl: I have noticed that `OnDataReport` event is not being listened in any place. Have you created this method for future uses?

unity-client/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -0,0 +1,505 @@
...
+            this.settings = settings;
+            profiles = new List<CullingControllerProfile> {settings.rendererProfile, settings.skinnedRendererProfile};
+            SetDirty();
+        }
+
+        /// <summary>
+        /// Get current settings copy. If you need to modify it, you must set them via SetSettings afterwards.
+        /// </summary>
+        /// <returns>Current settings object copy.</returns>
+        public CullingControllerSettings GetSettings()
```
sandrade-dcl: I would call this function as something like `GetSettingsCopy()` only to clarify in the own name that the returned info is a not the original entity but a copy. It would help to understand better the code in the consumer side. What do you think?

unity-client/Assets/Rendering/Culling/CullingControllerProfile.cs
```diff
@@ -0,0 +1,41 @@
...
+        public float shadowMapProjectionSizeThreshold;
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="p1"></param>
+        /// <param name="p2"></param>
+        /// <param name="t"></param>
+        /// <returns></returns>
+        public static CullingControllerProfile Lerp(CullingControllerProfile p1, CullingControllerProfile p2, float t)
```
sandrade-dcl: You forgot to write this documentation :)

unity-client/Assets/Rendering/Culling/CullingControllerProfile.cs
```diff
@@ -0,0 +1,41 @@
+﻿using UnityEngine;
+
+namespace DCL.Rendering
+{
+    [System.Serializable]
+    public class CullingControllerProfile
```
sandrade-dcl, line 10: Could we add a little description about what is this class?

unity-client/Assets/Rendering/Culling/CullingControllerSettings.cs
```diff
@@ -0,0 +1,42 @@
+﻿namespace DCL.Rendering
+{
+    [System.Serializable]
+    public class CullingControllerSettings
```
sandrade-dcl, line 7: Could we add a little description about what is this class?

unity-client/Assets/Rendering/Culling/CullingControllerSettingsData.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace DCL.Rendering
+{
+    [CreateAssetMenu(menuName = "Create CullingControllerSettingsData", fileName = "CullingControllerSettingsData", order = 0)]
+    public class CullingControllerSettingsData : ScriptableObject
+    {
+        [Header("Renderer Profile")] public CullingControllerProfile rendererProfileMin;
+        public CullingControllerProfile rendererProfileMax;
+
+        [Header("Skinned Renderer Profile")] public CullingControllerProfile skinnedRendererProfileMin;
+        public CullingControllerProfile skinnedRendererProfileMax;
```
AjimenezDCL:
```suggestion
        [Header("Renderer Profile")]
        public CullingControllerProfile rendererProfileMin;
        public CullingControllerProfile rendererProfileMax;

        [Header("Skinned Renderer Profile")]
        public CullingControllerProfile skinnedRendererProfileMin;
        public CullingControllerProfile skinnedRendererProfileMax;
```

unity-client/Assets/Rendering/Culling/CullingObjectsTracker.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Collections;
+using System.Linq;
+using UnityEngine;
+
+namespace DCL.Rendering
+{
+    public class CullingObjectsTracker
```
sandrade-dcl: Could we add a little description about what is this class?

unity-client/Assets/Rendering/Culling/CullingControllerUtils.cs
```diff
@@ -0,0 +1,38 @@
+﻿using UnityEngine;
+
+namespace DCL.Rendering
+{
+    public static class CullingControllerUtils
+    {
+        public static void DrawBounds(Bounds b, Color color, float delay = 0)
```
AjimenezDCL, line 142: How do you use this debug method? Should we wrap it in a `UNITY_EDITOR` directive?

sandrade-dcl (changes requested):
A very clear code and easy to follow :) Good work man! I only left some little requests.

AjimenezDCL (approved):

AjimenezDCL:
What a clean PR, great job. Also the feature works flawlessly. I was a bit concerned of having items popping on/off in the background but it's barely noticeable even in the most aggresive setup.

Just to confirm, the time budget feature within will prevent us from switching the bottleneck from the GPU to the CPU, right?

unity-client/Assets/Rendering/Culling/CullingObjectsTracker.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        /// <summary>
+        /// If the dirty flag is true, this coroutine will re-populate all the tracked objects.
+        /// </summary>
+        public IEnumerator PopulateRenderersList()
+        {
+            if (!dirty)
+                yield break;
+
+            renderers = Object.FindObjectsOfType<Renderer>()
```
pravusjif, line 39: In the future we can probably track every renderer we create at that point in time and we would avoid calling FindObjectsOfType repeatedly.

unity-client/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -0,0 +1,551 @@
...
+                if (timeBudgetCount > maxTimeBudget)
+                {
+                    timeBudgetCount = 0;
+                    yield return null;
+                }
+
+                Renderer r = renderers[i];
+
+                if (r == null)
+                    continue;
```
pravusjif, line 157: We could also skip renderers that aren't inside the frustrum culling area using the `Renderer.isVisible` property, what do you think?

unity-client/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -0,0 +1,551 @@
...
+                }
+
+                bool shouldCheck = objectsTracker.IsDirty();
+
+                yield return objectsTracker.PopulateRenderersList();
+
+                Vector3 playerPosition = CommonScriptableObjects.playerUnityPosition;
+
+                if (Vector3.Distance(playerPosition, lastPlayerPos) > 1.0f)
```
pravusjif: We could also subscribe to the `CommonScriptableObjects.playerUnityPosition` `OnChange` event and raise a dirty flag  or something when that event is triggered, instead of comparing vectors here.

unity-client/Assets/Rendering/Culling/CullingController.cs
BrianAmadori: That's correct.

unity-client/Assets/Rendering/Culling/CullingControllerProfile.cs
BrianAmadori: Done

unity-client/Assets/Rendering/Culling/CullingController.cs
BrianAmadori: Yep, needed to add it for test mocking. Done.

unity-client/Assets/Rendering/Culling/CullingController.cs
BrianAmadori: Done!

unity-client/Assets/Rendering/Culling/CullingControllerUtils.cs
BrianAmadori: The caller is already wrapping it in `UNITY_EDITOR`.

unity-client/Assets/Rendering/Culling/CullingObjectsTracker.cs
BrianAmadori: Absolutely. For the future, I think we may encapsulate all of this "renderer tracking" behaviour in a some kind of data store that fits the `MeshesInfo` specs, and have the `SceneBoundariesChecker` query that instead of `MeshesInfo`.

unity-client/Assets/Rendering/Culling/CullingController.cs
BrianAmadori: Changed to be fully event-driven

unity-client/Assets/Rendering/Culling/CullingController.cs
BrianAmadori: Added comment

unity-client/Assets/Rendering/Culling/ICullingController.cs
BrianAmadori: done

unity-client/Assets/Rendering/Culling/CullingController.cs
BrianAmadori: now all the logic is fully event driven

unity-client/Assets/Rendering/Culling/CullingController.cs
BrianAmadori: Few issues to go for this approach: 

* `Renderer.isVisible` uses bounding box culling. This wouldn't work with skinned meshes because the animations tend to go off the bounds. 

* If we have out of frustum renderers and they are not processed, we have to mark them as dirty as they go into the frustum to process them again. Ideally, we must do this without frame delay and with a O(1) operation when setting the dirtyness. This kind of lazy updating logic tends to be buggy and usually requires a lot of test coverage.

So, going for this one will be a bit costly. At this stage, I'd prefer to go with the current approach and try new optimizations later.

unity-client/Assets/Rendering/Culling/CullingControllerSettings.cs
BrianAmadori: Done

unity-client/Assets/Rendering/Culling/CullingControllerProfile.cs
BrianAmadori: Done

unity-client/Assets/Rendering/Culling/CullingController.cs
BrianAmadori: Changed it with a bit of reluctance because it breaks the symmetry of the `Get`/`Set` pairs. Maybe its just better to make the settings object a struct and be done with it. Going to think about this improvement for a later time though.

unity-client/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -150,31 +204,11 @@ internal IEnumerator ProcessProfile(CullingControllerProfile profile)
...
             {
-                if (!CommonScriptableObjects.rendererState.Get())
-                {
-                    timeBudgetCount = 0;
-                    yield return null;
-                    continue;
-                }
-
-                bool shouldCheck = sceneObjects.dirty;
+                bool shouldCheck = objectsTracker.IsDirty() | playerPositionDirty;
```
AjimenezDCL: Just out of curiosity, is there any benefit of using bitwise vs logical operations with booleans?

AjimenezDCL (approved):

sandrade-dcl (approved):

AjimenezDCL:
Feature works flawlessly. Maybe it's too noticeable for emissive objects, but I dont find it too disruptive.

I found a small issue that you can take down in another PR. When an object is instantiated but is going to be culled, it pops-out for a second and then hide itself. I think it should appear invisible from the beginning.

unity-client/Assets/Rendering/Culling/CullingController.cs
BrianAmadori: In this case not really, this was just a style choice. It may be a bad idea because I'm not sure if the expression is short-circuited (i.e. stopped being evaluated if the left side is true). I guess It's better to change it just in case.

pravusjif (approved):

PR #1630, opened 2020-11-16
feat: Adds fake names for guest user

caminoezequiel:
# What? 
We add a name pool for guest user


kernel/packages/shared/profiles/transformations/processServerProfile.ts
```diff
@@ -36,7 +37,7 @@ export function calculateDisplayName(userId: string, profile: any): string {
     return `${profile.unclaimedName}#${userId.slice(-4)}`
   }
 
-  return 'Guest-' + userId.substr(2, 6)
+  return createFakeName(userId)
```
pablitar, line 40: The behaviour with this is that every time you enter you'll have a different name. Is that the expected behaviour?

caminoezequiel: Yes, It is because it's only for guest user from deep link

PR #1640, opened 2020-11-17
fix: Make the GPU warning waits for tutorial ends

sandrade-dcl:
Make the GPU warning waits for tutorial ends (if it is actived).

### DIFF
--- a/kernel/packages/entryPoints/website.ts
+++ b/kernel/packages/entryPoints/website.ts
@@ -122,12 +122,12 @@ namespace webApp {
         i.ConfigureHUDElement(HUDElementID.PROFILE_HUD, { active: true, visible: true })
         i.ConfigureHUDElement(HUDElementID.USERS_AROUND_LIST_HUD, { active: voiceChatEnabled, visible: false })
         i.ConfigureHUDElement(HUDElementID.FRIENDS, { active: identity.hasConnectedWeb3, visible: false })
-        i.ConfigureHUDElement(HUDElementID.GRAPHIC_CARD_WARNING, { active: true, visible: true })
 
         EnsureProfile(identity.address)
           .then((profile) => {
             i.ConfigureEmailPrompt(profile.tutorialStep)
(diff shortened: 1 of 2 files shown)


AjimenezDCL (approved):

kernel/packages/entryPoints/website.ts
```diff
@@ -122,12 +122,12 @@ namespace webApp {
         i.ConfigureHUDElement(HUDElementID.PROFILE_HUD, { active: true, visible: true })
         i.ConfigureHUDElement(HUDElementID.USERS_AROUND_LIST_HUD, { active: voiceChatEnabled, visible: false })
         i.ConfigureHUDElement(HUDElementID.FRIENDS, { active: identity.hasConnectedWeb3, visible: false })
-        i.ConfigureHUDElement(HUDElementID.GRAPHIC_CARD_WARNING, { active: true, visible: true })
 
         EnsureProfile(identity.address)
           .then((profile) => {
             i.ConfigureEmailPrompt(profile.tutorialStep)
             i.ConfigureTutorial(profile.tutorialStep, HAS_INITIAL_POSITION_MARK)
+            i.ConfigureHUDElement(HUDElementID.GRAPHIC_CARD_WARNING, { active: true, visible: true })
```
pravusjif, line 130: this wouldn't make the warning to appear always by default?

AjimenezDCL: The GPU check is done in Unity (through the jslib) this just enables the whole process. For the next release we will disable it but right now we want to test that it's indeed, working

pravusjif (approved):

kernel/packages/entryPoints/website.ts
pravusjif: awsome, thx

sandrade-dcl: @pravusjif, as @AjimenezDCL said, this is only for initialize the process but will not make the warning to appear always by default 👍

PR #1643, opened 2020-11-17
fix: misplaced fps UI position (behind the taskbar)

BrianAmadori:


AjimenezDCL (approved):

eordano (approved):
Still a couple pixels below from perfect but good enough. Approved
image

pravusjif (approved):

PR #1663, opened 2020-11-19
fix: Make the avatar edition more fluid in the SignUp flow

sandrade-dcl:
Make the avatar edition more fluid in the SignUp flow.


BrianAmadori (approved):
LGTM!

AjimenezDCL (approved):

PR #1666, opened 2020-11-19
feat: set HQ images for static picture frames on demand

pbosio:
Try to get a high quality image for static (non gif) picture frames when player is looking at it and restore to low res image when player stop looking at it.
Gifs not supported yet.


unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeHQImageHandler.cs
```diff
@@ -0,0 +1,159 @@
...
+        if (config.controller.collider == null)
+            return;
+
+        if (!isPlayerNear)
+            return;
+
+        isCameraInFront = Vector3.Dot(config.controller.transform.forward, config.controller.transform.position - camera.transform.position)
+            > config.nftConfig.hqImgInFrontDotProdMinValue;
+
+        if (config.nftConfig.verbose)
```
AjimenezDCL, line 57: What are the benefits of having this `verbose` flag in the config instead of a const? I dont think this will change at runtime, does it?

AjimenezDCL (approved):
Looks good!

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeHQImageHandler.cs
pbosio: I did change it in runtime while debugging :P
but you are right, I will make it a constant

PR #1676, opened 2020-11-20
chore: handling errors

caminoezequiel:
# What?
send uncaught errors to segment with error info


kernel/packages/shared/store/store.ts
```diff
@@ -4,11 +4,21 @@ import { reducers } from './rootReducer'
...
 import { DEBUG_REDUX } from '../../config'
+import { ReportFatalError } from '../loading/ReportFatalError'
 
 export let store: Store<RootState>
 
 export const buildStore = () => {
-  const sagaMiddleware = createSagaMiddleware()
+  const sagaMiddleware = createSagaMiddleware({
+    onError: (error: Error, { sagaStack }: { sagaStack: string }) => {
+      ReportFatalError(error.message as any, {
```
pablitar, line 14: This would make any uncaught error in sagas to bring down the client, right?

I think I agree with this. We had errors on sagas that were fatal and wouldn't bring down the client.

pablitar (approved):

eordano (approved):

Issue #1622, opened 2020-11-14
Setting a new VideoTexture doesn't unload a previous VideoTexture's soundtrack


cfravel:
The issue is that if you pick a different movie and start it, the soundtracks from previously-played movies also play, and pausing the new movie doesn't pause the old soundtracks.

see https://gitlab.com/dcl-tech/bugdemo-multiplemovies for a scene that demonstrates this issue

----- Steps to reproduce

Load the demo scene in SDK preview
Select a movie

Leftmost cube selects previous movie from catalog
Rightmost cube selects next movie catalog
There are 3 movies in the catalog

Press the 2nd cube from the left (Play)

(Play resets any currently-playing movie, loads ths selected movie into the screen, and plays it)

// movie plays
Select a different movie
Press Play (2nd cube)

// expected: original movie stops, new movie plays
// actual: original movie visual stops, new movie plays, but so does the sound from the original movie.

(The 3rd button is a pause button.)
Press Pause when a subsequent mnovie is playing

// expected: vidoe and sound stop
// actual: sound from previously played movies continues playing

cfravel:
see https://gitlab.com/dcl-tech/bugdemo-multiplemovies fpr a scene that demonstrates this issue

cfravel:
I have determined that if I use videotexture.playing=false instead of videotexture.reset() both the video and audio are freed up to load and start a different video without hearing the old audio.

I recommend that whatever you are doing about audio during videotextureplaying=false be done in reset()
Perhaps your reset just needs to call playing=false.

In the meantime, i'm going to simply use playing=false instead of reset().

Not sure what reset() is _supposed_ to do, but it seems it is missing something, compared to what i would expect from the docs.

PR #1685, opened 2020-11-23
fix: video player reset

pravusjif:
WHY?
the documentation on the video player at https://docs.decentraland.org/development-guide/video-playing/ states that the `reset()` method should stop the video. We are not doing this and so there is undesired behaviour like playing audio from a previous video as stated in #1622 

WHAT?
removed playing the video when resetting and the bug is fixed.


AjimenezDCL (approved):

sandrade-dcl (approved):

pbosio:
I think this should be done from the component and not from the jslib
`packages/decentraland-ecs/src/decentraland/Components.ts` (L:1052)
```
  reset() {
    this.seekTime(0)
    + this.playing = false
  }
```
and probably change the order at `Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs`
to make sure that `model.seek` field is checked before `model.playing`
cause I don't know if we want component's `seekTime` method to stop the video also and that will happen with the approach used in ths PR
Also, removing `play()` from jslib will leave the component field `playing` as `true` so there is no way to know from the sdk side that the video is actually stopped

pbosio (changes requested):
comment leaved before :P

pbosio (approved):
LGTM

PR #1691, opened 2020-11-24
feat: add name component

D4rWiNSS:
# What? 
This PR adds a new component called "NAME" for entities

# Why? 
The Builder In World needs to give a descriptive name so we decide to create a new component to represent this value.

This is created as a shared component for performance reasons. In order to avoid mono behaviors being created and added for each name component we used shared component.

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
@@ -1,0 +1,34 @@
+using DCL;
+using DCL.Components;
+using DCL.Controllers;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class DCLName : BaseDisposable
+{
+    [System.Serializable]
+    public class Model
+    {
+        public string value;
+    }
+
+    public Model model;
+
+    public DCLName(ParcelScene scene) : base(scene)
+    {
+        model = new Model();
+    }
+
+    public override IEnumerator ApplyChanges(string newJson)
+    {
+        Model newModel = SceneController.i.SafeFromJson<Model>(newJson);
+        if(newModel.value != model.value)
+        {
+            model = newModel;
+            RaiseOnAppliedChanges();
+        }
+    
+        yield return null;
+    }
+}
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
@@ -5,6 +5,10 @@ using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
 
+/// <summary>
+/// This component is a descriptive name of the Entity. In the BuilderInWorld you can give an entity a descriptive name through the entity list.
+/// Builder in World send a message to kernel to change the value of this component in order to assign a descriptive name
+/// </summary>
 public class DCLName : BaseDisposable
 {
     [System.Serializable]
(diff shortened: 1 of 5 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
```diff
@@ -0,0 +1,34 @@
+using DCL;
+using DCL.Components;
+using DCL.Controllers;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class DCLName : BaseDisposable
```
BrianAmadori, line 13: Can we add a xml comment explaning why this component exists? I think a newcomer will scratch his head on this one.

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
```diff
@@ -0,0 +1,34 @@
...
+    public override IEnumerator ApplyChanges(string newJson)
+    {
+        Model newModel = SceneController.i.SafeFromJson<Model>(newJson);
+        if(newModel.value != model.value)
+        {
+            model = newModel;
+            RaiseOnAppliedChanges();
+        }
+    
+        yield return null;
```
BrianAmadori:
```suggestion
        return null;
```
We don't need to skip a frame here, the `return null;` is properly handled from outside to cover this case and don't skip a frame after this coroutine execution when null is returned.

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
```diff
@@ -0,0 +1,38 @@
...
+using DCL.Controllers;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+/// <summary>
+/// This component is a descriptive name of the Entity. In the BuilderInWorld you can give an entity a descriptive name through the entity list.
+/// Builder in World send a message to kernel to change the value of this component in order to assign a descriptive name
+/// </summary>
+public class DCLName : BaseDisposable
```
AjimenezDCL, line 13: Why is this component `Shared (Disposable)`? Can a same name be shared accross different entities?

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public Model model;
+
+    public DCLName(ParcelScene scene) : base(scene)
+    {
+        model = new Model();
+    }
+
+    public override IEnumerator ApplyChanges(string newJson)
```
AjimenezDCL, line 28: I would add something like:
```
#if UNITY_EDITOR
foreach (DecentralandEntity decentralandEntity in this.attachedEntities)
                decentralandEntity.meshRootGameObject.name += model.value;
#endif
```
If we are gonna have named entities, we might as well be able to find them in the hierarchy =).

Be aware that the code above is just a sample, probably you will have to ensure that multiple `ApplyChanges` are processed correctly and the format could be way better.

AjimenezDCL (approved):
Approving with a question and a small suggestion!

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
BrianAmadori: No, but the alternative involves forcing a name `MonoBehaviour`, and that is worse. We'll have to improve the design of the components in the future -_-

BrianAmadori (approved):

PR #1692, opened 2020-11-24
fix: add position validation to emote trigger

nchamo:
In https://github.com/decentraland/explorer/pull/1489 we added the ability to trigger emotes from the SDK. However, we forgot to add a validation that makes sure that the user is inside the scene when the animation is triggered. That is what we are adding here


pablitar (approved):
LGTM!

PR #1695, opened 2020-11-24
chore: autosettings capped fps

AjimenezDCL:
Autosettings now is compatible with capped fps.

This includes:

- A small refactor. Now AutoSettings implements a HumbleObject pattern that allows us to have different autosettings implementations (one for uncapped framerate and another one for capped framerate).

- New implementation for capped settings based on margins. If the average FPS of the last 5 seconds is above the 90% (26fps) of the target capped FPS, the system increases quality.  If the average FPS is between the 80% (24fps) and 90%, the quality stays. In any other case the quality is decreased. We increment the quality agressively, 2 steps when needed vs 1 step decrement.

- Removed the evaluator entity, it ended up having so little of responsibility that made sense to include it in the new controllers themselves.

- Update of the test suite.


sandrade-dcl (approved):
Good work! 👍

AjimenezDCL:
~Something I forgot to mention, you wont see the autoquality option in the settings yet. I'm waiting for the settings prefab to be merged from other PR to avoid conflicts.~

~I will let you all know when that's ncluded so you can test the feature if you want to. You can still review the PR since the code wont change.~

Nevermind, just pushed a temporary settings hud so you can test it.

pravusjif (approved):
awsome!

AjimenezDCL:
Feature works flawlessly. Maybe it's too noticeable for emissive objects, but I dont find it too disruptive.

I found a small issue that you can take down in another PR. When an object is instantiated but is going to be culled, it pop-out for a second and then hide itself. I think it should appear invisible from the beginning.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Settings/SettingsControllers/AutoQualitySettingsComponent.cs
```diff
@@ -17,24 +17,44 @@ public class AutoQualitySettingsController : MonoBehaviour
...
+
+    internal void OnQualitySettingsChanged(QualitySettings settings)
+    {
+        if (fpsCapped == settings.fpsCap)
             return;
-        autoQualityEnabled.OnChange -= SetAutoSettings;
+
+        fpsCapped = settings.fpsCap;
+        if (settings.fpsCap)
+            controller = new AutoQualityCappedFPSController(30, currentQualityIndex, qualitySettings);
```
sandrade-dcl: It would be fine to extract the value 30 as a constant with an identifying name.

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Settings/SettingsControllers/AutoQualitySettingsComponent.cs
AjimenezDCL: Oopsie, I forgot about that. Done

pravusjif (approved):

PR #1616, opened 2020-11-13
chore: adjust mouse sensitivity

AjimenezDCL:
#1002

PR #1701, opened 2020-11-25
fix: reduce acceleration

eordano:
# What? 
Drop the acceleration for the mouse

# Why? 
image


BrianAmadori:
This is already being taken care of at https://github.com/decentraland/explorer/pull/1616. Please let's try to centralize the efforts here.

eordano:
> This is already being taken care of at #1616. Please let's try to centralize the efforts here.

The purpose of this is to trigger a build so we can share different links with users testing different settings

AjimenezDCL:
We could use this: https://play.decentraland.zone/branch/chore/1002_mouse_sensitivity_test/index.html for that

eordano:
> We could use this: https://play.decentraland.zone/branch/chore/1002_mouse_sensitivity_test/index.html for that

Yes, this is useful to find good settings, but in order to do end-user testing we need actual links. Asking the user to use those sliders contaminates the tests.

PR #1703, opened 2020-11-25
fix: network warning on preview when user doesn't use web3

### LINKED ISSUE #1700: Local preview scenes attempt to connect to Infura
When I run a preview of a scene locally, it seems that a request is sent out to Infura while the scene is loading, for example

wss://mainnet.infura.io/ws/v3/074a68d50a7c4e6cb46aec204a50cbf0

There are three problems with this:

- Yesterday there was some downtime for that service and I couldn't run any previews at all, even scenes that didn't use the blockchain in any way. This is an avoidable risk, since there are no benefits from connecting like this, I still see the default avatar & use a random user id.
- The loading time of the preview is slowed down because of this request
- If I don't include the &ENABLE_WEB3 URL parameter, I'm not asked to sign on Metamask but I get the warning that I'm connected to ehtereum on mainnet on the top of the screen. I haven't confirmed this, but I fear I might be connected using a fake wallet address, and the risk there is that if I run transactions like this, I could be accidentally burning tokens sending them to non-existent accounts.

caminoezequiel:


nchamo (approved):
Looks good!

Just to confirm, this PR only intends to fix the third issue described in https://github.com/decentraland/explorer/issues/1700, right?

PR #1705, opened 2020-11-25
fix: Beginners Guide

2fd:
# What? 

- add links to the `Beginners Guide` section


caminoezequiel (approved):

website/src/components/auth/BeginnersGuide.tsx
```diff
@@ -124,7 +124,7 @@ export const BeginnersGuide: React.FC<React.HTMLAttributes<HTMLDivElement>> = (
             Decentraland on-chain assets.
         </p>
           <p>
-            Access the Marketplace at{" "}
```
caminoezequiel: @2fd I think we need to use it when a tag is on the next line, also we should apply Prettier

caminoezequiel (changes requested):

caminoezequiel (approved):

PR #1711, opened 2020-11-26
refactor: separate builder in world favorites and quick bar functionality

D4rWiNSS:
# What? 
The favorites and the quick bar of the catalog were mixed in strange behavior. This PR solves that making it a cleaner code

# Why? 
Since favorites and quick bar are no longer related, we can separate them.


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/FavoritesController.cs
```diff
@@ -0,0 +1,43 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class FavoritesController : MonoBehaviour
```
BrianAmadori: `FavoritesController` and `QuickbarController` are having a mixed responsibility of view and controller. If we are going to refactor this to match our project-wide HUD patterns, we should find a way to decouple the UI logic from the presentation logic. 

This decoupling is important to make better tests. If we decouple the view, then the logic of the HUD element is easier to test because it wouldn't be a `MonoBehavior`.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/QuickBarController.cs
```diff
@@ -0,0 +1,118 @@
...
+    public RawImage[] shortcutsImgs;
+    public Canvas generalCanvas;
+    public CatalogGroupListView catalogGroupListView;
+
+    public event System.Action OnResumeInput;
+    public event System.Action OnStopInput;
+
+    public event System.Action<SceneObject> OnSceneObjectSelected;
+
+    List<SceneObject> quickBarShortcutsSceneObjects = new List<SceneObject>() { null, null, null, null, null, null, null, null, null };
```
AjimenezDCL: I think we should use an array for this since it's going to have a fixed size

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/FavoritesController.cs
AjimenezDCL: I agree with Brian. The controller should just handle the array of items, update their `isFavorite` state and call an update for the view (or generate events for the view to listen, I like both approaches).

It would be perfect if the View is either mockable or not even needed, so we can test the controller in an editor test suite without the need of playmode.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/QuickBarController.cs
```diff
@@ -0,0 +1,118 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.EventSystems;
+using UnityEngine.UI;
+
+public class QuickBarController : MonoBehaviour
+{
+    public RawImage[] shortcutsImgs;
```
AjimenezDCL: This array points me to something missing in the architecture. I believe an entity `QuickBarItemSlot` is asking for existance. 

```
public class QuickBarItemSlot : Monobehaviour
{
    [SerializeField] private RawImage image;
    private SceneObject sceneObject;

    public void SetSceneObject(SceneObject object);
    public void SetEmpty();
}
```

Therefore `QuickBarController` would have an array of these `QuickBarItemSlot` and they would be self sufficient and more testable.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/QuickBarController.cs
```diff
@@ -0,0 +1,118 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.EventSystems;
+using UnityEngine.UI;
+
+public class QuickBarController : MonoBehaviour
```
AjimenezDCL: This class suffers from the same responsibility mix than `FavoritesController`.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/QuickBarController.cs
AjimenezDCL: Obviously this array and such would go into a `View` class. The Controller would just need the `SceneObjects`

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/QuickBarController.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            if (sceneObjectIteration == null)
+            {
+                index = cont;
+                break;
+            }
+            cont++;
+        }
+        return index;
+    }
```
AjimenezDCL:
```suggestion
    int FindEmptyShortcutSlot()
    {
        for(int i = 0; i < quickBarShortcutsSceneObjects.Length; i++)
        {
            if (quickBarShortcutsSceneObjects[i] == null)
                return i;                
        }
        return -1;
    }
```

I think that's more readable and using -1 as a control value for "not found" is more reliable. You can also implement it like:
```
 int FindEmptyShortcutSlot(out index)
    {
        index = -1;
        for(int i = 0; i < quickBarShortcutsSceneObjects.Length; i++)
        {
            if (quickBarShortcutsSceneObjects[i] == null)
            {
                index = i;
                return true;                
            }
        }
        return false;
    }
```
If you prefer it more explicit

AjimenezDCL (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/QuickBarController.cs
D4rWiNSS: The problem is that the SceneObject doesn't contain the texture, but I will create it with the texture as parameter

AjimenezDCL: Sure but the `SceneObject` can build the thumbnail URL right? Therefore we can just use AssetPromise within this new `QuickBarItemSlot` to get the texture. You dont have to worry about said texture being downloaded multiple times because the AssetPromise system will take care of that and cache stuff.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/QuickBarController.cs
```diff
@@ -0,0 +1,36 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class QuickBarController 
+{
+    public event System.Action<SceneObject> OnSceneObjectSelected;
+
+    SceneObject[] quickBarShortcutsSceneObjects = new SceneObject[9];
```
AjimenezDCL: I would have this in a const. It would also allow us to have a really tiny integration test that compares that the view and the controller are in sync with the same amount of slots.

AjimenezDCL (approved):

AjimenezDCL:
Looking good!

I added a small suggestion about moving the amount of entries of the quickbar slots to a const.

Also take a look at this https://github.com/decentraland/explorer/pull/1711#discussion_r531502239 and check if it makes sense. It can be done in another PR if you agree with it.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/QuickBarController.cs
D4rWiNSS: I already have the texture when the scene object is added, so I think it is more performance-wise to just assign the reference instead of asking the AssetPromise if it has the reference (It will always have the texture because if there is no texture you can't use the scene object, however, it is not the most beautiful from an architecture point of view

AjimenezDCL: You dont really ask the AssetPromise, just request the asset by an id (in this case the URL) and the system will take care of everything for you. The cool thing about the AssetPromise (and the AssetLibrary built behind) is that you dont really care whether the asset is cached, has (or is in the process) to be downloaded. Also requesting a cached asset is O(0) since it's in a dictionary.

I agree passing the Texture is more perfomant, but I think the gain is marginally and the benefits outscale it.
The `QuickBarItemSlot` would be more independent and self-sufficent, it would make it more testable and the code would be easier to read/extend since you would decrease the amount of params to track down when following the architecture. 

In any case, this is something we can tackle down later, it is good as it is right now.

PR #1717, opened 2020-11-26
fix: cache and dispose gifs properly

pbosio:
* Cache gif data and texture in unity to avoid downloading the same gif more than once
* Delete textures stored in browser when picture frame is disposed and no other picture frame is using them.
* Cancel gif download when picture frame is disposed and gif is still downloading or enqueued to be downloaded
* Use kernel config to set if we handle gifs using javascript worker or unity
* Typed gif processor and worker

NOTE: Some stuffs will be refactored later, like removing the caching responsibility from the bridge class


AjimenezDCL (approved):

pravusjif:
places with GIFs to test:
Mintbase building has the most animated picture frames right now: https://play.decentraland.zone/branch/fix/gif-dispose-properly/index.html?ENV=org&position=52%2C101

SuperRare building: https://play.decentraland.zone/branch/fix/gif-dispose-properly/index.html?ENV=org&position=55%2C94

SOHO plaza art gallery: https://play.decentraland.zone/branch/fix/gif-dispose-properly/index.html?ENV=org&position=-85%2C10&realm=fenrir-amber

kernel/packages/gif-processor/worker.ts
```diff
@@ -1,85 +1,125 @@
...
-      framesAsArrayBuffer.push(processedImageData.data.buffer)
-    }
-
-    self.postMessage({
-      arrayBufferFrames: framesAsArrayBuffer,
-      width: finalWidth,
-      height: finalHeight,
-      delays: frameDelays,
-      id: e.data.id,
-    }, framesAsArrayBuffer)
```
pravusjif, line 79: We should still pass this second parameter of the `self.postMessage()` (framesAsArrayBuffer) if we want the data to be passed "by reference".

Can you try adding that to your `self.postMessage()` call? if it doesn't break anything then it works, otherwise I guess it's fine

pravusjif (approved):

kernel/packages/gif-processor/processor.ts
```diff
@@ -69,14 +108,7 @@ export class GIFProcessor {
...
-        this.GLctx.TEXTURE_2D,
-        0,
-        this.GLctx.RGBA,
-        this.GLctx.RGBA,
-        this.GLctx.UNSIGNED_BYTE,
-        image
-      )
+      this.GLctx.texImage2D(this.GLctx.TEXTURE_2D, 0, this.GLctx.RGBA, this.GLctx.RGBA, this.GLctx.UNSIGNED_BYTE, image)
     } else {
       this.GLctx.texImage2D(
```
BrianAmadori, line 113: In WebGL 2, `texSubImage2D` has better performance. Not sure why we didn't do it like this in the first place.

BrianAmadori (approved):
Didn't test it but looks good enough code-wise. Having the GIF cancellation/requests flow spread on kernel and unity is somewhat confusing, so I'm hoping the refactor we talked about mitigates this somehow.

Maybe we can add a bit more of documentation and comments to make this flow more palatable?

unity-client/Assets/Scripts/MainScripts/DCL/Bridges/GIFProcessingBridge/GIFProcessingBridge.cs
```diff
@@ -37,10 +31,12 @@ void Awake()
             }
 
             i = this;
+
+            KernelConfig.i.EnsureConfigInitialized().Then(config => { jsGIFProcessingEnabled = config.gifSupported; });
```
BrianAmadori, line 35: I think we should spread the knowledge of this promise class so we make more uses of this pattern in C#. 

Usually, we use callbacks for this behaviour. Having this pattern here and there breaks the project code consistency. I like it but I think we should aim to standardize it or something.

kernel/packages/gif-processor/processor.ts
pbosio: cause we where in a hurry to implement it cause we had like 3 days before an event or something like that :P

pbosio:
> Didn't test it but looks good enough code-wise. Having the GIF cancellation/requests flow spread on kernel and unity is somewhat confusing, so I'm hoping the refactor we talked about mitigates this somehow.
> 
> Maybe we can add a bit more of documentation and comments to make this flow more palatable?

Yeap, it's tricky, kernel has to maintain a list of the gl textures and unity has to maintain a list for the textures created to contain those gl textures...
I'm open to any suggestion for ways to improve that... It would be great if we can keep commenting here in this PR even after is merged so I can come back here when the moment for refactoring comes :)

pravusjif (approved):
tested after adding the arrayAsBuffer parameter and still works great!

PR #1723, opened 2020-11-27
release: 2020-11-27

caminoezequiel:
release: 2020-11-27

Changelog (cherry pick):
* feat: adds Dapper wallet (#1715)


BrianAmadori (approved):
Non-dapper flow ran pretty flawless to me. Don't have a dapper wallet so I can't review that flow easily, but approving it from the non-dapper side.

PR #1726, opened 2020-11-27
fix: generated meshes on GLTF import were not optimized

BrianAmadori:
For some reason, GLTFSceneImporter was not optimizing the meshes at all. 

The mesh optimization is done to shuffle the mesh attributes in memory and make them as small as possible. This can improve rendering times, because we are taking advantage of the GPU cache. 

The optimization will only be done in Unity editor so its only reflected in asset bundle conversion until further notice.

### DIFF
--- a/unity-client/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
+++ b/unity-client/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
@@ -2131,13 +2131,16 @@ namespace UnityGLTF
                 yield return YieldOnTimeout();
             }
 
+// Disable it in runtime so this optimization only takes place in
+// asset bundle converter time.
+#if UNITY_EDITOR
             mesh.Optimize();
 
             if (ShouldYieldOnTimeout())
             {
                 yield return YieldOnTimeout();
             }
-
+#endif
             if (!KeepCPUCopyOfMesh)
             {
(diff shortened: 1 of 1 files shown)


unity-client/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -2126,6 +2126,18 @@ protected IEnumerator ConstructUnityMesh(MeshConstructionData meshConstructionDa
                 mesh.RecalculateNormals();
             }
 
+            if (ShouldYieldOnTimeout())
+            {
+                yield return YieldOnTimeout();
+            }
+
+            mesh.Optimize();
```
pravusjif, line 2137: I'm not sure how this may impact performance when done in runtime, if that hits performance somehow, maybe we should check somehow if this is running in the world or at "ab-conversion-time" ?

pravusjif (approved):

unity-client/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
BrianAmadori: Just changed it to take place only in editor time until we can benchmark it better.

pbosio (approved):

PR #1727, opened 2020-11-27
feat: set HQ images for gif picture frames on demand

pbosio:
Try to get a high quality image for gif picture frames when player is looking at it and restore to low res image when player stop looking at it.


unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeAssetsType.cs
```diff
@@ -0,0 +1,180 @@
...
+internal interface INFTAsset : IDisposable
+{
+    bool isGif { get; }
+    bool isHQ { get; }
+    DCL.ITexture previewAsset { get; }
+    DCL.ITexture hqAsset { get; }
+    Action<Texture2D> UpdateTextureCallback { set; }
+    void FetchAndSetHQAsset(string url, Action onSuccess, Action onFail);
+    void RestorePreviewAsset();
+}
```
AjimenezDCL: I love the interface implementation!

Nevertheless I would remove `isGif` from here. I dont think it strictly belongs to the idea of an NFTAsset and the only use of this getter is to determine the resolution of the texture to fetch. We could replace it by an `int resolution { get; }` that would end up returning whatever the NTFShapeConfig has or a more complex behaviour if needed.

AjimenezDCL (changes requested):
Great job, it looks really clean with the factory + interface approach.

I left a small request and also I would suggest to create some tests for the factory and any new functionality added.

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeAssetsType.cs
pbosio: yeah, you are totally right, at the beginning I was using that boolean for other stuffs to avoid checking type in runtime but later I polish the code better and forgot about it :P

AjimenezDCL (approved):

PR #1729, opened 2020-11-30
fix: enable shader variants prewarm

pravusjif:
* Created shader variants collection files for all shaders (not used as it takes forever to compile when the explorer starts)
* Created shader variants collection with a few common selected shaders and variants, added a prewarm call for that when the application starts
* Disabled 'optimize mesh data' in player settings as it's been reported that it causes problems when switching materials and shaders in runtime, besides making builds take longer when compiling shader variant files.

If the shaders prewarm time is unacceptable we'll have to implemenet a new event from kernel telling the renderer when the loading will begin (or the play button was clicked) and move the shaders warmup to that point


sandrade-dcl (changes requested):
Testing the build I noticed that some materials in my avatar are being drawed as blank:

image

sandrade-dcl (approved):

sandrade-dcl:
> Testing the build I noticed that some materials in my avatar are being drawed as blank:
> 
> image

Solved!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -45,6 +45,9 @@ void Awake()
             Debug.unityLogger.logEnabled = false;
 #endif
 
+            // Warmup some shader variants
+            Resources.Load<ShaderVariantCollection>("ShaderVariantCollections/shaderVariants-selected").WarmUp();
```
BrianAmadori: If this call is blocking, this means the kernel messages can start pouring in before the rest of the code in this `Awake()` call, leading to issues. We should move this call at the very least to the end of this method.

BrianAmadori (changes requested):
The compiling delay is quite noticeable. In my case, the freeze could be observed *after* the rendering was activated.

Options over the top of my head:
- Try to narrow the variants in the shadervariants file even further
- Try to ensure the prewarm happens before the rendering is activated, or block the rendering ack until the prewarm is finished
- Try to bundle the variants inside some asset bundle file and see if it takes less time.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -222,99 +225,99 @@ public bool ProcessMessage(MessagingBus.QueuedSceneMessage_Scene msgObject, out
...
-                    }
+                            break;
+                        }
                     case MessagingTypes.ENTITY_REPARENT:
-                    {
-                        if (msgObject.payload is Protocol.SetEntityParent payload)
-                            scene.SetEntityParent(payload.entityId, payload.parentId);
+                        {
+                            if (msgObject.payload is Protocol.SetEntityParent payload)
+                                scene.SetEntityParent(payload.entityId, payload.parentId);
```
AjimenezDCL, line 199: Any reason for this extra indentation?

AjimenezDCL (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
pravusjif: the auto formatter file we have in the repo should be responsible for that, probably we have 2 different ones, one for VSCode and another for Rider? or where do you guys store your formatting rules?

AjimenezDCL: We dont have a common one, we should though. @BrianAmadori

pravusjif:
@BrianAmadori I moved the warmup next to the pools warmup in `SceneController.Start()` and now I don't notice the stall at all, check it out

BrianAmadori (approved):
Tested it in incognito + cache disabled and the stall is no more. Marking as approved!

PR #1732, opened 2020-11-30
chore: reduce GLTF tests duration

pravusjif:
replaced Lantern GLB to reduce tests duration

* Removed Lantern glb usage (almost 10mb)
* Added Trunk and PalmTree usage (26kb and 32kb)

Local testing showed a 14 seconds gain


pravusjif:
from 401 to 409 =( no sense at all.

PR #1737, opened 2020-12-01
feat: add components

D4rWiNSS:
# What? 
This PR adds 2 components for the BuilderInWorld: LockOnEdit and VisibleOnEdit

# Why? 
The BuilderInWorld needs a way to communicate with kernel regarding the visibility and lock status of the entities during edit mode to maintain the same state across all clients.


unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLLockedOnEdit.cs
```diff
@@ -0,0 +1,38 @@
...
+/// <summary>
+/// This component describes the lock status of the Entity in the builder in world.
+/// Builder in World send a message to kernel to change the value of this component in order to lock/unlock it
+/// </summary>
+public class DCLLockedOnEdit : BaseDisposable
+{
+    [System.Serializable]
+    public class Model
+    {
+        public bool value;
```
AjimenezDCL: I think this var name is a constrain of having a common pipeline, but if no... Can we rename it to something more meaningful? Something like `locked`

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLVisibleOnEdit.cs
```diff
@@ -0,0 +1,39 @@
...
+/// <summary>
+/// This component describes the visibility of the Entity in the builder in world.
+/// Builder in World send a message to kernel to change the value of this component in order to show/hide it
+/// </summary>
+public class DCLVisibleOnEdit : BaseDisposable
+{
+    [System.Serializable]
+    public class Model
+    {
+        public bool value;
```
AjimenezDCL: Same as before, if possible, rename to something more useful.

AjimenezDCL (approved):

AjimenezDCL:
Approved with two small requests. Not sure if they are possible though!

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLLockedOnEdit.cs
D4rWiNSS: The problem is that the kernel part name is "value". Since we get the model and convert it to JSON and send to kernel, we can't change the name without changing it on the kernel side

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLVisibleOnEdit.cs
D4rWiNSS: Same as before. The problem is that the kernel part name is "value". Since we get the model and convert it to JSON and send to kernel, we can't change the name without changing it on the kernel side

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLLockedOnEdit.cs
AjimenezDCL: Yeah, I assumed that. It should be quite trivial, shouldnt it?

AjimenezDCL: @nchamo what do you think?

nchamo: Yeah, we can change it without any problems 😄 

Go ahead

pbosio: Why not just change the `Model` class name? that would be easier and more meaningful in my opinion... like `IsLockedModel` or something like that

pbosio (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLLockedOnEdit.cs
AjimenezDCL: The model class is nested within `DCLLockedOnEdit` class so the full name is actually `DCLLockedOnEdit .Model` which is quite meaningful. We also have this structure across every component in our project so I wouldn't change that without changing the rest.

D4rWiNSS: Renamed to `isLocked` @nchamo  let me know when you change the name so I can push the changes!

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLVisibleOnEdit.cs
D4rWiNSS: Renamed to `isVisible ` @nchamo let me know when you change the name so I can push the changes!

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLLockedOnEdit.cs
nchamo: You can go ahead with the changes, there is nothing to be done on Kernel's side. Same with the other component

sandrade-dcl (approved):

PR #1738, opened 2020-12-01
feat: use lambda for explore (unity side)

pbosio:
Test with lambda: https://explorer.decentraland.zone/branch/feat/explore-lambda-unity/index.html?ENV=org&CATALYST=peer-testing.decentraland.org&realm=testing-amber
Test without lambda: https://explorer.decentraland.zone/branch/feat/explore-lambda-unity/index.html?ENV=org


pravusjif (approved):

BrianAmadori (commented):
Loving the huge code removal!

image

I'm having this issue trying to test right now.

Can you explain a bit what this PR is about and why you managed to remove the crowd manager from the unity side?

pbosio:
> Loving the huge code removal!
> 
> [IMAGE WAS HERE!]
> 
> I'm having this issue trying to test right now.
> 
> Can you explain a bit what this PR is about and why you managed to remove the crowd manager from the unity side?

We should ask @nchamo about the issue in the screenshot. I seems to be something with that testing catalyst...
I removed some handlers cause pretty much every scene information is passed by the lambda.
So there was no need any more for complex handlers to wait for kernel reporting scene's info (`MinimapSceneInfo`) and after removing that dependency I was able to simplify a lot of the code

AjimenezDCL (approved):

pravusjif (approved):
In my test right now, this optimized version took half the time to load the Explore scenes (5~ seconds vs 10~)

PR #1740, opened 2020-12-01
feat: use deployed scene image as thumbnail

pbosio:
Be able to use a image deployed with the scene as a thumbnail.


kernel/packages/shared/atlas/reducer.ts
```diff
@@ -25,6 +25,7 @@ const MAP_SCENE_DATA_INITIAL_STATE: MapSceneData = {
   estateId: 0,
   sceneJsonData: undefined,
   alreadyReported: false,
+  contents: [],
```
nchamo, line 28: This is my only concern. I've seen some scenes that had up to 1000 files. Would you think it might be an overkill to store all content files here, even though we only want one? Is there a downside to having all of them here?

@pablitar , what do you think?

pbosio: > This is my only concern. I've seen some scenes that had up to 1000 files. Would you think it might be an overkill to store all content files here, even though we only want one? Is there a downside to having all of them here?
> 
> @pablitar , what do you think?

I don't know how that would behave, I'm not doing a copy of the content just using a reference `    mapScene.contents = land.mappingsResponse.contents` but not sure if js will box/unbox it or just be a reference :P

pablitar: Are these long lived objects? If so, how many of them are populated at a normal session? Also, how many scenes do have thousands of files?

Before doing any optimization, maybe making the calculation of an hypothetical worst case would be nice.

pbosio: In average scenario we maybe have like 30/40 scenes or so with the explore without using the lambda + 15 scenes around every position of the mouse while using the map

pbosio: but I think that is filled with every scene in the world

pablitar: Can we do a quick and dirty memory test comparing this with master? Maybe load DCL in a heavy scene (like Genesis Plaza), and then load explore, and see the memory footprint of the tab (with shift+esc in chrome).

If we see no significant difference, then I wouldn't worry much.

pbosio: I'm getting the same memory footprint value in both master and this branch ~1.5gb

pablitar: That's good enough for me!

pablitar (approved):

nchamo (approved):
If @pablitar is ok with it, then I'm super ok with it!

PR #1742, opened 2020-12-01
feat: report realms info to renderer (kernel side)

pbosio:


pablitar:
Couldn't we make it so Unity can ask kernel for realms information when it needs it?

kernel/packages/unity-interface/realmsForRenderer.ts
```diff
@@ -0,0 +1,60 @@
...
+    fetchAndReportRealmsInfo()
+
+    reportInterval()
+  }
+}
+
+function reportInterval() {
+  setInterval(() => {
+    fetchAndReportRealmsInfo()
+    reportInterval()
```
pablitar: If you use `setInterval` you don't need to loop. It is done automatically. But I'd suggest replacing the `setInterval` for a `setTimeout`

kernel/packages/unity-interface/realmsForRenderer.ts
```diff
@@ -0,0 +1,60 @@
...
+    })
+
+    fetchAndReportRealmsInfo()
+
+    reportInterval()
+  }
+}
+
+function reportInterval() {
+  setInterval(() => {
```
pablitar: I'd suggest replacing the `setInterval` for a `setTimeout`, because then you can do this:
```typescript
function reportInterval() {
  setTimeout(async () => {
    await fetchAndReportRealmsInfo()
    reportInterval()
  }, REPORT_INTERVAL)
}
```
Ensuring that even if the fetch takes time, there is no chance that two updates are executed at the same time

(you'd need to make `fetchAndReportRealmsInfo` async for this to work, but that should be relatively easy)

kernel/packages/unity-interface/realmsForRenderer.ts
pbosio: upsi... I was thinking `setTimeout` and typed `setInterval` LOL Thx

kernel/packages/unity-interface/realmsForRenderer.ts
pbosio: > Couldn't we make it so Unity can ask kernel for realms information when it needs it?

This is probably going to be used by an UI that the user open... fetching in that moment will probably take time and the UI will be refreshed while the user has it open... I think it will look kind of ugly... but I can use that approach you suggested and if it look too bad we can change it back later

pbosio: Yeah, my bad... clearly my first intention was doing that function async... that's why it has an `await` there, I don't know what happen in the middle and I forgot 😂

pablitar: No need. If the information doesn't need to be up to date, then I think I'm OK with the regular update.

pbosio: Yeah, I think a 2 minutes interval is OK, we don't need to have the "current" precise info for now

caminoezequiel (commented):
LGTM!

pablitar (approved):

Issue #1741, opened 2020-12-01
Draw distance working weird in @Next


nearnshaw:
This simple example scene shows a cube that moves back and forth far away

[https://app.zenhub.com/files/220994476/ff2debdd-13e9-486a-8d5a-b6f3681482f2/download](https://app.zenhub.com/files/220994476/ff2debdd-13e9-48T6a-8d5a-b6f3681482f2/download)

In the @latest ecs version it works well, when it moves around 3 parcels away I stop seeing it, but it appears again when it comes back
But in the @next ecs version it works in a glitchy way. It sometimes works fine, it sometimes only reappears when it's really close, and it sometimes doesn't come back at all and I don't see it again till I move closer, even if it's coming right next to me.

These videos show both scenarios:

[https://app.zenhub.com/files/220994476/12e6592f-c73a-4930-97fa-1aaa7c97e490/download](https://app.zenhub.com/files/220994476/12e6592f-c73a-4930-97fa-1aaa7c97e490/download)
[https://app.zenhub.com/files/220994476/c32d2c30-7631-4f86-aad6-84cf7e89802b/download](https://app.zenhub.com/files/220994476/c32d2c30-7631-4f86-aad6-84cf7e89802b/download)

PR #1750, opened 2020-12-02
fix: culling controller mesh update issue

BrianAmadori:
addresses #1741

The fix just ensures `MarkDirty` is called when it should.

`MarkDirty` was missing on:
- Entity movement
- Pool `Get`

Also, dirtyness is now decoupled. Now has independent values for refilling the renderer list and updating the culling for the existing list.


unity-client/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -260,6 +260,8 @@ internal void SetCullingForRenderer(Renderer r, bool shouldBeVisible, bool shoul
             {
                 if (r.forceRenderingOff != !shouldBeVisible)
                     r.forceRenderingOff = !shouldBeVisible;
+
+                Debug.Log($"-- Asd: {r.name} ... {r.forceRenderingOff}");
```
AjimenezDCL: Remove debug

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -488,50 +485,49 @@ public void SetEntityParent(string entityId, string parentId)
...
-                    // In this case, the entity will attached to the first person camera
-                    // On first person mode, the entity will rotate with the camera. On third person mode, the entity will rotate with the avatar
-                    me.SetParent(DCLCharacterController.i.firstPersonCameraReference);
-                    SceneController.i.boundariesChecker.AddPersistent(me);
-                    Environment.i.physicsSyncController.MarkDirty();
+                    SceneController.i.boundariesChecker.RemoveEntityToBeChecked(me);
                 }
-                else if (parentId == "AvatarEntityReference" || parentId == "AvatarPositionEntityReference") // AvatarPositionEntityReference is for compatibility purposes
+
+                if (parentId == "0")
```
AjimenezDCL, line 512: Just curious about this comparison to "0". Mind to elaborate?

AjimenezDCL (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
BrianAmadori: When a `SetEntityParent` message comes with `0` as payload, this means that the entity should be attached to the "root" of the scene. Like setting `parent == null` in Unity.

unity-client/Assets/Rendering/Culling/CullingController.cs
BrianAmadori: Done.

pravusjif (approved):
the build was "buttery smooth"

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
AjimenezDCL: Maybe we can have a comment or have a static method `ParentIdIsRoot` that does the comparison.

PR #1753, opened 2020-12-02
feat: first implementation of DataStore

pbosio:


unity-client/Assets/Scripts/MainScripts/DCL/DataStorage/DataStorage.cs
```diff
@@ -0,0 +1,7 @@
+﻿using Variables.RealmsInfo;
+
+public static class DataStorage
```
BrianAmadori:
```suggestion
public static class DataStore
```

Maybe `DataStore` is a better option, as kernel uses the same word for its store.

BrianAmadori (approved):
Aside of the name comment, I see this as a good starting point for the store.

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/DataStore/DataStore.cs
```diff
@@ -0,0 +1,7 @@
+﻿using Variables.RealmsInfo;
+
+public static class DataStore
+{
+    static public readonly CurrentRealmVariable playerRealm = new CurrentRealmVariable();
+    static public readonly RealmsVariable realmsInfo = new RealmsVariable();
```
pravusjif, line 6: Since `CurrentRealmModel` also uses `layer` and `serverName` (`RealmsVariable` has 3 other variables that maybe we are not using today but it could come in handy to quickly know the url or users count of the current realm, the `playerRealm`), I think we could avoid having `CurrentRealmVariable` and use `RealmsVariable` (that I'd call `RealmVariable`) for both of these variables, transforming `realmsInfo` into an array of `RealmVariable`.

What do you think?

unity-client/Assets/Scripts/MainScripts/DCL/DataStore/Variables/BaseVariable.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+
+public delegate void Change<T>(T current, T previous);
+
+public class BaseVariable<T> : IBaseVariable<T>, IEquatable<T>
```
AjimenezDCL, line 6: How come we dont have compilation issues with this? `BaseVariable<T>` type will be duplicated right?

AjimenezDCL (approved):
It looks good! Added a small question.

sandrade-dcl (approved):
Looks good! As soon as you merge it into `master`, I will use this new `BaseVariable` for some stuff in my "New Settings" branch :)

unity-client/Assets/Scripts/MainScripts/DCL/DataStore/DataStore.cs
pbosio: making playerRealm a `RealmsVariable` will make it trigger Change event if a new user gets into that realm (that would me annoying) and I think most of the time we will just need an event for when player changes realm and the name of the realm...
If we need more info about the current realm then we have several ways to do that like creating a property that iterates through `realmsInfo` or creating a new variable for that

PR #1763, opened 2020-12-04
fix: early rendering activation when scenes take too long to load

pravusjif:
### **Problem**
When scenes take a while to load we are activating the rendering early and the user can appear in the middle of nowhere and see the scenes appearing later.

### **Implementation**

* Unity side: Added rendering lock/unlock when sorting scenes (called every time a batch of scenes is sent to the renderer and when we position the character) according to having or not a current scene (since Kernel sends the character position just-before activating the rendering, we don't know the current scene until then).

* Kernel side: corrected renderer activation (and loading screen toggling)

### **TEST**
1. Open a **new** incognito browser window
2. access the test URL. Around Zone 64,-65 we have many scenes, it takes a while to load and that's key to reproduce the issue

**master (bugged):** https://explorer.decentraland.zone/?ENV=zone&position=64%2C-65

**prod (bugged):** https://explorer.decentraland.org/?ENV=zone&position=64%2C-65

**this PR (fixed):** https://explorer.decentraland.zone/branch/fix/RenderingActivationInFirstGuestFlow/index.html?ENV=zone&position=64%2C-65

----------------------------------------
**Tested by Pravus**
* Rendering activation in sign in flow -> and rendering activation after teleporting
* Rendering activation in sign up flow
* Rendering activation in preview mode


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -652,9 +652,20 @@ private void SortScenesByDistance()
...
+            }
+            else
+            {
+                // 1. Set current scene id
                 CommonScriptableObjects.sceneID.Set(currentSceneId);
 
+                // 2. Attempt to remove SceneController's lock on rendering
+                CommonScriptableObjects.rendererState.RemoveLock(this);
+            }
+
```
pravusjif, line 668: this is the only relevant change in this file, the rest was auto-formatted

BrianAmadori:
Before I start the review, we made sure this doesn't break preview mode?

pravusjif:
> Before I start the review, we made sure this doesn't break preview mode?

Rendering activation in Preview Mode still works fine with this update, as stated in the description

pbosio (approved):
Tested. Working. Thanks!

BrianAmadori (approved):
Looks like its fixed! However, two things:

1. The responsibility of `SortScenesByDistance` is mutating into a big ball of mud. Having to rely on this lock mechanism smells bad and doesn't give reliance on which race conditions we could stumble upon later. So we should meet and try to define the current state of the lifecycle and see what can be done to fix this from the root.

2. The welcome notification was shown twice. Again, this is a minor issue and can be fixed in another PR.

pravusjif:
> Looks like its fixed! However, two things:
> 
> 1. The responsibility of `SortScenesByDistance` is mutating into a big ball of mud. Having to rely on this lock mechanism smells bad and doesn't give reliance on which race conditions we could stumble upon later. So we should meet and try to define the current state of the lifecycle and see what can be done to fix this from the root.
> 2. The welcome notification was shown twice. Again, this is a minor issue and can be fixed in another PR.

Regarding (1) yeah actually I think the whole initialization communication with Kernel should be revised and probably refactored into something simpler, for example right now Kernel:
**1.** Sends the batch of scenes to be loaded, and keeps the character position to itself 
**2.** After resolving the "current scene" spawn point (or that's what I undersood backtracing from [[here]](https://github.com/decentraland/explorer/blob/154dfaced35f4d5fe357585a98c6e87a78c66053/kernel/packages/unity-interface/dcl.ts#L182) to the `parcelSceneManager.ts`), just then, Kernel knows the character position
**3.** Kernel sends the character position to Unity
**4.** Unity receives the Character Position, applies it and sorts the scenes according to distance to the player (since now Unity knows the Character position), and at this point Unity sets the first "current scene"
**5.** If the current scene is not ready, the rendering activation is locked until the scene finishes loading

As you can see, Kernel knows the current scene from the beginning and Unity could lock the renderer knowing that, just an example.

PR #1768, opened 2020-12-04
feat: support disabling voice chat on scene json

nchamo:
## The feature
We want to allow scenes to disable the voice chat feature on their parcels. For that, we will be adding some new configuration to the `scene.json`. This new config can be used like this:
```
featureToggles: {
    voiceChat: 'disabled',
}
```

### New Observables
We added two new `Observable`s:
* `parcelObservable`: calls listeners when the user enters a new parcel
* `sceneObservable`: call listeners when the user enters a new scene

### Updates to Voice Chat
Now, when the client gets a new voice chat recording, it is discarded if the scene disabled the feature. The same happens if the client wants to start a new recording.

Finally, when the user changes scene, Kernel sends a message to the renderer, to let it know whether the scene supports the feature or not

### Minor refactor
We also replaced some uses of the `positionObservable` to the new `parcelObservable`


pablitar (approved):
LGTM! Great work!

pbosio (approved):
:ATR:

PR #1776, opened 2020-12-07
refactor: move world state to environment

BrianAmadori:
This refactor will get one step closer to removing the monolith status of `SceneController` class.

* Created new `WorldState` class
* Move `loadedScenes` and all scene-related selectors out of `SceneController` and into `WorldState`
* Move profiling events from SceneController to `ProfilingEvents` 

Next steps will involve removing more `SceneController` systems to `Environment` in order to be able to reduce `MainScripts` assembly dependencies and finish our initialization layer.


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Environment.cs
```diff
@@ -52,6 +54,7 @@ public void Initialize(IMessageProcessHandler messageHandler, ISceneHandler scen
             worldBlockersController = WorldBlockersController.CreateWithDefaultDependencies(sceneHandler, DCLCharacterController.i.characterPosition);
             parcelScenesCleaner.Start();
             cullingController.Start();
+            worldState = new WorldState();
```
pbosio: Why an instance of WorldState is created both inside Environment constructor and inside Initialize method?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Environment.cs
```diff
@@ -10,6 +10,7 @@ public class Environment
     {
         public static readonly Environment i = new Environment();
 
+        public WorldState worldState;
```
pbosio: can we make it readonly or convert it into a property so we avoid de possibility being set by mistake?

pbosio (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Environment.cs
BrianAmadori: To reset its state

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Environment.cs
BrianAmadori: I'll try and add an `Initialize` method so this is possible. In any case, the `readonly` pattern will dissappear in the short term because I found it to be not very viable to be able to mock systems on demand, its just too restrictive.

BrianAmadori: Done

unity-client/Assets/Tutorial/Scripts/TutorialController.cs
```diff
@@ -39,44 +39,49 @@ internal enum TutorialPath
 
         public static TutorialController i { get; private set; }
 
-        public HUDController hudController { get => HUDController.i; }
+        public HUDController hudController
+        {
+            get => HUDController.i;
+        }
```
sandrade-dcl, line 45: Hey Brian, a doubt: These kind of indexations are made automatically by your code editor? or are made manually by you? In my opinion, in cases like this (a simple property without complex code inside) we can avoid to use 4 lines for something that can be easily read in 1. But I understand that it is something very subjective :D

sandrade-dcl (approved):

unity-client/Assets/Tutorial/Scripts/TutorialController.cs
BrianAmadori: The indexing was already in place. It seems that the code formatter spread the code into multiple lines, I guess I have to review the rider C# style configuration.

PR #1780, opened 2020-12-08
refactor: move scene controller to environment

BrianAmadori:


BrianAmadori:
This PR became too dirty, will restart the process in other smaller PRs.

PR #1792, opened 2020-12-09
feat: invalidate scenes edited by builder in world

nchamo:
Before this change, scenes were never "forgotten" by the kernel. Scenes could be unloaded when out of sight and loaded again if they were close enough, but there was an association between parcel and scene id that was only removed when the browser was refreshed.

With the Builder In-World feature, we are actively modifying scenes and deploying changes to the content server. Therefore, when leaving edit mode, we might need to re-fetch scene, since they were probably modified. In this PR we are making two different changes:
1. We are adding the ability to reload a scene to Kernel
2. When the renderer triggers a scene publish, we are now sending back a message to it, to let it know whether it was successful or not


kernel/packages/decentraland-loader/lifecycle/controllers/scene.ts
```diff
@@ -26,49 +25,75 @@ export class SceneLifeCycleController extends EventEmitter {
     this.enabledEmpty = opts.enabledEmpty
   }
 
-  contains(status: SceneLifeCycleStatus, position: Vector2Component) {
-    return (
-      status.sceneDescription &&
-      status.sceneDescription.sceneJsonData.scene.parcels.includes(`${position.x},${position.y}`)
-    )
+  async reportSightedParcels(sightedParcels: string[], lostSightParcels: string[]) {
```
nchamo, line 28: Most changes to this file were refactors. The only thing that added was the `reloadScene` method. The refactors were:
1. We deleted methods `hasStarted` and `contains` that were not being used
1. Renamed `onSight` to `startSceneLoading`
1. Renamed `lostSight` to `unloadScenes`
1. Set a lot of methods to private (since they weren't being used outside this class), and moved them to the bottom of the file

pablitar (approved):
LGTM!

PR #1798, opened 2020-12-10
refactor: main entry point initial implementation (I)

BrianAmadori:
This PR lays the groundwork for moving `SceneController` to `Environment` and having an scene-independent entry point for tests.

- Add new `Main` MonoBehaviour that will take `SceneController` entry point responsibilities
- Added `Main` to `SceneController` prefab, replacing `SceneController`.
- `SceneController` is not a `MonoBehaviour` anymore.
- Added some lazy wiring to keep the `SendMessage` pipeline in working state


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Main.cs
```diff
@@ -0,0 +1,99 @@
+﻿using System;
+using DCL.Components;
+using DCL.Helpers;
+using UnityEngine;
+using UnityEngine.Serialization;
+
+namespace DCL
+{
+    public class Main : MonoBehaviour
```
sandrade-dcl, line 13: Can we add some basic XML documentation explaining the responsibilities of this Monobehaviour?

sandrade-dcl (approved):
Good work!

Perhaps a next step in the future could be to make the `SceneController` class implements an interface, remove the static reference `i` and use this new interface to be injected in the rest of classes that are currently using the static reference of `SceneController`. It would lead to a bit of refactor (I don't know if it would be very expensive or not) but I think it would improve a lot the implementation of the tests for all the entities that need to use the `SceneController` because it could be mocked easily.

eordano:
Some comments on naming:

* What is `Main`?
* Is `SceneController` a POCO or a Prefab?
* A `Controller` is generally the opposite of a `POCO`; only has code, no data... can we change the name of `SceneController`?

BrianAmadori:
@eordano @sandrade-dcl I'm going to remove `SceneController` and split it in a few interfaceable classes. Will not do this in this series of PRs because it would involve a massive change in the codebase and it wouldn't be very reviewable. Let's do this in baby steps.

@sandrade-dcl For the `SceneController.i` removal check the part 3 of this series, it does exactly that.

BrianAmadori:
@eordano about `SceneController` prefab/poco: we actually have a class and a prefab with that name. I will try to at least rename the prefab so we don't have this confusion anymore. The `SceneController` class its not supposed to hold any data btw (It will be wiped anyways as I stated in the last comment tho).

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Main.cs
BrianAmadori: done

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Main.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// Most of the application subsystems should be initialized from this class Awake() event.
+    /// </summary>
+    public class Main : MonoBehaviour
+    {
+        public static Main i { get; private set; }
+
+        [NonSerialized]
+        public SceneController sceneController;
+
+        [FormerlySerializedAs("factoryManifest")]
```
AjimenezDCL: We dont need this once the serialized asset is saved with the new property.

unity-client/Assets/Scripts/Tests/TestHelpers.cs
```diff
@@ -1098,47 +1099,47 @@ public static IEnumerator UnloadAllUnityScenes()
 
         public static SceneController InitializeSceneController(bool usesWebServer = false)
         {
-            var sceneController = UnityEngine.Object.FindObjectOfType<SceneController>();
+            var main = UnityEngine.Object.FindObjectOfType<Main>();
```
AjimenezDCL, line 1102: Not a big fan of having this kind of things in tests. But I understand we are just "porting" the old `SceneController` behaviour into this new `Main` entry point. 

Let's not forget about this in the future and find a more elegant solution!

AjimenezDCL (approved):
Approved with a small concern that can be addressed in the future. Great job!

unity-client/Assets/Scripts/Tests/TestHelpers.cs
BrianAmadori: Of course. This code is ancient.

pravusjif (approved):

PR #1803, opened 2020-12-10
fix: populate ethAddress with real eth address

pravusjif:
**WHY**
* The builder in-world needs the current user **real** ethereum address to validate ownership on the Picture Frames (NFTShapes) used while building. But we only receive the lowercase ethereum address (not valid as a checksum address, OpenSea doesn't recognize it) in Unity, so we need to receive the real one for the new need.

* The user profile `ethAddress` property was not being really used (as it contained the same value as the userId), and with this new need we can take advantage of that existent property and fix it to contain the real ethereum address.

**WHAT**
* This PR fixes the usage of the `ethAddress` property in the user profile, since it only contained the same value as the userId (the lowercase address) instead of the real ethereum address. Now we store the real ethereum address we get from the user when he enters the application.
* By fixing that, we get to transport the real eth address into the unity client, enabling the builder in-world to validate wether or not the current user is the owner of the NFTshape/PictureFrame that the user may want to place in their scene. (this was the original necessity that triggered this PR)

**Reusing 'ethAddress' existent property in the user profile**
* `ethAddress` was holding the same value as userId (the lowercased address) and sometimes it was holding the 'noeth' string. I did an extra check with logs everywhere we used `ethaddress` in master branch to confirm this.
* No real usage of the property was found since the old usages can be replaced by using userId. This can be seen [here](https://github.com/decentraland/explorer/blob/e159de309c55f8b9bdc7dd19bcaaf3ef953e5c30/kernel/packages/shared/session/sagas.ts#L171) and [here](https://github.com/decentraland/explorer/blob/e159de309c55f8b9bdc7dd19bcaaf3ef953e5c30/kernel/packages/shared/session/sagas.ts#L289). So the few old usages for that propety have been replaced to using `userId` as can be seen in the files changed in this PR.

**Another possible approach**
Another approach would be to add the necessary node modules for computing the checksum based on the lowercased address and process it (convert it back to the value we already had at the beginning) on demand (probably at profile preparation before sending it to the renderer). But i think it's highly unneeded to add more modules and converting-back to the original address we receive when the user connects, instead of saving it at that point in time.


kernel/packages/shared/ethereum/provider.ts
```diff
@@ -83,7 +83,7 @@ export async function getUserAccount(): Promise<string | undefined> {
       return undefined
     }
 
-    return accounts[0].toJSON().toLocaleLowerCase()
```
pablitar, line 86: Have you ensured that no other uses of this method are compromised somewhere? I'm worried that this change could have unforeseen consequences.

If we are not sure, I'd add a boolean parameter to the method called `withChecksum` that defaults to false, and then only set the parameter to true in the places we want to use the checksummed address.

kernel/packages/shared/session/sagas.ts
```diff
@@ -286,7 +295,7 @@ function* signUp() {
 
   const profile = yield select(getSignUpProfile)
   profile.userId = session.userId.toString()
-  profile.ethAddress = session.userId.toString()
+  profile.ethAddress = session.rawAddress ? session.rawAddress.toString() : profile.userId
```
pablitar: Why do you need to call `toString`?

kernel/packages/shared/session/sagas.ts
```diff
@@ -168,7 +169,7 @@ function* startSignUp(userId: string, identity: ExplorerIdentity) {
   let prevGuest = fetchProfileLocally(userId)
   let profile: Profile = prevGuest ? prevGuest : yield generateRandomUserProfile(userId)
   profile.userId = identity.address.toString()
-  profile.ethAddress = identity.address.toString()
+  profile.ethAddress = identity.rawAddress ? identity.rawAddress.toString() : profile.userId
```
pablitar: Why do you need to call `toString`?

pablitar: Also, according to the types you defined below this could never be undefined, so this check seems unnecessary

kernel/packages/shared/session/sagas.ts
pablitar: Also, according to the types you defined below this could never be undefined, so this check seems unnecessary

kernel/packages/decentraland-ecs/src/decentraland/Types.ts
```diff
@@ -420,6 +420,7 @@ export type ProfileForRenderer = {
   name: string
   description: string
   email: string
+  ethAddress: string
```
pablitar: This is SDK. I don't think we need to touch this for this use case.

pablitar:
> we already have an 'ethAddress' property in the user profile at packages/shared/profiles/types.ts:11 and apparently is not being used anywhere

Warning: It is being used. See for instance `kernel/packages/shared/profiles/selectors.ts:60` or `kernel/packages/unity-interface/delightedSurvey.ts:35`

I think we shouldn't change that property to be the one with checksums.

kernel/packages/shared/session/sagas.ts
pravusjif: I think the same but it was this way before I edited there, I can remove the toString, sure

pravusjif: Done

kernel/packages/shared/session/sagas.ts
pravusjif: Done

kernel/packages/decentraland-ecs/src/decentraland/Types.ts
pravusjif: Just tested without that and it works, removed.

pravusjif:
> > we already have an 'ethAddress' property in the user profile at packages/shared/profiles/types.ts:11 and apparently is not being used anywhere
> 
> Warning: It is being used. See for instance `kernel/packages/shared/profiles/selectors.ts:60` or `kernel/packages/unity-interface/delightedSurvey.ts:35`
> 
> I think we shouldn't change that property to be the one with checksums.

Since before i changed it, profile's `ethAddress` had the same value as the `userId` (lowercased address), I updated those 2 files you mentioned to use the userId just in case. I also checked for more references again.

kernel/packages/shared/ethereum/provider.ts
pravusjif: I had checked but just in case I just added the boolean, thx

kernel/packages/shared/ethereum/provider.ts
```diff
@@ -83,14 +83,14 @@ export async function getUserAccount(): Promise<string | undefined> {
...
+    return accounts[0].toJSON()
```
pablitar:
```suggestion
    return returnChecksum ? accounts[0].toJSON() : accounts[0].toJSON().toLowerCase()
```

kernel/packages/shared/profiles/transformations/processServerProfile.ts
```diff
@@ -57,7 +57,7 @@ export function processServerProfile(userId: string, receivedProfile: any): Prof
     hasClaimedName:
       typeof receivedProfile.hasClaimedName === 'undefined' ? !!receivedProfile.name : receivedProfile.hasClaimedName,
     description: receivedProfile.description || '',
-    ethAddress: userId || 'noeth',
+    ethAddress: receivedProfile.ethAddress || 'noeth',
```
pablitar, line 60: Keep in mind that if there is a profile that is stored on the server, it will probably have the address as lower case. For instance: https://peer.decentraland.org/content/entities/profile?pointer=0xcd6989ccbe98eaa0fa209cbaf688279c107d1061

I'm not sure how to solve this case, other than ensure that the address is checksummed here using an external library (which more or less defeats the purpose of avoiding the calculation)

kernel/packages/shared/ethereum/provider.ts
pravusjif: thx, i was rushing =(

kernel/packages/shared/profiles/transformations/processServerProfile.ts
pravusjif: I could apply the lowercase at that moment in time or just use userId in there

pravusjif: Also, probably in the servers everything is mapped to `userId` right? since ethAddress can contain "noeth" and couldn't be used as a key

pablitar (approved):

kernel/packages/shared/session/sagas.ts
```diff
@@ -286,7 +295,7 @@ function* signUp() {
 
   const profile = yield select(getSignUpProfile)
   profile.userId = session.userId.toString()
-  profile.ethAddress = session.userId.toString()
+  profile.ethAddress = session.identity.rawAddress ? session.identity.rawAddress : profile.userId
```
menduz: Why are we mixing `userId` and etherum addresses?

kernel/packages/unity-interface/delightedSurvey.ts
```diff
@@ -32,10 +32,10 @@ function delightedSurvey() {
       const profile = getCurrentUserProfile(globalThis.globalStore.getState())
       if (profile) {
         const payload = {
-          email: profile.ethAddress + '@dcl.gg',
+          email: profile.userId + '@dcl.gg',
           name: profile.name || 'Guest',
           properties: {
-            ethAddress: profile.ethAddress,
+            ethAddress: profile.userId,
```
menduz: why is ethAddres also userId?

kernel/packages/shared/session/types.ts
```diff
@@ -10,6 +10,7 @@ export type RootSessionState = {
 
 export type ExplorerIdentity = AuthIdentity & {
   address: string
+  rawAddress: string
```
menduz: what is the difference between address and rawAddress?

kernel/packages/shared/session/sagas.ts
```diff
@@ -168,7 +169,7 @@ function* startSignUp(userId: string, identity: ExplorerIdentity) {
   let prevGuest = fetchProfileLocally(userId)
   let profile: Profile = prevGuest ? prevGuest : yield generateRandomUserProfile(userId)
   profile.userId = identity.address.toString()
-  profile.ethAddress = identity.address.toString()
+  profile.ethAddress = identity.rawAddress ? identity.rawAddress : profile.userId
```
menduz: we either have or not have an eth address, please do not mix pears and bananas under names that may have impact on smart contracts.

What is the difference between userId and rawAddress?

menduz (changes requested):

kernel/packages/shared/session/sagas.ts
pravusjif: The userId always contained the **_lowercased ethereum address_** (stored in `identity.address`), now we store the real ethereum address in `ethAddress`.

Thanks for your suggestion, I removed the whole `rawAddress` existence check as it doesn't make sense (since we first get the raw address and then lowercase it for the userId), and tested everything again.

kernel/packages/unity-interface/delightedSurvey.ts
pravusjif: I updated the property we send to be called now `userId`, and validated with @2fd that that property name can be changed without risk =), fixing confusions with the previous implementation as you pointed out.

kernel/packages/shared/session/types.ts
pravusjif: the identity address contains the **_lowercased ethereum address_** of the user, while the new `rawAddress` has the real ethereum address of the user.

I added comments for clarification, as you suggested, thanks and check it out!

kernel/packages/shared/session/sagas.ts
pravusjif: The userId always contained the **_lowercased ethereum address_** (stored in `identity.address`), now we store the real ethereum address in `ethAddress`.

Thanks for your suggestion, I removed the whole `rawAddress` existence check as it doesn't make sense (since we first get the raw address and then lowercase it for the userId), and tested everything again.

menduz (approved):

PR #1805, opened 2020-12-11
feat: Move Settings button to the taskbar

sandrade-dcl:
Move the Settings button from the taskbar more menu to the own taskbar.

image


pravusjif (approved):
I'm approving it but take into account that probably taskbar items should all behave the same way (be toggleable, like the Explore feature) and right now this settings panel enables an overlay on top of everything and can't be toggled back from the taskbar.

AjimenezDCL (approved):

PR #1807, opened 2020-12-11
Settings key added on controls panel

RominaMarchetti:
P key reference added on controls panel.


pbosio (approved):
Is the P hotkey actually connected? I'm pressing P and nothing is happening 🤷

sandrade-dcl (approved):

sandrade-dcl:
> Is the P hotkey actually connected? I'm pressing P and nothing is happening 🤷

If you add the `&ENABLE_NEW_SETTINGS` in the url, you should be able to use the [P] key 👍

PR #1809, opened 2020-12-11
feat: Set the new Settings Panel as default

sandrade-dcl:
- If we don't use the `ENABLE_OLD_SETTINGS` feature flag: The new `SettingsPanelHUD` will be showed as default.
- If we use the `ENABLE_OLD_SETTINGS` feature flag: The old `SettingsHUD` will be showed as default.


pravusjif (approved):

AjimenezDCL (approved):

BrianAmadori (approved):

pbosio (changes requested):
I'm getting the following error when using `ENABLE_OLD_SETTINGS` flag
Is it possible that in that case `settingsPanelHud` is null and there are some null check missing there?
Screen Shot 2020-12-11 at 15 39 57

pbosio (approved):

PR #1815, opened 2020-12-14
fix: add missing nullcheck in loadable shape

pravusjif:
Recently I found a null argument exception happening at wonder mine scene like a minute after spawning there, in browser console ( https://play.decentraland.zone/?ENV=org&position=-29%2C55 )

* Added missing null check
* Removed debug.log

bugged (master): https://play.decentraland.zone/?ENV=org&position=-29%2C55
fixed: https://explorer.decentraland.zone/branch/fix/AddMissingNullCheckInLoadbleShape/index.html?ENV=org&position=-29%2C55


pbosio (approved):
This made me realize that there is an issue with NFTs and `GetLoaderForEntity` 😅

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadableShape.cs
```diff
@@ -25,6 +25,8 @@ public new class Model : BaseShape.Model
 
         public static LoadWrapper GetLoaderForEntity(DecentralandEntity entity)
         {
+            if(entity.meshRootGameObject == null) return null;
```
BrianAmadori: Can we add a `Log.Warning` here? I'm concerned that we might be hiding a bigger error by adding this check here, so it'd be better if we can still see when it happens at least in debug mode or editor.

BrianAmadori (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadableShape.cs
pravusjif: Sure, just added it.

PR #1824, opened 2020-12-14
refactor: scene lifecycle handler

BrianAmadori:
Moved all scene lifecycle logic to a specialized class. This decoupling will be useful to put `ParcelScene` into its own assembly later.


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneLifecycleHandler.cs
```diff
@@ -0,0 +1,147 @@
+﻿using System.Collections.Generic;
+using DCL.Components;
+using DCL.Models;
+using UnityEngine;
+
+namespace DCL.Controllers
+{
+    public class SceneLifecycleHandler
```
sandrade-dcl, line 12: As we usually are doing with all the new classes, it would be fine to add a basic XML documentation about the responsibilities of this class :)

sandrade-dcl (approved):
Good refactoring work 👍

AjimenezDCL (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -1174,94 +1114,51 @@ private DecentralandEntity GetEntityForUpdate(string entityId)
...
-                Debug.Log($"{sceneData.basePosition} Disposable objects left... {disposableNotReady.Count}");
-            }
-
-            if (disposableNotReady.Count == 0)
+            List<string> allDisposableComponents = disposableComponents.Select(x => x.Key).ToList();
+            foreach (string id in allDisposableComponents)
             {
-                SetSceneReady();
+                Environment.i.parcelScenesCleaner.MarkDisposableComponentForCleanup(this, id);
             }
```
pbosio, line 1123: Why not use `disposableComponents.Keys` here instead of doing a `Select` + `ToList`?

pbosio (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
```diff
@@ -1174,94 +1114,51 @@ private DecentralandEntity GetEntityForUpdate(string entityId)
...
-
-            state = State.READY;
-
-            Environment.i.sceneController.SendSceneReady(sceneData.id);
-            RefreshName();
-
-            OnSceneReady?.Invoke(this);
+#if UNITY_EDITOR
+            Debug.Log(GetStateString());
+#endif
```
pravusjif, line 1150: We could still apply the state string to the GO name right? in that case we should add that here, maybe even replace this `Debug.Log()` with that

pravusjif: We should refresh the GO name here, right? instead of the `Debug.Log()`

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
pravusjif: Since we are marking the disposables for cleanup (not destroying them immediately) I also think we should just traverse the `disposableComponents` existent dictionary keys instead of creating a strings list here.

pravusjif (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/ParcelScene.cs
BrianAmadori: The request is valid but we should address it in another PR. This PR is just to move the existing code to another class.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneLifecycleHandler.cs
BrianAmadori: Done

pravusjif (approved):

PR #1825, opened 2020-12-14
fix: send publish scene result properly

nchamo:
In https://github.com/decentraland/explorer/pull/1792, we started returning the result of the publish action to the renderer. However, this was done by sending a new `EntityAction` through the scene's messages.

This proved to be a mistake at this point. Scene messages need to be sent via native messages or protocol buffer, they can't be plain JSON. Also, this messages are received, decoded and interpreted by the SceneController, which is not what we wanted for this particular action. 

We tried to come up with a generic way to re-route messages that are sent this way, but this already exists by using the `UnityInterface`. For all these reasons, we are now sending the message directly from the controller.

### DIFF
--- a/kernel/packages/unity-interface/UnityInterface.ts
+++ b/kernel/packages/unity-interface/UnityInterface.ts
@@ -24,6 +24,7 @@ import { HotSceneInfo } from 'shared/social/hotScenes'
 import { defaultLogger } from 'shared/logger'
 import { setDelightedSurveyEnabled } from './delightedSurvey'
 import { renderStateObservable } from '../shared/world/worldState'
+import { DeploymentResult } from '../shared/apis/SceneStateStorageController/types'
 
 const MINIMAP_CHUNK_SIZE = 100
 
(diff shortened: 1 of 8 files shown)


kernel/packages/unity-interface/UnityInterface.ts
```diff
@@ -415,6 +416,10 @@ export class UnityInterface {
     this.gameInstance.SendMessage('Bridges', 'UpdateRealmsInfo', JSON.stringify(realmsInfo))
   }
 
+  public SendPublishSceneResult(result: DeploymentResult) {
+    this.gameInstance.SendMessage('Main', 'PublishSceneResult', JSON.stringify(result))
```
BrianAmadori, line 420: This method is already present on Unity side? Or we are going to add it via another PR?

BrianAmadori (approved):
I'm lacking the entire context of the stateful scenes deploy flow, but at a first glance it looks good to me!

kernel/packages/unity-interface/UnityInterface.ts
nchamo: We will need to add it on another PR

PR #1826, opened 2020-12-15
style: settings panel

RominaMarchetti:


BrianAmadori (approved):
Visually looks good. However, when activating/deactivating auto settings, the color grading remains turned off. Maybe we can address this on another PR?

pravusjif (approved):

PR #1827, opened 2020-12-15
feat: nearby users list add voicechat settings and other improvements

pbosio:


pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/UsersAroundListHUD/UsersAroundListHUDListElementView.cs
```diff
@@ -58,6 +59,21 @@ public void SetUserProfile(UserProfile profile)
...
+                SetupFriendship(status.friendshipStatus);
+            }
+            else
+            {
+                SetupFriendship(FriendshipStatus.NONE);
+            }
+
+            FriendsController.i.OnUpdateFriendship -= OnFriendActionUpdate;
+            FriendsController.i.OnUpdateFriendship += OnFriendActionUpdate;
+        }
```
AjimenezDCL: Let's invert this to avoid indendation! I would suggest to make the comparison to null explicit as well, some types might have implicit operators to bool and it's a good practice to avoid them.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/UsersAroundListHUD/UsersAroundListHUDListElementView.cs
```diff
@@ -89,6 +105,12 @@ public void OnPoolRelease()
             profile.OnFaceSnapshotReadyEvent -= SetAvatarPreviewImage;
             profile = null;
         }
+
+        if (FriendsController.i)
```
AjimenezDCL: As asked before, let's make the comparison to null explicit

AjimenezDCL (approved):
Approving with a couple of suggestions

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/UsersAroundListHUD/UsersAroundListHUDListElementView.cs
pbosio: I don't like inverting if inside methods that do other stuff also so I created a new one for friend's setup and inverted it there :P

PR #1831, opened 2020-12-16
feat: Add new tutorial steps related to social features

sandrade-dcl:
- [x] Add a new tutorial step (tooltip) for the **Users Around** list.

image

- [x] Add a new tutorial step (tooltip) for the **social features** groups in the taskbar (voice chat, world chat and friends).

image

- [x] Create a **new tutorial path** for users that already saw the tutorial (this path should be executed only one time per user). In this path the user should view the tooltip for the **Users Around** list but it will be automatically hidden after 5 seconds.
**NOTE**: _The implementation of this new path is something TEMPORAL until we refactorize the tutorial system to make it compatible with incremental features._

image

**TESTING URL**: https://play.decentraland.zone/branch/feat/New-tutorial-step-for-the-Users-Around-list/index.html?ENV=org

**NOTE**: Add the param `&RESET_TUTORIAL` in the url if you need to restart the tutorial flow.


unity-client/Assets/Tutorial/Scripts/TutorialController.cs
```diff
@@ -171,6 +178,11 @@ public void SetTutorialEnabled(string fromDeepLink)
             SetupTutorial(fromDeepLink, TutorialType.Initital);
         }
 
+        public void SetTutorialEnabledForUsersThatAlreadyDidTheTutorial()
+        {
+            SetupTutorial(false.ToString(), TutorialType.Initital, true);
```
BrianAmadori:
```suggestion
            SetupTutorial(false.ToString(), TutorialType.Initial, true);
```

BrianAmadori (commented):
Looks good! It's ok to show the new step every time the user goes into explorer? It can be annoying.

kernel/packages/unity-interface/UnityInterface.ts
```diff
@@ -306,6 +306,10 @@ export class UnityInterface {
     this.gameInstance.SendMessage('TutorialController', 'SetTutorialEnabled', JSON.stringify(fromDeepLink))
   }
 
+  public SetTutorialEnabledForUsersThatAlreadyDidTheTutorial() {
+    this.gameInstance.SendMessage('TutorialController', 'SetTutorialEnabledForUsersThatAlreadyDidTheTutorial')
```
AjimenezDCL, line 310: we really need a better way for this in the future. Shall we create a ticket so we dont forget?

AjimenezDCL (approved):
Approving with some concerns on the approach, but since the release is pushing, let's solve it later.

kernel/packages/unity-interface/UnityInterface.ts
pbosio: I was about to write the same comment 😂
I don't like this approach very much, I have an idea to doing it more scalable but we are now in a hurry, so we can revisit it later

sandrade-dcl:
> Looks good! It's ok to show the new step every time the user goes into explorer? It can be annoying.

Yes, it will be always showed but it automatically hides after 5 seconds.

unity-client/Assets/Tutorial/Scripts/TutorialController.cs
sandrade-dcl: Done

pbosio:
> > Looks good! It's ok to show the new step every time the user goes into explorer? It can be annoying.
> 
> Yes, it will be always showed but it automatically hides after 5 seconds.

We could add something like `UsersAroundListStep = 512` to the `TutorialFinishStep` and add it to the tutorialflag of the profile to avoid showing it again

kernel/packages/unity-interface/UnityInterface.ts
sandrade-dcl: Yes, I agree. In this case we have needed to apply these changes on the tutorial flow a day before the release and it has been the quicker way. I will annote this for a future improvement.

pravusjif (approved):

pbosio (approved):

PR #1833, opened 2020-12-16
refactor: new environment (I)

BrianAmadori:
This PR includes the new `Environment` design, now `Environment` is a container for three dependency contexts and will serve as the base of our new initialization layer:

#### MessagingContext
Contains all the scene messaging related dependencies. For now only `MessagingControllersManager` is here, but it'll contain more classes that will come when we split `SceneController` in more classes.

#### WorldRuntimeContext
Contains all the needed dependencies for the decentraland world runtime implementation (i.e. the SDK implementation).

#### PlatformContext
Contains all the needed dependencies for the low level systems. Memory management, IO, rendering, input and debugging related systems belong here.

---

#### What's next and not included in this PR

The contents and layout of these classes is obviously _not final_. We have a few systems that are not present here, like `DCLCharacterController`, `InputController`, etc. We should be adding those systems into the contexts in a near future.

As the next steps, the plan is to extract interfaces of all systems, include a new mocked path for the context factories and integrate at least a single test. The new test integration should drop the `TestsBase` inheritance and run solely based on the new environment setup. This will be addressed in the next PR, not included here.

---

#### Fixes included:

General:
- Adapted all our code to work with the new environment
- Rolled back the integration of `InteractionHoverCanvasController` with `Environment` as it should not inherit from monobehaviour to work well with this scheme.
- `PerformanceMetricsController` is not part of `Environment` anymore because it's a standalone system. Moved it to `Main`.

Tests:
- Adapted our current test initialization flow to work with the new environment
- Removed `TestsBase` inheritance from some `AssetPromiseKeeper` tests.
- fix: `CursorController` initialization race condition with `PointerEventsController` that made tests very flaky
- fix: `MemoryManager` was leaking `OnChange` subscriptions into `RenderingController`. This made tests get slower with each subsequent scene reload.
- Butchered the old `SetUp_SceneController` implementation in `TestsBase` as `SceneController` is not a monobehaviour anymore and the logic made no sense. 
- fix: In some tests, the entities pool prewarm was still occurring. 
- refactor: Remove `TestHelpers.ForceUnloadAllScenes`. This functionality was needed in `SceneController.Dispose()` so the force behaviour was integrated into the `UnloadAllScenes` with a flag. I'd rather drop the flag and make a new method, but not sure how I should name it. Suggestions are welcome.


unity-client/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/InteractionHoverCanvas/InteractionHoverCanvasController.cs
```diff
@@ -24,8 +25,7 @@ public class InteractionHoverCanvasController : MonoBehaviour
 
     void Awake()
     {
-        Environment.i.SetInteractionHoverCanvasController(this);
```
BrianAmadori, line 27: Had to remove the `InteractionHoverCanvasController` from environment for now. `Environment` is not supposed to deal with `Monobehaviour` and this flow in its current state will give race condition issues.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BlockerController/WorldBlockersController.cs
```diff
@@ -36,35 +36,43 @@ public class WorldBlockersController
             new Vector2Int(-1, 1)
         };
 
-        public WorldBlockersController(ISceneHandler sceneHandler, IBlockerInstanceHandler blockerInstanceHandler, DCLCharacterPosition characterPosition)
+        public void Initialize(ISceneHandler sceneHandler, IBlockerInstanceHandler blockerInstanceHandler, DCLCharacterPosition characterPosition)
```
BrianAmadori: For now, `WorldBlockersController` is initialized by this method instead of the constructor. This is done because we can't replace the class instance in the `Environment` initialization stage because its read only. 

Not a fan of having instantiation and initialization split like this, but fixing that issue would be a pretty big change. This should work for now.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/CursorController.cs
```diff
@@ -4,31 +4,25 @@
 
 public class CursorController : MonoBehaviour
 {
+    public static CursorController i { get; private set; }
     public Image cursorImage;
     public Sprite normalCursor;
     public Sprite hoverCursor;
 
-    void OnEnable()
+    void Awake()
```
BrianAmadori, line 12: Found that a big reason for our `InteractionHoverCanvas` flaky tests were due to `PointerEventsController` being re-initialized after `OnEnabled` is called, losing the event suscriptions. As this class should be at a higher layer of abstraction than `PointerEventsController`, the dependencies were inverted and the problem fixed.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/PointerEventsController.cs
```diff
@@ -35,7 +35,10 @@ public void Initialize()
             InputController_Legacy.i.AddListener(WebInterface.ACTION_BUTTON.PRIMARY, OnButtonEvent);
             InputController_Legacy.i.AddListener(WebInterface.ACTION_BUTTON.SECONDARY, OnButtonEvent);
 
-            hoverController = Environment.i.interactionHoverCanvasController;
+            hoverController = InteractionHoverCanvasController.i;
+
+            OnPointerHoverStarts += CursorController.i.SetHoverCursor;
+            OnPointerHoverEnds += CursorController.i.SetNormalCursor;
```
BrianAmadori: The rest of the `InteractionHoverCanvasController` dependency inversion referred on the other comment.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Main.cs
```diff
@@ -18,6 +18,8 @@ public class Main : MonoBehaviour
 
         public DebugConfig debugConfig;
 
+        private PerformanceMetricsController performanceMetricsController;
```
BrianAmadori, line 21: `PerformanceMetricsController` is an implementation detail and nobody will reference it, so moved this away from `Environment` onto `Main`.

unity-client/Assets/Scripts/MainScripts/DCL/Environment/Environment.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        public static void Setup(System.Func<MessagingContext> messagingBuilder,
+            System.Func<PlatformContext> platformBuilder,
+            System.Func<WorldRuntimeContext> worldRuntimeBuilder)
+        {
+            Environment.messagingBuilder = messagingBuilder;
+            Environment.platformBuilder = platformBuilder;
+            Environment.worldRuntimeBuilder = worldRuntimeBuilder;
+            Setup();
+        }
```
BrianAmadori, line 43: We pass `Func` as builder methods in order to be able to reset the environment state with the same implementations.

unity-client/Assets/Scripts/MainScripts/DCL/Environment/Environment.cs
```diff
@@ -0,0 +1,97 @@
...
+            model.world.pointerEventsController.Initialize();
+            model.platform.memoryManager.Initialize();
+            model.world.worldState.Initialize();
+            model.world.worldBlockersController.InitializeWithDefaultDependencies(
+                model.world.worldState,
+                DCLCharacterController.i.characterPosition);
+            model.platform.parcelScenesCleaner.Start();
+            model.platform.cullingController.Start();
+            model.world.sceneBoundsChecker.Start();
+        }
```
BrianAmadori: In this method, each subsystem dependencies are wired together. I don't like this approach at all because it can turn to a big ball of mud pretty easily. 

For now, this is necessary because a system can reference another that doesn't exist yet. 

An approach to keep using constructors and get rid of these Start/Initialize/etc methods is to have promises that ensure that a subsystem will be available at a later time, and pass them through each subsystem constructor. Of course this would be a major change, so I think its best to leave this method like this right now and improve it in the future.

unity-client/Assets/Scripts/MainScripts/DCL/Environment/MessagingContext/MessagingContext.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace DCL
+{
+    public class MessagingContext : System.IDisposable
+    {
+        public readonly MessagingControllersManager messagingControllersManager;
```
BrianAmadori: When we split `SceneController` in different classes (`MessageBridge`, `MessageQueuer`, `MessageProcessor`, etc). This context will grow,

unity-client/Assets/Scripts/MainScripts/DCL/Environment/PlatformContext/PlatformContext.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace DCL
+{
+    public class PlatformContext : System.IDisposable
+    {
+        public readonly MemoryManager memoryManager;
+        public readonly ICullingController cullingController;
+        public readonly Clipboard clipboard;
+        public readonly PhysicsSyncController physicsSyncController;
+        public readonly IParcelScenesCleaner parcelScenesCleaner;
+        public readonly DebugController debugController;
```
BrianAmadori: We use the `readonly` pattern instead of private setters to avoid the access performance penalty.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/MemoryManager/MemoryManager.cs
```diff
@@ -6,28 +7,42 @@
...
-            CoroutineStarter.Start(AutoCleanup());
+            autoCleanupCoroutine = CoroutineStarter.Start(AutoCleanup());
+        }
+
+        public void Dispose()
+        {
+            CleanupPoolsIfNeeded(true, immediate: true);
+
+            if (autoCleanupCoroutine != null)
+                CoroutineStarter.Stop(autoCleanupCoroutine);
```
AjimenezDCL, line 29: We should nullify `autoCleanupCoroutine`. It's harmless the way it is since we are only stopping the coroutine when disposing, but I rather have it nulled just in case.

unity-client/Assets/Scripts/MainScripts/DCL/Environment/Environment.cs
AjimenezDCL: I love the idea of the builders. Really well thought!

unity-client/Assets/Scripts/MainScripts/DCL/Environment/Environment.cs
```diff
@@ -0,0 +1,126 @@
+﻿using System.ComponentModel;
+using DCL.Controllers;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+
+namespace DCL
+{
+    public class Environment
+    {
+        public static Model i = new Model();
```
AjimenezDCL, line 10: I find a bit confusing that `Environment.i` wont return an `Environment` instance. I have no alternative though.

unity-client/Assets/Scripts/MainScripts/DCL/Environment/Environment.cs
```diff
@@ -0,0 +1,126 @@
...
+            Dispose();
+            Setup();
+        }
+
+        /// <summary>
+        /// Dispose() all the current environment systems. 
+        /// </summary>
+        public static void Dispose()
+        {
+            i.Dispose();
```
AjimenezDCL: Do we need a null check here?

unity-client/Assets/Scripts/MainScripts/DCL/Environment/PlatformContext/PlatformContext.cs
```diff
@@ -0,0 +1,41 @@
...
+            this.debugController = debugController;
+        }
+
+        public void Dispose()
+        {
+            memoryManager.Dispose();
+            parcelScenesCleaner.Dispose();
+            cullingController.Dispose();
+            debugController.Dispose();
+        }
```
AjimenezDCL, line 39: Could two different contexts share a manager/controller? Could that lead to issues where a dispossing context cleans up a manager/controller still in use by other context?

AjimenezDCL (approved):
Looks great, I just left minor comments and concerns!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -17,67 +17,49 @@ public class SceneController : IMessageProcessHandler, IMessageQueueHandler
...
-#if !UNITY_EDITOR
-            worldEntryPoint = new EntryPoint_World(this); // independent subsystem => put at entrypoint but not at environment
-#endif
-
-            // TODO(Brian): This should be fixed when we do the proper initialization layer
-            if (!EnvironmentSettings.RUNNING_TESTS)
-            {
-                Environment.i.cullingController.Start();
-            }
+            //TODO(Brian): Move those subscriptions elsewhere.
```
sandrade-dcl, line 45: Only a doubt about this `TODO` comment: Why should these subscriptions be moved from the `Initialize()`?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -725,9 +700,17 @@ public void UnloadParcelSceneExecute(string sceneKey)
             ProfilingEvents.OnMessageProcessEnds?.Invoke(MessagingTypes.SCENE_DESTROY);
         }
 
-        public void UnloadAllScenes()
+        public void UnloadAllScenes(bool includePersistent = false)
         {
-            var list = Environment.i.worldState.loadedScenes.ToArray();
+            if (includePersistent)
+            {
+                foreach (var keyValuePair in Environment.i.world.state.loadedScenes.Where(x => x.Value.isPersistent))
```
sandrade-dcl: In this case, wouldn't it better to store the result of `Environment.i.world.state.loadedScenes.Where(x => x.Value.isPersistent)` in a variable before the loop?

unity-client/Assets/Scripts/MainScripts/DCL/Environment/Environment.cs
sandrade-dcl: In order to avoid the confussion, maybe in this case we should not call this variable as `i` (we always associate `i` as "instance") and rename it by something like `model` or something like that. What do you think?

sandrade-dcl (changes requested):
Good job! I love the `Environment` modularization! I left some little comments.

unity-client/Assets/Scripts/MainScripts/DCL/Environment/PlatformContext/PlatformContext.cs
BrianAmadori: No, they shouldn't. That could lead to issues as you said.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
BrianAmadori: Indeed. I just moved the existing code, I agree it's too lazy. The request seems to be harmless so I'm going to try it!

unity-client/Assets/Scripts/MainScripts/DCL/Environment/Environment.cs
BrianAmadori: We can rename it to `model`, but it will add considerable syntax overhead and that's rather undesirable. Let me think about this. In any case we can agree to make this change in another PR? Because it will add a lot of changes and break the subsequent PRs.

AjimenezDCL: Sure thing, I approved it already taking that into account

sandrade-dcl: Of course, if it will provoke a lo of syntax overhead, we can do it on another PR in the near future 👍 Even if we decide that it is something that it's not worth it, for me it is only a detail.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
BrianAmadori: Those subscriptions are related to the `physicsSyncController` and `cullingController`. Having them there is a heavy violation on encapsulation, the subscriptions should be performed in the same systems that use it.

unity-client/Assets/Scripts/MainScripts/DCL/Environment/Environment.cs
BrianAmadori: Done

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
BrianAmadori: Done

pravusjif (approved):

AjimenezDCL:
https://github.com/decentraland/explorer/pull/1833#discussion_r545798934

We agreed to leave that for another PR. Gonna merge this

PR #1842, opened 2020-12-17
fix: World Chat Window is being hidden when click on the world

sandrade-dcl:
World Chat Window is being hidden when click on the world.

It started to happen when we changed the sorting order of the chat in order to not to overlap the new tutorial tooltip `TutorialStep_Tooltip_SocialFeatures`. So we have revert the changes in the chat prefab and have fixed the problem from the own step controller.


pravusjif (approved):
the bug is fixed!

pbosio (approved):

PR #1847, opened 2020-12-17
refactor: decouple DCLCharacterPosition usages

BrianAmadori:
This change is needed to fully implement the new environment, as some dependencies are coupled by this class.

So make an infinite world, every few meters the whole world has to be adjusted, so the positions don't get too far of zero. If this happens, precision errors occur. `DCLCharacterPosition` is responsible for handling this behaviour. 

Some systems need `DCLCharacterPosition` in order to:
- Subscribe to the world reposition.
- Get the player adjusted position after a world reposition.

`DCLCharacterPosition` is not on `Environment` and is unwise to move it there. Right now, the offset of the world reposition is already handled by a BaseVariable (`WorldOffset`). So we can just subscribe to that SO and handle all the use cases using that approach.


AjimenezDCL (approved):
Really nice!

pbosio (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraController.cs
```diff
@@ -132,7 +138,7 @@ public Vector3 GetPosition()
 
     private void OnDestroy()
     {
-        CommonScriptableObjects.playerUnityToWorldOffset.OnChange -= PrecisionChanged;
+        CommonScriptableObjects.worldOffset.OnChange -= OnWorldReposition;
```
sandrade-dcl:
```suggestion
        worldOffset.OnChange -= OnWorldReposition;
```

sandrade-dcl (approved):

PR #1851, opened 2020-12-17
fix: error modal in preview mode

caminoezequiel:


menduz:
If we merge this, how do we visualize the errors in preview mode?

caminoezequiel:
> If we merge this, how do we visualize the errors in preview mode?

@menduz We never had an error screens in the preview mode

menduz:
I believe this PR exists because that `showErrorModal` function was called from the preview mode and it broke the user's scenes, otherwise we should not have this PR and we should close it, because that function should never be called

menduz:
What I meant is that it happened, so we need to find a way to show the user errors on the preview, not silently fail.

menduz (changes requested):
We should find a way to show the user errors in preview mode, not silently fail.

PR #1853, opened 2020-12-18
feat: biw entity information UI

D4rWiNSS:
# What? 
This PR gives the user the ability to inspect the entity attributes and change them 

# Why?
Right now there is no way to inspect the entity's attributes. Furthermore, this will connect with the implementation of the smart items


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/EntityInformation/AttributeXYZ.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    public event Action<Vector3> OnChanged;
+
+    Vector3 currentValue;
+
+    bool isSelected = false;
+    
+    public void SetValues(Vector3 value)
+    {
+        if (!isSelected)
```
BrianAmadori: We can invert this if statement to reduce nesting

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/EntityInformation/EntityInformationController.cs
```diff
@@ -0,0 +1,201 @@
...
+        {
+            framesCount++;
+        }
+    }
+
+    public void ToggleDetailsInfo()
+    {
+        detailsGO.SetActive(!detailsGO.activeSelf);
+        Vector3 angle = detailsBtn.rotation.eulerAngles;
+        angle.z += 180;
```
BrianAmadori: What's this for? If this is for inverting some arrow or something like that, I think we should handle it better (by toggling on some object and off another). 

With this approach, Romi or any UI designer is going to be unable to edit the visuals without modifying the code.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/EntityInformation/AttributeXYZ.cs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public void ChangeZValue(string value)
+    {
+        if (!isSelected)
+            return;
+
+        currentValue.z = float.Parse(value);
+        OnChanged?.Invoke(currentValue);
+    }
```
BrianAmadori, line 72: These methods are used somewhere? I don't see any reason to keep them if we are just going to use the more comfy `SetValues` method.

BrianAmadori (changes requested):
Nice work! just left a few comments. What does `biw` mean?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldEntityHandler.cs
```diff
@@ -40,6 +41,24 @@ public class BuilderInWorldEntityHandler : MonoBehaviour
...
+        HUDController.i.buildModeHud.OnDuplicateSelectedAction += DuplicateSelectedEntitiesInput;
+        HUDController.i.buildModeHud.OnDeleteSelectedAction += DeleteSelectedEntitiesInput;
+        HUDController.i.buildModeHud.OnEntityClick += ChangeEntitySelectionFromList;
+        HUDController.i.buildModeHud.OnEntityLock += ChangeEntityLockStatus;
+        HUDController.i.buildModeHud.OnEntityChangeVisibility += ChangeEntityVisibilityStatus;
+        HUDController.i.buildModeHud.OnEntityRename += ChangeEntityName;
+
+        HUDController.i.buildModeHud.OnSelectedObjectPositionChange += UpdateEntityPosition;
+        HUDController.i.buildModeHud.OnSelectedObjectRotationChange += UpdateEntityRotation;
+        HUDController.i.buildModeHud.OnSelectedObjectScaleChange += UpdateEntityScale;
```
AjimenezDCL: I'm worried about coupling the `HUDController` too much. We should try to move this to a data driven approach in the future. I would even prefer an ugly `FindObjectOfType` (assuming `buildModeHud` is a `Monobehaviour`).

Anyway, let's leave it for another PR!

AjimenezDCL (dismissed):
Just left some concerns but not to be addressed in this PR. Good job!

D4rWiNSS:
> Nice work! just left a few comments. What does `biw` mean?

It means builder-in-world. It just to short the name, but I can use the full name if you want

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/EntityInformation/AttributeXYZ.cs
D4rWiNSS: Yes they are used from the Unity Inspector - `TMP_InputField` `ChangeValue ` They are called every time that a user made a change on the input so we can update the value accordly, same for the `InputSelected `and `InputDeselected`

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldEntityHandler.cs
D4rWiNSS: It is not a MonoBehaviour, however, I agree that right now the HUDController have a lot of responsibility, I will like to talk about the direction of this controller

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/EntityInformation/AttributeXYZ.cs
AjimenezDCL: Can we have these events explicitly subscribed in the code then? Linking them in Unity makes tracking references really difficult.

D4rWiNSS: Sure! However, this will limit the way a designer can assign that kind of behavior

pravusjif (dismissed):
It looks good to me, but I'd wait for the requesters to check on you final changes before merging

BrianAmadori (dismissed):

pravusjif:
Merged into master using an intermediary branch (because of the non-signed commits problem)

PR #1854, opened 2020-12-18
fix: make camera target position relative

### LINKED ISSUE #1777: MovePlayerTo function has Camera Position using global position
Someone from the community using the SDK reported that in movePlayerTo(), the second field, camera target, is using a global position, not relative to the scene's 0,0 but to the world's 0,0.
We should always use these things relative to the scene. Especially since the other field in this function is using the pos relative to the scene, so it's very inconsistent.

So if I have the scene deployed in 0,0,  and I teleport the player to 0,0,0 and the camera target to 0,0,16 the teleported player would look north.
But if I have that same scene deployed on parcel 0,50, then the teleported player will look south, to the global 0,0,16 position, near the center of the map.

nchamo:
The `movePlayerTo` action, takes both a position and a camera rotation.

The problem is that the position was relative, but the `cameraTarget` had to be global for it to work. We are now fixing that

Fixes https://github.com/decentraland/explorer/issues/1777


caminoezequiel (approved):

nchamo:
This PR should be merged on 2021, after all XMas events are over

PR #1855, opened 2020-12-18
chore: debug dump renderer locked status

pbosio:

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Bridges/DebugBridge.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Bridges/DebugBridge.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using DCL.Components;
+using DCL.Helpers;
 using UnityEngine;
 using UnityEngine.Serialization;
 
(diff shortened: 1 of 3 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Bridges/DebugBridge.cs
```diff
@@ -52,5 +53,31 @@ public void SetDisableAssetBundles()
...
+
+            Debug.Log($"Renderer is locked? {!renderingController.renderingActivatedAckLock.isUnlocked}");
+            Debug.Log($"Renderer is active? {CommonScriptableObjects.rendererState.Get()}");
+
+            System.Collections.Generic.HashSet<object> lockIds =
+                renderingController.renderingActivatedAckLock.GetLockIdsCopy();
+
+            foreach (var lockId in lockIds)
+            {
+                Debug.Log($"Renderer is locked by id: {lockId} of type {lockId.GetType()}");
```
BrianAmadori, line 77: Maybe we can find a way to have a descriptive string for the locks?

A few ways to do this:
- The locks can work with an interface with `string GetLockDescription()` method instead of an `object`.
- change `AddLock` signature to receive an optional `description` string and change the internals to be able to get that description.

BrianAmadori (approved):
Approved with an optional suggestion to improve the feedback.

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Bridges/DebugBridge.cs
pbosio: I was about to use the type of the locker and cast it to get more info... but right now we don't have that much lockers, so I think it doesn't worth the effort right now... and most of the` lockId.ToString() ` are very descriptives, specially scenes

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Bridges/DebugBridge.cs
BrianAmadori: Fair enough

PR #1875, opened 2020-12-21
fix: animation culling issue

pravusjif:
some animations are being culled even though they look close enough to avoid culling, it's probably due to a far away pivot point.

The problem is described here: https://app.zenhub.com/workspaces/explorer-5dc988c462b0b700019b0b42/issues/decentraland/explorer/1874

### DIFF
--- a/unity-client/Assets/Rendering/Culling/CullingControllerSettings.cs
+++ b/unity-client/Assets/Rendering/Culling/CullingControllerSettings.cs
@@ -11,7 +11,7 @@ namespace DCL.Rendering
         [NonSerialized] public float maxTimeBudget = 4 / 1000f;
         public bool enableObjectCulling = true;
         public bool enableShadowCulling = true;
-        public bool enableAnimationCulling = true;
+        public bool enableAnimationCulling = false;
 
         public float enableAnimationCullingDistance = 7.5f;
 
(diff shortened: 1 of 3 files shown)


AjimenezDCL (approved):

BrianAmadori (approved):

unity-client/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -308,9 +307,15 @@ internal IEnumerator ProcessAnimations()
...
 
-                if (distance > settings.enableAnimationCullingDistance)
+                Bounds bounds = anim.localBounds;
+                bounds.center += t.position;
+
+                Vector3 boundingPoint = bounds.ClosestPoint(playerPosition);
+                float distance = Vector3.Distance(playerPosition, boundingPoint);
+                bool boundsContainsPlayer = bounds.Contains(playerPosition);
+
+                if (distance > settings.enableAnimationCullingDistance && !boundsContainsPlayer)
```
BrianAmadori, line 318: After all we were checking the distance against the pivot. 

Now the culling works if:
- The distance against the animation bounding box nearest point is above a limit
- The player is not inside the bounds. i.e.: If the player is inside the bounds the animation will ALWAYS be played.

PR #1878, opened 2020-12-22
fix: calculate display name exception

caminoezequiel:


BrianAmadori (approved):
Maybe we can rename this PR to `fix: calculate display name exception` before merging.

PR #1882, opened 2020-12-22
fix: getAccounts in preview mode

caminoezequiel:
#WHY
We were executing `getUserAccount` without waiting to user connect her wallet


pablitar (approved):
LGTM!

kernel/packages/shared/ethereum/provider.ts
```diff
@@ -64,7 +64,7 @@ export function getProviderType() {
 }
 
 export async function awaitWeb3Approval(): Promise<void> {
-  if (!WORLD_EXPLORER) {
```
menduz, line 67: @caminoezequiel how does this affect the preview mode?

caminoezequiel: In preview mode without `ENABLE_WEB3` flag we are auto signing as GUEST. I changed it to prevent two authentications, but when builder is running we need to setup a `WebSocketProvider`

PR #1886, opened 2020-12-28
Added debug info when an error occurs while sending messages to Unity.

marcosnc:
# What? 
...

# Why? 
...


menduz (approved):

kernel/packages/unity-interface/UnityInterface.ts
```diff
@@ -510,6 +511,36 @@ export class UnityInterface {
...
+    try {
+      this.gameInstance.SendMessage(object, method, payload)
+    } finally {
+      restoreSetThrew()
+    }
+
+    if (isError) {
+      const error = `Error while sending Message to Unity. Object: ${object}. Method: ${method}. Payload: ${payload}.`
+      defaultLogger.error(error)
+      ReportRendererInterfaceError(error, error)
```
pbosio, line 550: @menduz are all console errors reported automatically to rollbar? I'm not sure if calling this will result in the error being reported twice :S

menduz: Based on what I read, rollbar only handles uncaught exceptions and literal calls to Rollbar.error

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -225,125 +225,142 @@ public bool ProcessMessage(MessagingBus.QueuedSceneMessage_Scene msgObject, out
...
-
-#if UNITY_EDITOR
-                OnMessageProcessInfoEnds?.Invoke(sceneId, method);
-#endif
-
-                res = true;
             }
-
-            else
+            catch (Exception e)
```
menduz, line 359: Do this preserves the stack trace?

pbosio: Looks like it does. I just tested it in unity editor and stacktrace was ok.
In browser is also more understandable as well.

With the try/catch:
Screen Shot 2021-01-03 at 18 15 52

Without it:
Screen Shot 2021-01-03 at 18 26 35

kernel/packages/unity-interface/UnityInterface.ts
```diff
@@ -513,6 +513,11 @@ export class UnityInterface {
   }
 
   private SendMessageToUnity(object: string, method: string, payload: any = undefined) {
+    if (!this.Module) {
```
menduz, line 520: Could you be more verbosic about this? or probably adding a comment like `"this.Module" is not present in using remote websocket renderer`?

It is very hard to understand that if without reading the commit message

menduz (changes requested):

menduz (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/SceneController.cs
```diff
@@ -225,125 +225,142 @@ public bool ProcessMessage(MessagingBus.QueuedSceneMessage_Scene msgObject, out
...
+                        if (msgPayload is Protocol.OpenNftDialog payload)
+                            OnOpenNFTDialogRequest?.Invoke(payload.contactAddress, payload.tokenId,
+                                payload.comment);
+                        break;
+                    }
 
                     default:
-                        Debug.LogError($"Unknown method {method}");
-                        return true;
+                        throw new Exception($"Unknown method.");
```
BrianAmadori: I wouldn't throw an exception here. If we are testing an outdated unity build against a new kernel build that implements a new type of message, it's more dev-friendly if we just fail gracefully and output the error in console. (already had this use case with the builder in world and it helped).

pbosio: 👍 done

BrianAmadori (approved):

PR #1887, opened 2020-12-28
fix: voice chat undefined setremotedescription

pravusjif:
error: https://rollbar.com/decentraland/World/items/11398/?item_page=4&item_count=100&#traceback

possible hint of a solution: https://github.com/webRTC-io/webrtc.io-client/issues/30#issuecomment-15991572

### DIFF
--- a/kernel/packages/voice-chat-codec/VoiceCommunicator.ts
+++ b/kernel/packages/voice-chat-codec/VoiceCommunicator.ts
@@ -289,6 +289,12 @@ export class VoiceCommunicator {
     const src = new RTCPeerConnection()
     const dst = new RTCPeerConnection()
 
+    // Apparently the RTCPeerConnection can be 'undefined' until the user accepts/blocks the microphone usage in the browser (https://github.com/webRTC-io/webrtc.io-client/issues/30#issuecomment-15991572)
+    // @ts-ignore
+    if (dst === 'undefined' || src === 'undefined') {
+      return this.createRTCLoopbackConnection()
+    }
+
     let retryNumber = currentRetryNumber
 
     ;(async () => {
--- a/kernel/packages/voice-chat-codec/VoiceCommunicator.ts
+++ b/kernel/packages/voice-chat-codec/VoiceCommunicator.ts
@@ -326,7 +326,6 @@ export class VoiceCommunicator {
           dst.setLocalDescription(answer)
           src.setRemoteDescription(answer)
         })
-        .catch((e) => console.error(e))
     })().catch((e) => {
       defaultLogger.error('Error creating loopback connection', e)
     })
--- a/kernel/packages/voice-chat-codec/VoiceCommunicator.ts
+++ b/kernel/packages/voice-chat-codec/VoiceCommunicator.ts
@@ -293,7 +293,7 @@ export class VoiceCommunicator {
       return this.createRTCLoopbackConnection(retryNumber + 1)
     }
 
-    ;(async () => {
+    (async () => {
       // When having an error, we retry in a couple of seconds. Up to 10 retries.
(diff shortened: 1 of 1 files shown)


pablitar:
This is unrelated to the issue.

See the stack trace:
image

I think the issue is related to the use of an unsupported browser or something similar. In any case, in theory there is code that should reduce the noise generated by the issue. See here:

https://github.com/decentraland/catalyst/pull/244/files

And peer library has in theory been updated, so it shouldn't show up anymore:

https://github.com/decentraland/explorer/commit/3dec2aaf87eacfc2e0a384b714ccc32a1f39944b

See for instance that the stack trace in rollbar is not the same as the latest code. For some reason, those who are experiencing the issue seem to not be running the latest code.

kernel/packages/voice-chat-codec/VoiceCommunicator.ts
```diff
@@ -283,28 +283,27 @@ export class VoiceCommunicator {
...
-    currentRetryNumber: number = 0
-  ): { src: RTCPeerConnection; dst: RTCPeerConnection } {
+  private createRTCLoopbackConnection(retryNumber: number = 0): { src: RTCPeerConnection; dst: RTCPeerConnection } {
     const src = new RTCPeerConnection()
     const dst = new RTCPeerConnection()
 
-    let retryNumber = currentRetryNumber
+    // Apparently the RTCPeerConnection can be 'undefined' until the user accepts/blocks the microphone usage in the browser (https://github.com/webRTC-io/webrtc.io-client/issues/30#issuecomment-15991572)
+    // @ts-ignore
+    if (!dst || !src) {
```
menduz, line 292: This condition will **_always_** return false.

There is NO WAY in that calling `new X` returns other than an object. It was a misinterpretation of the comment in the link.

That `if` should be removed

kernel/packages/voice-chat-codec/VoiceCommunicator.ts
```diff
@@ -317,20 +316,26 @@ export class VoiceCommunicator {
 
       const offer = await src.createOffer()
 
-      await src.setLocalDescription(offer)
+      await src.setLocalDescription(offer).catch((err) => {
+        return Promise.reject(err)
```
menduz, line 320: The following two snippets are exactly the same:

```ts
await src.setLocalDescription(offer)
```

```ts
await src.setLocalDescription(offer).catch((err) => {
  return Promise.reject(err)
})
```

We should remove the `.catch` to avoid confussions

kernel/packages/voice-chat-codec/VoiceCommunicator.ts
```diff
@@ -317,20 +316,26 @@ export class VoiceCommunicator {
...
+      })
 
-      await dst.setRemoteDescription(offer)
+      await dst.setRemoteDescription(offer).catch((err) => {
+        return Promise.reject(err)
+      })
```
menduz, line 325: same here

kernel/packages/voice-chat-codec/VoiceCommunicator.ts
```diff
@@ -317,20 +316,26 @@ export class VoiceCommunicator {
...
 
-      await dst.setLocalDescription(answer)
+      await dst.setLocalDescription(answer).catch((err) => {
+        return Promise.reject(err)
+      })
 
-      await src.setRemoteDescription(answer)
+      await src.setRemoteDescription(answer).catch((err) => {
+        return Promise.reject(err)
+      })
```
menduz, line 338: and here too

menduz (changes requested):
I'd recommend to close this PR since it adds no functionality

PR #1893, opened 2020-12-30
chore: add kernel test scenes watch

pravusjif:
added new makefile command `make watch-only-test-scenes` for watching kernel test scenes


menduz (approved):
Can you add a line in the README regarding this?

PR #1896, opened 2021-01-04
fix: collectible info asset destruction error

pravusjif:
Added check to avoid trying to destroy the preloaded (not-instantiated) asset sprite as that triggers an error in unity.

Added a placeholder sprite check in case we decide to pool these instances some day (Currently, they are not being recycled)

This error was seen a lot during the new years eve event:

Screen Shot 2020-12-30 at 20 56 49

Screen Shot 2020-12-30 at 21 04 48

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PlayerInfoCardHUD/PlayerInfoCollectibleItem.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PlayerInfoCardHUD/PlayerInfoCollectibleItem.cs
@@ -49,7 +49,7 @@ public class PlayerInfoCollectibleItem : MonoBehaviour
     private void OnThumbnailReady(Asset_Texture texture)
     {
         // we override the previously stored placeholder image (a referenced asset), we don't destroy it as it
-        // references the asset and it will provoke a "Destroying assets it not permitted to avoid data loss" error
+        // references the asset and it will provoke a "Destroying assets is not permitted to avoid data loss" error
         thumbnail.sprite = ThumbnailsManager.CreateSpriteFromTexture(texture.texture);
     }
 }
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PlayerInfoCardHUD/PlayerInfoCollectibleItem.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PlayerInfoCardHUD/PlayerInfoCollectibleItem.cs
@@ -8,9 +8,11 @@ public class PlayerInfoCollectibleItem : MonoBehaviour
 
     internal WearableItem collectible;
     private AssetPromise_Texture thumbnailPromise;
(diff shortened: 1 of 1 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PlayerInfoCardHUD/PlayerInfoCollectibleItem.cs
```diff
@@ -48,7 +50,8 @@ private void ForgetThumbnail()
 
     private void OnThumbnailReady(Asset_Texture texture)
     {
-        if (thumbnail.sprite != null)
+        // we can't destroy the referenced asset placeholder sprite or we get the "Destroying assets it not permitted to avoid data loss" error
+        if (thumbnail.sprite != null && thumbnail.sprite != placeholderSprite)
             Destroy(thumbnail.sprite);
 
         thumbnail.sprite = ThumbnailsManager.CreateSpriteFromTexture(texture.texture);
```
pbosio: I think that probably we should change `Image thumbnail` here for `RawImage thumbnail` so we don't have to create a Sprite and we can assign the texture directly. Also we are using an `AssetPromise_Texture` to get the texture but we never call `Forget`...
So I imagine a RawImage will solve both of the issues

pravusjif: I'll answer here since slack is not working...

* when the collectible info is loaded with the real texture (it has a referenced texture asset as placeholder until then), we use the asset promise, and then we do call the `forget` [here](https://github.com/decentraland/explorer/blob/8555602cfb056f1b3a53ee546d89566e024b08c9/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PlayerInfoCardHUD/PlayerInfoCollectibleItem.cs#L48) that ends up calling [this](https://github.com/decentraland/explorer/blob/8555602cfb056f1b3a53ee546d89566e024b08c9/unity-client/Assets/Scripts/MainScripts/DCL/Helpers/ThumbnailsManager/ThumbnailsManager.cs#L37)

* The `forget` is not called inside `OnThumbnailReady()` because the previous existent sprite is the placeholder one (I added the caching and checking there in case we decide to pool these player info collectible items in the future, which we are not doing right now, we don't recycle them, we create and destroy them on demand)

* I did a quick test replacing the `Image` component of one of the collectibles prefabs (there is 1 for every rarity tier) with a `RawImage` and we lose certain settings like the `image type: sliced`, `fill center` and its configurations. The image in the right is the RawImage one:
Screen Shot 2021-01-04 at 13 39 15

So if we replace the Image in those prefabs with a RawImage we should first create a new image/sprite that works for the RawImage component

pbosio (approved):

pravusjif:
Changed the approach, since we don't recycle the `CollectibleInfoAsset` instances, the problematic texture destruction is not needed at all.

PR #1906, opened 2021-01-06
fix: avatar editor culling not rendering wearables

pbosio:
What?
set culling manager to ignore `CharacterPreview` layer.

Why?
In some situations when opening avatar editor wearables are not show on the avatar.

I could reproduce it in most cases doing the following:
https://play.decentraland.zone/?ALL_WEARABLES&ENV=zone&position=141%2C-1&realm=hermes-amber
1- enter as guest as a new profile
2- create your profile
3- enter the world
4- open avatar editor

if your see your avatar broken in the editor you can switch to this build to see it fixed:
https://play.decentraland.zone/branch/fix/avatar-editor-culling/index.html?ALL_WEARABLES&ENV=zone&position=141%2C-1&realm=hermes-amber

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/CharacterPreviewController.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/CharacterPreviewController.cs
@@ -54,6 +54,9 @@ public class CharacterPreviewController : MonoBehaviour
             {CameraFocus.FaceSnapshot, faceSnapshotTemplate},
             {CameraFocus.BodySnapshot, bodySnapshotTemplate},
         };
+        var cullingSettings = DCL.Environment.i.platform.cullingController.GetSettingsCopy();
+        cullingSettings.ignoredLayersMask |= 1 << CHARACTER_PREVIEW_LAYER;
+        DCL.Environment.i.platform.cullingController.SetSettings(cullingSettings);
     }
 
     public void UpdateModel(AvatarModel newModel, Action onDone)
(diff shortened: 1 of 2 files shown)


pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/CharacterPreviewController.cs
```diff
@@ -54,6 +54,9 @@ private void Awake()
             {CameraFocus.FaceSnapshot, faceSnapshotTemplate},
             {CameraFocus.BodySnapshot, bodySnapshotTemplate},
         };
+        var cullingSettings = DCL.Environment.i.platform.cullingController.GetSettingsCopy();
+        cullingSettings.ignoredLayersMask |= 1 << CHARACTER_PREVIEW_LAYER;
+        DCL.Environment.i.platform.cullingController.SetSettings(cullingSettings);
```
AjimenezDCL, line 59: Wouldn't it be better to have this in the CullingController initialization?

Approving anyway!

AjimenezDCL (approved):
Good job!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/CharacterPreviewController.cs
pbosio: I thought about that in a first approach. My main concern was that unity force you to ask for the layer value using a string, this means that if in the future we change the name the most probable scenario is that we forget to change it on the CullingController and we end up we this issue again :P
I couldn't use the constant in this class cause to do that cause that will require to reference Avatar Editor asmdef in the Culling Controller asmdef.
So the other viable option was to use the Config script for that, but I found it kind of an overkill for something so specific

PR #1907, opened 2021-01-06
chore: request own profile update on profileHUD open

pravusjif:


sandrade-dcl (approved):
Only curiossity: why this change was needed?

pravusjif:
> Only curiossity: why this change was needed?

With the new builder release, people can claim more than 1 name from their interface and change their name when they want, with this PR the name will change in real time (although it has some seconds delay) when opening the profile hud

pbosio (approved):

pablitar (approved):
LGTM!

PR #1909, opened 2021-01-07
fix: green blockers in preview mode

pravusjif:


AjimenezDCL (approved):
Do we know the reason the green blockers started appering? I'm sure this will solve the issue but I'm worried we are not addressing the underlying cause.

In any case I am approving!

sandrade-dcl (approved):

pravusjif:
> Do we know the reason the green blockers started appering? I'm sure this will solve the issue but I'm worried we are not addressing the underlying cause.
> 
> In any case I am approving!

Apparenlty only in build sometimes the "set debug mode" that comes from kernel comes in a different order, during initialization and before we activate the rendering, and it only messes with the green blockers

PR #1916, opened 2021-01-08
chore: Refactorize GeneralSettingsController and QualitySettingsController

sandrade-dcl:
### Why?
After the implementation of the new **Settings Panel** and having each setting control separated and working with its own controller, which contains all the logic related to its behaviour, a second step to implement is to refactorize the classes `GeneralSettingsController` and `QualitySettingsController` (these classes contain all the logic related to the application of ALL the settings and it is very coupled in only 2 classes) in order to move each piece belonging to each specific control to its corresponding controller.

In this way these 2 classes would only contain the needed external references and all the contained logic (behaviour and aplication of the setting) would be moved to each specific control. It would make the system more scalable and would comply with the Single Responsibility Principle.

### Changes included in this PR
- Now `GeneralSettingsController` and `QualitySettingsController` classes only contain the scene references (for example: environment light, post-process volume, 1st and 3rd person cameras, etc.) that will be needed in any moment by the different settings.
- All the logic that was contained in `GeneralSettingsController` and `QualitySettingsController` has been moved to each specific settings control. In this way all the logic belong to a specific setting is encapsulated in its corresponding controller (before this refactor, the logic was separated between GeneralSettingsController, QualitySettingsController and the specific controller).
- SettingsPanelHUD improvements: Now the controls update is working better. Each time a setting is changed in the Settings class, the UI control is automatically updated without the needed of manage manual events.
- All the tests that was contained in `GeneralSettingsShould` class have been separated (one for each settings control) and moved to `SettingsControlsShould` class.
- General code cleaning.

### ADR updated
The corresponding ADR document has been updated in this **[PR](https://github.com/decentraland/adr/pull/7)**.

### Testing link
https://explorer.decentraland.zone/branch/chore/Refactorize-GeneralSettingsController-and-QualitySettingsController-new/index.html?ENV=org


pbosio (commented):
Moving the settings implementation to the controllers means that we can't modify any setting if `SettingsHUDController` doesn't exist?

sandrade-dcl:
> Moving the settings implementation to the controllers means that we can't modify any setting if `SettingsHUDController` doesn't exist?

Hmmm... yes, each specific controller depend on the SettingsPanel system. I hadn't thought about the possibility of changing settings in a context where there is no exist the SettingsPanel. Do you think it could be necessary?

If so, I could move back the logic from each controller to the `GeneralSettingsController` and `QualitySettingsController` classes but separating it in a method for each individual setting, instead of only one method that applies all the settings at once (what was there before).

Thanks for your annotation, Pato! I would like to discuss it with the team because it seems to be an important point.

pbosio:
> > Moving the settings implementation to the controllers means that we can't modify any setting if `SettingsHUDController` doesn't exist?
> 
> Hmmm... yes, each specific controller depend on the SettingsPanel system. I hadn't thought about the possibility of changing settings in a context where there is no exist the SettingsPanel. Do you think it could be necessary?
> 
> If so, I could move back the logic from each controller to the `GeneralSettingsController` and `QualitySettingsController` classes but separating it in a method for each individual setting, instead of only one method that applies all the settings at once (what was there before).
> 
> Thanks for your annotation, Pato! I would like to discuss it with the team because it seems to be an important point.

👍
In my opinion applying settings should be 100% independent from any UI controller
Plus we currently have the case of Builder that modifies settings but doesn't instantiates the SettingsHUDController

sandrade-dcl:
@pbosio I have just update the refactor implementation to move back the logic from each controller to the `GeneralSettingsController` and `QualitySettingsController` classes but separating it in a method for each individual setting. Now the settings could be applied from other scenarios where the settings HUD wouldn't be needed.

This also let me separate all the tests in 2 different sides: one for the settings applying and other one for UI controls behaviour 👍

pbosio (dismissed):

BrianAmadori (commented):
I agree with @pbosio comment. The settings apply should be completely decoupled from the view. But, I think that moving the settings application back to `GeneralSettingsController`/`QualitySettingsController` is not the proper way to go towards this direction. The problem with this is that we are still centralizing the settings code into those classes, and the maintenance for adding a new setting still depends on wiring those two classes into the controllers. These controllers will keep growing over time and the application domain will not match the design, as we can chose to split our settings into more categories than "General" and "Quality".

To improve this design, the setting controllers should be completely detached from the view existence so they can be applied without instancing the view.

After looking at the code, I can propose the following:

- Removing the `view`, `generalSettingsController` and `qualitySettingsController` references from each `SettingsControlController`.
- The sole responsibility of `SettingsControlController` should be to apply the setting and storing the current value for saving it later.
- controller-view relationship with `SettingsControlController` should be inverted using events, so it can be completely abstracted from the view. The views should subscribe to controller events to change state.
- Then, we can have a `SettingsControlController` collection on a composite class, so we can serialize the controller to save/load the data. Maybe we can do this on the `Settings` class.

After this is done, we should be able to just remove `IGeneralSettingsController` and `IQualitySettingsController` and at the same time keep the controllers abstracted from the views. With this approach, the settings would be modularized by the single responsibility principle. We could add new settings in a simpler way.

Just came with this proposal after looking quickly into the code, if the proposal doesn't match 100% I'm sure we can find another similar way to come up with an alternative that can let us go into the more modularized direction. 

What do you think?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/SettingsPanelHUD/Scripts/ControlsModule/SettingsControlView.cs
```diff
@@ -16,7 +17,11 @@ public interface ISettingsControlView
         /// </summary>
         /// <param name="controlConfig">Model that will contain the configuration of the CONTROL.</param>
         /// <param name="settingsControlController">Controller associated to the CONTROL view.</param>
-        void Initialize(SettingsControlModel controlConfig, SettingsControlController settingsControlController);
+        void Initialize(
```
BrianAmadori: I agree with @pbosio comment. The settings apply should be completely decoupled from the view. But, I think that moving the settings application back to `GeneralSettingsController`/`QualitySettingsController` is not the proper way to go towards this direction. The problem with this is that we are still centralizing the settings code into those classes, and the maintenance for adding a new setting still depends on wiring those two classes into the controllers. These controllers will keep growing over time and the application domain will not match the design, as we can chose to split our settings into more categories than "General" and "Quality".

To improve this design, the setting controllers should be completely detached from the view existence so they can be applied without instancing the view.

After looking at the code, I can propose the following:

- Removing the `view`, `generalSettingsController` and `qualitySettingsController` references from each `SettingsControlController`.
- The sole responsibility of `SettingsControlController` should be to apply the setting and storing the current value for saving it later.
- controller-view relationship with `SettingsControlController` should be inverted using events, so it can be completely abstracted from the view. The views should subscribe to controller events to change state.
- Then, we can have a `SettingsControlController` collection on a composite class, so we can serialize the controller to save/load the data. Maybe we can do this on the `Settings` class.

After this is done, we should be able to just remove `IGeneralSettingsController` and `IQualitySettingsController` and at the same time keep the controllers abstracted from the views. With this approach, the settings would be modularized by the single responsibility principle. We could add new settings in a simpler way.

Just came with this proposal after looking quickly into the code, if the proposal doesn't match 100% I'm sure we can find another similar way to come up with an alternative that can let us go into the more modularized direction. 

What do you think?

sandrade-dcl:
I have been able to completely decouple all the settings controllers from the entire HUD system:

- `SettingsControlController` no longer depends on any view and all the controllers for each specific setting can be used in an independent way (without the necessity of the HUD panel).
- Now `SettingsControlController` only have the responsibility of appling and saving the corresponding value for each setting.
- I have not deleted the `GeneralSettingsController` and `QualitySettingsController` classes but left them as 2 Monobehaviours with ONLY the scene references needed by certain controllers.
- Any interaction that is necessary between a view and its controller is now only done by events, so that the view would subscribe to the corresponding event of the controller.
- I have also moved all the settings controllers code out of the SettingsPanelHUD folder. Now they're placed into the Settings folder (outside the HUDs). It makes more sense to have it like that.

@BrianAmadori @pbosio could you take a look to this new approach? I think we already have what we wanted, but next monday I would like to confirm it with you and if everything is ok, we can merge :)

BrianAmadori (dismissed):
Great work!

pbosio (dismissed):

pbosio (approved):

pravusjif (approved):

PR #1923, opened 2021-01-11
refactor: migrate first test to init layer

BrianAmadori:
This PR adds the necessary modifications to integrate the `FontTests` suite to the initialization layer.

With this migration, the following is simplified:

- Tests scene load is not needed to run the tests anymore
- All the systems are mocked with the exception of the needed ones to run the tests. This make the tests way more reliable because new subsystems that are added to the test scene will not break them.
- State disposal between tests is way more reliable and involves less code

Relevant changes:

- `DCLComponentFactory` renamed to `RuntimeComponentFactory` and moved to its own assembly.
- Added `RuntimeComponentFactory` to world environment context.
- The component factory is injected to `SceneController` to take advantage of the environment mocking pattern.


AjimenezDCL (approved):
It looks awesome, I cannot wait to use it!

sandrade-dcl (approved):

PR #1928, opened 2021-01-12
fix: make camera report position from start

### LINKED ISSUE #1911: Camera.instance.position returns 0,0,0 the first time it's called
When you call `Camera.instance.position` for the first time in a scene, it returns the position as 0,0,0. It doesn't matter where the player has moved to, or if the scene has been loaded for a long time already when you call this.
If you call it a second time, it starts working well from there onwards in the scene.

This issue has been around for years, ever since we had Camera.instance.position. It has complicated consequences for any application that tries to leverage this. People end up having to do nasty workarounds to skip the first call whenever they want to use this.

nchamo:
Fixes https://github.com/decentraland/explorer/issues/1911

### DIFF
--- a/kernel/packages/decentraland-ecs/src/index.ts
+++ b/kernel/packages/decentraland-ecs/src/index.ts
@@ -17,8 +17,8 @@ import { DecentralandSynchronizationSystem } from './decentraland/Implementation
 import { Engine } from './ecs/Engine'
 import { Entity } from './ecs/Entity'
 
-const entity = new Entity('scene');
-(entity as any).uuid = '0'
+const entity = new Entity('scene')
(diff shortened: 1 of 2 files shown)


AjimenezDCL (approved):

menduz:
This PR generates another problem. 100% of the ECS scenes will subscribe to the events of the camera. That is, at least one message per frame, per let's say 25 scenes loaded in memory. 60*25=1500 messages generated and serialized and sent to the scenes even though they don't need it

A cleaner solution would be to expose a function like this

```ts

export async function getCameraInstance() {
  const instance = Camera.instance
  await nextFrame()
  return instance
}
```

I'd revert this PR

Issue #1913, opened 2021-01-07
Something's broken with Camera.instance.rotation.eulerAngles


nearnshaw:
This video shows it very clearly:

https://images.zenhubusercontent.com/220994476/c5e58380-fb2c-44a5-bd18-e39dedf921e4/screen_recording_2021_01_07_at_14_54_02.mov

Here I'm using the player's rotation and position to set a cube in the same position as the blue sphere.
Each time I click you can see in the console the result of logging `log(Camera.instance.rotation.eulerAngles.y)`, which I use to set the rotation and position of the cubes.
You can already see there that the values are weird and repeat themselves completely on one half of the circle.

Full code:

```ts
import utils from '../node_modules/decentraland-ecs-utils/index' /// --- Set up a system ---


let myCube: Entity

async function addMyCube() {
  myCube = await utils.addTestCube(
    { position: new Vector3(0, 0, 1) },
    () => {
      let worldPos = utils.getEntityWorldPosition(myCube)

      let cube = spawnCube(worldPos.x, worldPos.y, worldPos.z)
      cube.getComponent(Transform).rotation = getEntityWorldRotation(myCube)
    },
    '',
    Color3.Blue(),
    true
  )
  myCube.setParent(Attachable.AVATAR)
}

addMyCube()


export function getEntityWorldPosition(entity: IEntity): Vector3 {
    let entityPosition: Vector3 = entity.hasComponent(Transform)
      ? entity.getComponent(Transform).position.clone()
      : Vector3.Zero()
    let parentEntity = entity.getParent()

    if (parentEntity != null) {
		if( parentEntity.uuid == "FirstPersonCameraEntityReference"){
			//log('ATTACHED TO CAMERA')
			let parentRotation = Camera.instance.rotation.clone()
			return Camera.instance.position.clone().add(entityPosition.rotate(parentRotation))
		} else if(parentEntity.uuid == "AvatarEntityReference"){
			//log('ATTACHED TO AVATAR')
			let camRotation = Camera.instance.rotation
			let parentRotation = Quaternion.Euler(0, camRotation.eulerAngles.y, 0)
			log(Camera.instance.rotation.eulerAngles.y)
			return Camera.instance.position.clone().add(entityPosition.rotate(parentRotation)).add(new Vector3(0, -0.875, 0))
		} else {
			let parentRotation = parentEntity.hasComponent(Transform)
			? parentEntity.getComponent(Transform).rotation
			: Quaternion.Identity
[...]

PR #1935, opened 2021-01-14
fix: camera euler rotation gimbal lock

pravusjif:
This PR was triggered by the issue #1913 .

**WHY**
- Testing and comparing live euler values between kernel conversion and unity editor, our calculations solution often
got desynchronized with the real rotation and showed wrong values.

- Also, after reaching the euler singularity that's known as the Gimbal Lock (reaching +-90 degrees in Y euler rotation), the rotation values calculated where wrong (this can be seen when debugging with the camera's rotation, as stated in the issue).

- During this issue investigation I found other algebra implementations in our SDK that weren't compatible with Unity (they were more based on the previous engine used, babylon) so I updated those too, enabled tests and added more tests.

**WHAT**
- Updated our **quaternion-to-euler** conversion to cope with gimbal locking (the problem in issue #1913 ) and now we have in kernel/SDK the same euler values displayed in Unity editor.
- Updated **euler-to-quaternion** implementation
- Updated `Quaternion.fromToRotation()` implementation
- Updated `quaternion.setFromToRotation()` implementation
- Re-enabled kernel Vector3 and quaternion tests, adding more `setFromToRotation()` tests


pbosio (dismissed):

kernel/packages/decentraland-ecs/src/decentraland/math/Quaternion.ts
```diff
@@ -533,30 +533,28 @@ export class Quaternion {
...
-      out.z = RAD2DEG * Math.atan2(-m[3], m[0])
-    } else {
-      out.x = RAD2DEG * Math.atan2(m[5], m[4])
-      out.z = 0
-    }
+    // pitch / heading / y-axis rotation
+    out.y = RAD2DEG * Math.atan2(2 * (this.y * this.w - this.x * this.z) , 1 - 2 * (this.y * this.y - this.z * this.z))
+
+    // yaw / attitude / z-axis rotation
+    out.z = RAD2DEG * Math.asin(2 * (this.x * this.y + this.z * this.w))
```
menduz: With Dani we had maaany implementation of euler angles. I recall we wrote some tests, if those were lost in some migration. Could you create a couple of test cases for this? 

I recall one test that was auler->quaternion->euler to asses we are consistenly transforming the values. It was applied to several corner case numbers and those numbers also had to be the same as unity created

menduz (changes requested):
Looks good, please write some tests

kernel/packages/decentraland-ecs/src/decentraland/math/Quaternion.ts
```diff
@@ -526,37 +527,54 @@ export class Quaternion {
...
-    MathTmp.tmpMatrix = Matrix.Zero() // clean up preallocated matrix
-    Matrix.LookAtLHToRef(from, to, up, MathTmp.tmpMatrix)
-    MathTmp.tmpMatrix.invert()
-    Quaternion.FromRotationMatrixToRef(MathTmp.tmpMatrix, this)
+    const result = Quaternion.FromToRotation(from, to)
+    this.x = result.x
+    this.y = result.y
+    this.z = result.z
+    this.w = result.w
   }
```
pbosio: Is it OK for this not to use UP vector? rotation values shouldn't be different depending on UP vector?

pbosio (approved):

kernel/test/unityIntegration/ecs/math/quaternion.test.ts
```diff
@@ -33,16 +33,13 @@ const results = {
...
+  eulerAngles03: '(80.0, -170.0, -140.1)', // In Unity we get (80.0, 190.0, 220.0), but both represent the same rotation -> 360 - 170 = 190; 360 - 140 = 220;
+  eulerAngles04: '(-0.1, 10.0, 0.0)',
   normalized01: '(0.1, 0.1, 0.1, 1.0)',
   normalized02: '(0.0, 0.7, 0.0, 0.7)',
   normalized03: '(-0.7, 0.2, -0.2, -0.6)',
   normalized04: '(0.0, -0.1, 0.0, -1.0)',
-  setFromToRotation01: '(0.0, 0.0, 0.0, 1.0)',
-  setFromToRotation02: '(0.0, 0.7, 0.0, 0.7)',
-  setFromToRotation03: '(0.0, 0.0, 0.4, 0.9)',
-  setFromToRotation04: '(0.4, 0.0, 0.0, 0.9)'
```
menduz, line 45: Why were these deleted?

kernel/packages/decentraland-ecs/src/decentraland/math/Quaternion.ts
```diff
@@ -214,24 +214,25 @@ export class Quaternion {
...
+    const s3 = Math.sin(halfRoll)
 
-    result.x = s1 * c2 * c3 + c1 * s2 * s3
-    result.y = c1 * s2 * c3 - s1 * c2 * s3
-    result.z = c1 * c2 * s3 + s1 * s2 * c3
-    result.w = c1 * c2 * c3 - s1 * s2 * s3
+    result.x = c2 * s1 * c3 + s2 * c1 * s3
+    result.y = s2 * c1 * c3 - c2 * s1 * s3
+    result.z = c2 * c1 * s3 - s2 * s1 * c3
+    result.w = c2 * c1 * c3 + s2 * s1 * s3
```
menduz, line 236: This is the exact same calculation with the operands of the multiplications flipped. The only thing that chanes are the z and w componets addition and substraction which are swapped

kernel/packages/decentraland-ecs/src/decentraland/math/Quaternion.ts
pravusjif: fuck, you are right, OK I'll see what I can do

kernel/packages/decentraland-ecs/src/decentraland/math/Quaternion.ts
```diff
@@ -526,37 +527,54 @@ export class Quaternion {
...
+      out.z = Math.atan2(2 * this.w * this.z + 2 * this.x * this.y, 1 - 2 * (this.z * this.z + this.x * this.x))
     }
+    out.x *= RAD2DEG
+    out.y *= RAD2DEG
+    out.z *= RAD2DEG
+
+    // ensure the degree values are between 0 and 360
+    out.x %= 360
+    out.y %= 360
+    out.z %= 360
```
menduz: This normalization yields strange results, in % is not the modulus operation you may be wanting. It is only the rest of the integer division.

`-97 % 360 = -97` based on your comment you may be expecting it to be `263`

You may be wanting the modulo function:

```ts
function mod(n, p) {
    return n - p * Math.floor( n / p );
}
```

menduz (changes requested):

kernel/test/unityIntegration/ecs/math/quaternion.test.ts
pravusjif: we weren't using them, so there is no calculation showing the original vector nor the rotation that was applied...
I could create 3 new ones, but I already spent more than 1 week with this issue.

kernel/packages/decentraland-ecs/src/decentraland/math/Quaternion.ts
pravusjif: Yes i know, but it is exactly as used in the referenced calculations, I thought it would be less confusing to leave it that way.

kernel/packages/decentraland-ecs/src/decentraland/math/Quaternion.ts
menduz: YES it should take into account the up vector.

kernel/packages/decentraland-ecs/src/decentraland/math/Quaternion.ts
pravusjif: however the calculation works great using those % operations but I can test what you propose.

kernel/test/unityIntegration/ecs/math/quaternion.test.ts
menduz: Ok if not used. Anyways, I want to reinforce how important is this kind of changes and how much important are tests VS the time to ship the feature. It takes what it takes in order to have quality and not produce regressions.

pravusjif: Epic, I'll add at least 3 more tests on `setFromToRotation` then!

kernel/packages/decentraland-ecs/src/decentraland/math/Quaternion.ts
BrianAmadori: If this is just to clamp the angles between 0-360 it should work, BUT modulo doesn't work well for negative numbers, we should be wary about this. Also, using this operator may remove the decimals, and this will incur in loss of precision.

menduz: Both `mod(3.21, 2)` and `3.21 % 2` yield `1.21` there are no integers in JS. Are you looking to something I am not seeing? @BrianAmadori 

Negative numbers test:
`mod(-11.2, 360) = 348.8` = `360 - 11.2`
`mod(-1001.2, 360) = 78.8` = `-1001.2 + 360+ 360 + 360`

pravusjif: Ended up using the `Scalar.Repeat()` method (it does the same caculation as that mod function) because it returns values that are identical to Unity's. (instead of equivalent euler values like `310 and -50`)

kernel/packages/decentraland-ecs/src/decentraland/math/Vector3.ts
```diff
@@ -97,6 +97,7 @@ export class Vector3 {
    * @param normal - direction of the normal
    * @returns the angle between vector0 and vector1
    */
+  // Unity returns a different result -> unsigned, smallest angle: https://docs.unity3d.com/ScriptReference/Vector3.Angle.html
```
pbosio: This comment here doesn't break the method documentation? (when vscode intellisense displays the documentation of the method you are writing)

kernel/packages/decentraland-ecs/src/decentraland/math/Quaternion.ts
```diff
@@ -483,33 +484,24 @@ export class Quaternion {
 
   /**
    * Creates a rotation which rotates from fromDirection to toDirection.
-   * @param from - defines the first Vector
-   * @param to - defines the second Vector
+   * @param from - defines the first direction Vector
+   * @param to - defines the target direction Vector
+   * Unity-based calculations implemented from https://forum.unity.com/threads/quaternion-lookrotation-around-an-axis.608470/#post-4069888
```
pbosio: Can we leave unity related comment outside method documentation?

kernel/packages/decentraland-ecs/src/decentraland/math/Vector3.ts
pravusjif: Done

kernel/packages/decentraland-ecs/src/decentraland/math/Quaternion.ts
pravusjif: Done

pbosio (approved):

kernel/test/unityIntegration/ecs/math/quaternion.test.ts
```diff
@@ -2,23 +2,23 @@ import { expect } from 'chai'
...
-  staticFromToRotation07: '(0.1, 0.4, -0.1, 0.9)',
-  staticFromToRotation08: '(0.3, 0.0, -0.3, 0.9)',
+  staticFromToRotation01: '(0.0, 180.0, 0.0)',
+  staticFromToRotation02: '(0.0, 93.9, 0.0)',
+  staticFromToRotation03: '(0.0, 0.0, 44.5)',
+  staticFromToRotation04: '(45.0, 360.0, 360.0)',
+  staticFromToRotation05: '(335.2, 276.2, 152.5)',
+  staticFromToRotation06: '(39.1, 335.6, 352.5)',
+  staticFromToRotation07: '(12.8, 45.1, 351.9)',
+  staticFromToRotation08: '(35.3, 345.0, 315.0)',
```
menduz, line 21: I'm worried about a regression here, I trust you having this values checked. This actually deletes the tests for quaternions being generated correctly with the fromToRotation.. Want me to contribute with a commit?

pravusjif: Just for the record I'll state here what we talked offline:
- Those test results were changed to euler checks because in Unity we didn't get the exact same Quaternions, but got quaternions representing the same euler as the quaternions produced by our SDK. (I added a comment stating this but it's further below, where the test runs)
- I re-tested the gimbal lock fix working fine in the original test scene that triggered this investigation
- I tested objects rotation in the world to see that nothing broke, also playing The Salmonicon that has heavy usage of rotations (the monster chase, the monster speed if you look at it with a certain angle)

menduz (approved):

pravusjif:
I will rebase to sign the old unsigned commits before merging this

PR #1936, opened 2021-01-14
fix: motd fetch error triggering a fatal error

pbosio:
fetch exception for message of the day was not being handled.
a time-out was triggering a fatal error not letting the user enter the world.


pravusjif (approved):
nice catch!

sandrade-dcl (approved):

PR #1950, opened 2021-01-19
feat: add ignore list for console.error

pbosio:


BrianAmadori (approved):

pravusjif (approved):

menduz:
Is this the only way we have to do it?. I would merge this only as last resort

kernel/packages/unity-interface/setFilteredConsoleError.ts
```diff
@@ -0,0 +1,16 @@
...
+const ignoredErrors = ['Asset Bundle download is complete, but no data have been received\n']
+
+export function setFilteredConsoleError() {
+  if (isOverridden) return
+
+  isOverridden = true
+  const defaultConsoleError = window.console.error
+
+  window.console.error = function () {
+    if (!ignoredErrors.includes(arguments[0])) {
```
menduz, line 12: does this work if we write "console.error(null) or console.error(new Error())"?

pbosio: yeap, just tested that, it works as expected

D4rWiNSS (approved):

menduz:
The asset bundler now works, it has been a while since I've seen this error. Closing to avoid hacks

PR #1953, opened 2021-01-20
fix: gltf set curve keyframe optimizer issue

BrianAmadori:
Museum district GLTF animation importing was failing due to a weird keyframe issue on Unity's animation API. This is what caused the weird animation glitches and asset bundle conversion failures.

Fix applied according to the following Unity forum thread:
https://forum.unity.com/threads/animationutility-seteditorcurve-cant-add-curve-with-one-keyfram.247372/

NOTE: Right now, the museum district issue can't be reproduced because the offending robot was removed from the scene, but it should apply for future failures.


pbosio (approved):

pravusjif (approved):
tested animations in several scenes in prod with this build and didn't see any issue.

PR #1957, opened 2021-01-21
fix: fix color and avoid sending unnecessary information

nchamo:
We are making a few small changes:
**Color**
Eye, hair and skin colors are sometimes stored incorrectly:
```
"eyes": {
    "color": {
        "color": {
            "r": 0.37109375,
            "g": 0.22265625,
            "b": 0.1953125,
            "a": 1
        }
    }
}
```
We are fixing it

**Avoid sending unnecessary information**
We were sending data like `inventory` (which can be quite big) and `hasClaimedName` (which is calculated by lambdas) to the content server. This data needs to be fetched from the blockchain, it is never read from the content server. So we are not sending it any more


kernel/packages/shared/profiles/transformations/profileToServerFormat.ts
```diff
@@ -1,20 +1,18 @@
...
-  if (
+  const invalidWearables =
     !avatar.wearables ||
     !Array.isArray(avatar.wearables) ||
-    !avatar.wearables.reduce(
-      (prev: boolean, next: any) => prev && typeof next === 'string' && next.startsWith('dcl://'),
-      true
-    )
-  ) {
+    !avatar.wearables.reduce((prev: boolean, next: any) => prev && typeof next === 'string', true)
```
nchamo, line 14: We removed the check `next.startsWith('dcl://')` to consider wearables valid, since they won't have that format anymore

kernel/packages/shared/profiles/transformations/analizeColorPart.ts
```diff
@@ -1,15 +1,16 @@
...
-export function analizeColorPart(avatar: any, ...alternativeNames: string[]) {
+export function analizeColorPart(avatar: any, ...alternativeNames: string[]): Optional<ReadOnlyColor4, 'a'> {
   for (let name of alternativeNames) {
     if (!avatar[name]) {
       continue
     }
     if (typeof avatar[name] === 'string') {
       if (avatar[name].length === 7) {
-        return { color: convertToRGBObject(avatar[name]) }
+        return convertToRGBObject(avatar[name])
```
nchamo, line 10: In this cases we were returning something of the type:
```
{ color: { r, g, b} } 
```

Below, we were returning 
```
{ r, g, b}
```

Now we are always returning the second one

pablitar (approved):
LGTM!

PR #1958, opened 2021-01-21
feat: Added signed fetch as an SDK exposed API

pablitar:
This PR adds SignedFetch, an API that allows scenes to make authenticated requests to external URLs.

This will be used by scenes using the quests client, but also can be used for any other request.


kernel/packages/decentraland-ecs/types/@decentraland/QuestsController/index.d.ts
```diff
@@ -0,0 +1,107 @@
+declare module '@decentraland/QuestsController' {
+  export enum ProgressStatus {
```
menduz: .d.ts files cannot have enums (values). Use instead union types

kernel/packages/shared/apis/QuestsController.ts
```diff
@@ -0,0 +1,33 @@
+import { ExposableAPI } from './ExposableAPI'
+import { exposeMethod, registerAPI } from 'decentraland-rpc/lib/host'
+import { ClientResponse, PlayerQuestDetails, ProgressData } from 'dcl-quests-client'
+import { questsRequest } from '../quests/client'
+
+export type QuestsResponse<T> = Omit<ClientResponse<T>, 'status'>
+
+@registerAPI('QuestsController')
```
menduz: We used to expose publicly visible modules with an experimental prefix to warn everyone and cover ourselves to make improvements

kernel/packages/shared/quests/client.ts
```diff
@@ -0,0 +1,19 @@
...
+  return new QuestsClient({
+    baseUrl: servers.questsUrl,
+    authChainProvider: (payload) => Authenticator.signPayload(getIdentity()!, payload)
+  })
+}
+
+export async function questsRequest<T>(request: (client: QuestsClient) => Promise<ClientResponse<T>>) {
+  const client = await questsClient()
+  return await request(client)
+}
```
menduz, line 19: Nice

kernel/packages/decentraland-ecs/types/@decentraland/QuestsController/index.d.ts
nchamo: Actually, we can. But it has to be a `const` enum

https://github.com/decentraland/explorer/blob/f7225a16bb6b0cf4338af684a1cc11d2df25e8b2/kernel/packages/decentraland-ecs/types/@decentraland/RestrictedActions/index.d.ts#L23

menduz: the problem will come in this scenario:

```ts
import { ProgressStatus } from '@decentraland/QuestsController'

log(ProgressStatus.NOT_STARTED) 
// => `Error: cannot read property NOT_STARTED of undefined`
```

Because these modules are merely a proxy of an RPC object. And `ProgressStatus` is not only a type, it is a value.

kernel/package.json
```diff
@@ -99,6 +99,7 @@
     "cids": "^0.7.3",
     "dcl-catalyst-client": "^3.0.1",
     "dcl-crypto": "^2.2.0",
+    "dcl-quests-client": "^1.2.0",
```
pablitar, line 102: We are still using the quests client because eventually we will need it for the Quest Board (global quests list)

kernel/packages/shared/apis/SignedFetch.ts
```diff
@@ -0,0 +1,65 @@
...
+    const path = new URL(url).pathname
+    const actualInit = {
+      ...init,
+      headers: {
+        ...getAuthHeaders(
+          init?.method ?? 'get',
+          path,
+          { sceneId: this.parcelIdentity.cid, parcel: this.getSceneData().scene.base, tld: getTLD() },
+          (payload) => Authenticator.signPayload(identity, payload)
+        ),
```
menduz, line 55: How is this different than the ephemeral key helper? Is it compatible?

pablitar: I don't understand. What is the ephemeral key helper?

kernel/public/ecs-scenes/0.107.signed-fetch/game.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const text = new TextShape('Wait for it')
+text.billboard = true
+text.isPickable = true
+cube.addComponentOrReplace(text)
+
+engine.addEntity(cube)
+
+executeTask(async () => {
+  const questsResponse = await signedFetch('https://quests-api.decentraland.io/quests', { responseBodyType: 'json' })
```
pablitar, line 19: We are able to make this particular request without the client because this is not obfuscated.

I didn't test the client here since it is an external dependency.

kernel/packages/shared/apis/SignedFetch.ts
menduz: https://github.com/decentraland/ephemeralkey#usage-1

It is the library (already used by the explorer) used to sign and validate signatures

```ts
import { getHeaders } from 'ephemeralkey'

async function fetchWithEphemeralKey(request: RequestData): Promise<any> {
  const userData = JSON.parse(
    localstorage.getItem('ephemeral-data', JSON.stringify(userData))
  )
  const headers = getHeaders(userData, request)

  return fetch(request.url, {
    method: request.method,
    headers: {
      'Content-Type': 'application/json; charset=utf-8',
      ...headers
    },
    body: request.body.toString()
  })
}

const response = await fetchWithEphemeralKey({
  url: 'market.decentraland.org/api/v1/land',
  method: 'POST',
  timestamp: Date.now(),
  body: Buffer.from(JSON.stringify({ param1: 'param1', param2: 'param2' }))
})
```

That signed request could be validated by any of our servers to verify the auth chain

pablitar: >  (already used by the explorer)

Where?

We could start using it though. I didn't know it existed until now.

I based the authentication model on the builder server: https://github.com/decentraland/builder-server/blob/master/src/middleware/authentication.ts

menduz: Hmm it was removed in favor of `dcl-crypto#Authenticator` @nachomazzara do you know why we are no longer using ephemeralkey to generate the keys and sign?

pablitar: I don't see it being used that much, at least not in recent projects: https://github.com/search?p=3&q=org%3Adecentraland+from+ephemeralkey&type=Code

nachomazzara: If I remember well, `dcl-crypto` has replaced ephemeralkey because it has the possibility to create an auth chain where ephemeral is just a key/pair. It has started as a util inside the Catalyst repo, and then it was moved to its own as a library. 

CC @marcosnc or @nchamo. They may remember if it was another reason rather than the pointed above.

menduz: I'd say to merge this as long as it works and doesn't allow anyone to use it maliciously like @nchamo said on slack. We can figure out late the details of the headers functions. I will create a reminder.

An example of malicious request would be to execute a command in a scene that overrides the main plaza via deployment if the right person is executing the scene.

pablitar: We are forcibly including `sceneId`, `parcel`, `tld` in the signature, so this cannot be used for deployments, authenticating with comms, or the builder. And it probably cannot be used anywhere else.

Other than that, future services that allow this format should validate those parameters to be sure they are not abused. But previous services don't use this format, so we should be OK.

nchamo (approved):
Great work!

PR #1964, opened 2021-01-22
refactor: some more profile changes

nchamo:
Some more minor changes


kernel/packages/shared/dao/reducer.ts
```diff
@@ -122,7 +121,7 @@ function realmProperties(realm: Realm, configOverride: boolean = true): Partial<
   return {
     profileServer: FETCH_PROFILE_SERVICE && configOverride ? FETCH_PROFILE_SERVICE : domain + '/lambdas/profile',
     fetchContentServer: FETCH_CONTENT_SERVICE && configOverride ? FETCH_CONTENT_SERVICE : domain + '/lambdas/contentv2',
-    fetchMetaContentServer: FETCH_META_CONTENT_SERVICE && configOverride ? FETCH_META_CONTENT_SERVICE : domain,
+    catalystServer: domain,
```
pablitar, line 121: Aren't we removing functionality here?

kernel/packages/shared/dao/reducer.ts
```diff
@@ -122,7 +121,7 @@ function realmProperties(realm: Realm, configOverride: boolean = true): Partial<
   return {
     profileServer: FETCH_PROFILE_SERVICE && configOverride ? FETCH_PROFILE_SERVICE : domain + '/lambdas/profile',
```
pablitar: If I understood correctly, we are replacing this. Can we delete it?

pablitar (dismissed):
I'm like simplifying our configuration.

Just in case, please consider the comments :)

kernel/packages/shared/dao/reducer.ts
nchamo: Yeah, indeed. It's not overridable anymore, but I don't think it's being used. If we think we need it back, we can add it back in the future

kernel/packages/shared/dao/reducer.ts
nchamo: Good catch, deleting it

pablitar (approved):
LGTM!

PR #1967, opened 2021-01-26
chore: night render profile

pravusjif:
**WHY**
right now our **halloween** and **Xmas** render profiles set their respective empty parcels and they also use the **night mode** by default.

The problem is that we frequently need to enable the **night mode** by default without the change in the empty parcels, and we can't do that using the existing render profiles.

**WHAT**
Added a **night** render profile that only enables the night mode, without changing the empty parcels.
Now **Xmas** and **Halloween** render profiles use the same underlying settings applied for the new **night** render profile.

**TEST**
* `DEFAULT` -> default empty parcels, day mode
https://explorer.decentraland.zone/branch/chore/NightRenderProfile/index.html?ENV=org&FORCE_RENDERING_STYLE=0

* `HALLOWEEN` -> halloween empty parcels, night mode
https://explorer.decentraland.zone/branch/chore/NightRenderProfile/index.html?ENV=org&FORCE_RENDERING_STYLE=1

* `XMAS` -> Xmas empty parcels, night mode
https://explorer.decentraland.zone/branch/chore/NightRenderProfile/index.html?ENV=org&FORCE_RENDERING_STYLE=2

* `NIGHT` -> default empty parcels, night mode **(new)**
https://explorer.decentraland.zone/branch/chore/NightRenderProfile/index.html?ENV=org&FORCE_RENDERING_STYLE=3


pbosio (approved):
Maybe for later, but, Is it possible to decouple this a little bit more? If we want Halloween empty parcels with day mode we will find our self kind of in the same situation we are right now

pravusjif:
> Maybe for later, but, Is it possible to decouple this a little bit more? If we want Halloween empty parcels with day mode we will find our self kind of in the same situation we are right now

Yeap, we talked about this with Brian and we will decouple it completely in the future.

Also I'd like to have these kind of settings in the senes JSON so that we can lerp diferent render profiles when you enter/exit a scene (and we wouldn't be bothered with changing the defults of the world if the event developers can set this directly in their scenes...)

AjimenezDCL (approved):

BrianAmadori (approved):
Didn't test the build but it looks neat codewise.

PR #1959, opened 2021-01-21
feat: Portable Experiences implementation (base branch)

sandrade-dcl:
### Kernel side
- Implementation of the needed functions to spawn/kill portable experiences (as global scenes) from Kernel. There are 3 flows supported:
   - **spawn portable experience**: The action of starting a portable experience. This can be executed using the SDK from any scene or the `hacky-way`, that is running from the console: `spawnPortableExperienceScene("urn:decentraland:off-chain:static-portable-experiences:pe2", "main")` 
   - **kill portable experience**: The action of stoping a portable experience. This can be executed using the SDK from the scene that created the portable experience (the parent) or the `hacky-way`, that is running from the console: `killPortableExperienceScene("urn:decentraland:off-chain:static-portable-experiences:pe2", "main")` 
   - **exit portable experience**: The action of stoping a portable experience from the experience itself. This is like committing suicide.  This can be executed using the SDK from the code of the portable experience.
- Integration with S3 bucket that stores the portable experiences obtained from the [Assets Repository](https://github.com/decentraland/portable-experiences-assets).
- Integration with [URN-Resolver](https://github.com/decentraland/urn-resolver) to identify the portable experiences.

### Unity side
- Global scenes flow modified to allow the creation of multiple global scenes received from Kernel.
- A new "**Portable Experience Manager**" in the taskbar allows the user to see all the active PE's and kill them.
- Use of `currentSceneId` reviewed and adapted to take into account also the execution of the active PE scenes.

### Testing link
- Link: https://play.decentraland.zone/branch/feat/PortableExperiences-implementation-kernel-and-unity-side/index.html?ENV=org

### Some PE examples
You can execute any of these PE examples from the Chrome console:
- **A cube above you!**: `spawnPortableExperienceScene("urn:decentraland:off-chain:static-portable-experiences:pe1", "main")`
A cube will be instantiated above your avatar and will follow you to any place.

- **Find the Dragon**: `spawnPortableExperienceScene("urn:decentraland:off-chain:static-portable-experiences:pe2", "main")`
You will have to go to Dragon Rush scene and find the dragon before a minute. After the time countdown reachs zero, the PE is auto-destroyed.

- **Radio**: `spawnPortableExperienceScene("urn:decentraland:off-chain:static-portable-experiences:radio", "main")`
You will see an UI where you will be able to select several streaming radios and listen them. Pressing [F] you will show/hide this UI.

[...]

PR #1971, opened 2021-01-29
Add portable experiences API (extract PE from dcl.ts)

agusaldasoro:
# What? 
- Refactor in `dcl.ts` to extract code for handling portable experiences into a separate file: `portableExperiencesHelper.ts`
- Implementation for `spawn` and `kill` for SDK interface according to https://github.com/decentraland/proposals/pull/181/files


kernel/packages/shared/world/parcelSceneManager.ts
```diff
@@ -46,14 +46,6 @@ export function stopParcelSceneWorker(worker: SceneWorker) {
   }
 }
 
-export function stopPortableExperienceWorker(cid: string) {
```
agusaldasoro, line 49: I've moved this method to `dcl.ts`, is that expected?

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -465,7 +470,11 @@ export class BrowserInterface {
   }
 
   public KillPortableExperience(data: { portableExperienceId: string }) {
-    stopPortableExperienceWorker(data.portableExperienceId)
```
agusaldasoro, line 468: This code is duplicated from `dcl.ts` as I can't import the method here as it's a circular reference.

kernel/packages/shared/apis/PortableExperiences.ts
```diff
@@ -0,0 +1,58 @@
...
+   * @param  {SpawnPortableExperienceParameters} [spawnParams] - Information to identify the PE
+   *
+   * Returns the handle of the portable experience.
+   */
+  @exposeMethod
+  async spawn(spawnParams: SpawnPortableExperienceParameters): Promise<PortableExperienceHandle> {
+    await spawnPortableExperienceScene(spawnParams.portableExperienceId)
+    // TODO: Fill correctly the result
+    return {
+      pid: 'test',
```
agusaldasoro: which should be the value here?

kernel/packages/shared/apis/PortableExperiences.ts
```diff
@@ -0,0 +1,58 @@
...
+      identifier: spawnParams.portableExperienceId,
+      parentProcess: { type: ExecutorType.QUEST_UI, identifier: 'test' }
```
agusaldasoro: how do I obtain the correct value for the parent process?

kernel/packages/decentraland-ecs/types/@decentraland/PortableExperiences/index.d.ts
```diff
@@ -0,0 +1,26 @@
+declare module '@decentraland/PortableExperiences' {
+  enum ExecutorType {
+    SCENE = 'SCENE',
+    WEARABLE = 'WEARABLE',
+    QUEST_UI = 'QUEST_UI'
+  }
```
menduz: Enums are values, which cannot be used in .d.ts files, because enums are also values in typescript. Please use a union type 
```ts
type ExecutorType = 'SCENE' | 'WEARABLE' | 'QUEST_UI'
```

kernel/packages/decentraland-ecs/types/@decentraland/PortableExperiences/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+  type ContentIdentifier = string
+  type PortableExperienceIdentifier = string
+  type PortableExperienceHandle = {
+    pid: string
+    identifier: PortableExperienceIdentifier
+    parentProcess: Executor
+  }
+  type SpawnPortableExperienceParameters = {
+    contentIdentifier: ContentIdentifier
```
menduz: We should use URN here

```suggestion
    urn: ContentIdentifier
```

kernel/packages/shared/apis/PortableExperiences.ts
menduz: `this.getAPIInstance(ParcelIdentity).land` or something like that

kernel/packages/shared/apis/PortableExperiences.ts
menduz: Should be the scene id returned by `spawnPortableExperienceScene`

kernel/packages/shared/apis/PortableExperiences.ts
```diff
@@ -0,0 +1,58 @@
...
+   * Stops a portable experience. Only the executor that spawned the portable experience has permission to kill it.
+   * @param {Executor} [executor] - The executor that will stop the PE
+   * @param  {string} [pid] - The portable experience process id
+   *
+   * Returns true if was able to kill the portable experience, false if not.
+   */
+  @exposeMethod
+  async kill(pid: PortableExperienceIdentifier): Promise<boolean> {
+    killPortableExperienceScene(pid)
+    // TODO: Check if we want to return a boolean
```
menduz: definitely

kernel/packages/unity-interface/dcl.ts
```diff
@@ -138,9 +125,10 @@ export async function initializeEngine(_gameInstance: GameInstance) {
 
   if (!EDITOR) {
     await startGlobalScene(unityInterface, 'dcl-gs-avatars', 'Avatars', hudWorkerUrl)
+
     // Temporal: Try to create several global scenes
-    await startPortableExperienceScene(unityInterface, 'dcl-pe-example1', 'pe1')
-    await startPortableExperienceScene(unityInterface, 'dcl-pe-example2', 'pe2')
+    await spawnPortableExperienceScene('pe1')
+    await spawnPortableExperienceScene('pe2')
```
menduz: this should be removed before merge

kernel/packages/unity-interface/dcl.ts
```diff
@@ -185,8 +173,8 @@ export async function startGlobalScene(
   })
 }
 
-export async function startPortableExperienceScene(unityInterface: UnityInterface, cid: string, peId: string) {
-  const scene = new UnityPortableExperienceScene(await getPortableExperienceFromS3Bucket(cid, peId))
+export async function spawnPortableExperienceScene(peId: string) {
+  const scene = new UnityPortableExperienceScene(await getPortableExperienceFromS3Bucket(peId))
```
menduz:
```suggestion
  const parsedUrn = await parseUrn(portableExperienceUrn)
  if(!parsedUrn || isPortableExperience(parsedUrn)) return null // or throw
  const scene = new UnityPortableExperienceScene(await getPortableFromParsedURN(parsedUrn))
```

menduz (changes requested):

kernel/packages/unity-interface/dcl.ts
agusaldasoro: Sure, it'll be done in https://github.com/decentraland/explorer/pull/1959

kernel/packages/shared/apis/PortableExperiences.ts
```diff
@@ -0,0 +1,58 @@
...
+   * Returns the handle of the portable experience.
+   */
+  @exposeMethod
+  async spawn(spawnParams: SpawnPortableExperienceParameters): Promise<PortableExperienceHandle> {
+    const sceneId: string = await spawnPortableExperienceScene(spawnParams.portableExperienceId)
+    const parcelIdentity: ParcelIdentity = this.options.getAPIInstance(ParcelIdentity)
+
+    return {
+      pid: sceneId,
+      identifier: spawnParams.portableExperienceId,
```
menduz: what does `identifier` mean?

kernel/packages/shared/world/parcelSceneManager.ts
menduz: It could be a "portable experiences" special file/namespace

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -465,7 +470,11 @@ export class BrowserInterface {
...
+    const peWorker = getSceneWorkerBySceneID(data.portableExperienceId)
+
+    if (peWorker) {
```
menduz: We should add a validation here to verify that only the parent process can kill the portable experience, otherwise people would be able to kill any scene from the SDK

kernel/packages/unity-interface/dcl.ts
```diff
@@ -185,82 +174,41 @@ export async function startGlobalScene(
...
-        mappings: mappingsResponse.contents,
-        sceneJsonData: scene
-      })
-    } else {
-      throw new Error('Could not load scene.json')
-    }
-  } else {
-    throw new Error('Could not load scene.json')
-  }
+function isPortableExperience(dclId: DecentralandAssetIdentifier): boolean {
```
menduz: Type assertions to the rescue

```suggestion
function isPortableExperience(dclId: DecentralandAssetIdentifier): dclId is OffChainAsset {
```

It makes this `as unknown` no longer necessary

```ts
(parsedUrn as unknown) as OffChainAsset
```

kernel/packages/unity-interface/dcl.ts
```diff
@@ -185,82 +174,41 @@ export async function startGlobalScene(
...
+  const offChainAsset = (dclId as unknown) as OffChainAsset
+  return !!offChainAsset.registry && offChainAsset.registry === 'static-portable-experiences'
```
menduz: dclId can be null that should be verified before anything, as a rule of thumb, avoid as unknown at all costs

kernel/packages/unity-interface/portableExperiencesHelper.ts
```diff
@@ -0,0 +1,83 @@
...
+          sceneJsonData.scene.parcels &&
+          sceneJsonData.scene.parcels.map(parseParcelPosition)) ||
+        [],
+      baseUrl: baseUrl,
+      baseUrlBundles: '',
+      contents: mappings,
+      icon: sceneJsonData.display?.favicon
+    }
+  }
+}
```
menduz: cool file, lifecyle functions may live here too

kernel/packages/shared/apis/PortableExperiences.ts
```diff
@@ -0,0 +1,85 @@
...
+  pid: string
+  parentProcess: Executor
+}
+type SpawnPortableExperienceParameters = {
+  urn: PortableExperienceIdentifier
+}
+
+@registerAPI('PortableExperiences')
+export class PortableExperiences extends ExposableAPI {
+  private currentPortableExperiences: Map<string, Executor> = new Map()
```
menduz: Imagine this scenario:
- You go to 1,1
- LAND1,1 spawns "PE1"
- In order to complete the quest, you go to LAND150,150
- For distance reasons, scene in LAND1,1 gets unloaded
- In the process of scene unloading the state of `LAND1,1.PortableExperience.currentPortableExperiences` is garbage collected and emptied.
- You go back to 1,1, LAND1,1 is loaded again.
- LAND1,1 wants to kill the PE1 (it should be able to, since LAND1,1 created PE1)

I believe that having this map here couples the implementation to an ephemeral state.

As a solution to the previous example, I suggest to store the information of who spawned the portable experience in the portable experience itself. Like this

```ts
const portableExperienceScene = await spawnPortableExperienceScene(spawnParams.urn)
portableExperienceScene.spawnerCID = this.options.getAPIInstance(ParcelIdentity).cid
```

That way, all the comparisons will have a unique source of truth and the data would be more normalized (in database normalization terms).


That way, the implementation is very much straight forward without JSON parsing
```ts
async kill(pid: string){
  const peToKill = getPortableExperienceBySceneId(pid)
  const selfCID = this.options.getAPIInstance(ParcelIdentity).cid
  if (peToKill.spawnerCID == selfCID) {
    // can kill
  } else {
    // can't
  }
}
```

kernel/packages/shared/apis/PortableExperiences.ts
```diff
@@ -0,0 +1,85 @@
...
+  async exit(): Promise<boolean> {
+    const parcelIdentity: ParcelIdentity = this.options.getAPIInstance(ParcelIdentity)
+    const executorCid = parcelIdentity.cid
+
+    if (this.currentPortableExperiences.has(executorCid)) {
+      killPortableExperienceScene(executorCid)
+      this.currentPortableExperiences.delete(executorCid)
+      return true
+    }
+    return false
```
menduz: I think this condition is not really necessary as it is in `kill`, any portable experience should be able to perform a "suicide", like any other process.

```suggestion
    const parcelIdentity: ParcelIdentity = this.options.getAPIInstance(ParcelIdentity)
    killPortableExperienceScene(parcelIdentity.cid)
    return true
```

menduz (changes requested):

kernel/packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -39,7 +62,11 @@ export async function spawnPortableExperienceScene(portableExperienceUrn: Portab
     icon: scene.data.data.icon,
     isPortableExperience: true
   })
-  return parcelSceneId
+
+  const contentId = cid ?? ''
```
menduz: In which cases this is ""? Why do we support it instead of throwing an error?

kernel/packages/shared/apis/PortableExperiences.ts
```diff
@@ -53,13 +41,10 @@ export class PortableExperiences extends ExposableAPI {
...
   async kill(pid: string): Promise<boolean> {
     const parcelIdentity: ParcelIdentity = this.options.getAPIInstance(ParcelIdentity)
-    const currentExecutor: Executor = { type: ExecutorType.SCENE, identifier: parcelIdentity.cid }
-    if (
-      this.currentPortableExperiences.has(pid) &&
-      JSON.stringify(this.currentPortableExperiences.get(pid)) === JSON.stringify(currentExecutor)
-    ) {
+
+    const portableExperience = getPortableExperience(pid)
+    if (!!portableExperience && portableExperience.cid == parcelIdentity.cid) {
```
menduz: Shouldn't it be the parent cid?
```suggestion
    if (!!portableExperience && portableExperience.parentCid == parcelIdentity.cid) {
```

kernel/packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -20,7 +20,30 @@ import { parseUrn, DecentralandAssetIdentifier, OffChainAsset } from '@dcl/urn-r
...
+  if (currentPortableExperiences.has(pid)) {
+    return { pid: pid, cid: currentPortableExperiences.get(pid)! }
+  } else {
+    return undefined
+  }
+}
+
+export async function spawnPortableExperienceScene(
+  portableExperienceUrn: PortableExperienceUrn,
+  cid?: string
```
menduz:
```suggestion
  parentCid?: string
```

menduz (changes requested):

kernel/packages/unity-interface/portableExperiencesUtils.ts
agusaldasoro: Currently, all CID must be present, so the best would be to throw an error.

agusaldasoro: Currently, we are starting the PE in the main, that's why we didn't have a parent process.  I'll put the parentCid as mandatory and hardcode it in the main call, as it will be removed after testing.

kernel/packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -0,0 +1,162 @@
...
+  return false
+}
+
+export async function killPortableExperienceScene(portableExperienceUrn: PortableExperienceUrn): Promise<boolean> {
+  const parsedUrn: DecentralandAssetIdentifier | null = await parseUrn(portableExperienceUrn)
+  if (!parsedUrn || !isPortableExperience(parsedUrn)) {
+    throw new Error(`Could not parse portable experience from urn: ${portableExperienceUrn}`)
+  }
+  const sceneId: string = parsedUrn.id
+  const peWorker = getSceneWorkerBySceneID(sceneId)
```
menduz: 👾 The `sceneId` of the portable experience should be the whole URN. It will also prevent duplicates and it will prevent us from doing `urn:decentraland:off-chain:static-portable-experiences:${data.portableExperienceId}` everywhere.

kernel/packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -0,0 +1,162 @@
...
+  } else {
+    return false
+  }
+}
+
+export async function getPortableExperienceFromS3Bucket(pe: OffChainAsset) {
+  const peId: string = pe.id
+  const baseUrl: string = `${STATIC_PORTABLE_SCENES_S3_BUCKET_URL}/${peId}/`
+
+  const mappingsFetch = await fetch(`${baseUrl}mappings`)
```
menduz: We have a function to perform the URN to URL resolution https://github.com/decentraland/urn-resolver/blob/main/test/content-url.spec.ts#L12-L15

Using that function is good because it will enable us to change and extend the resolution of portable experiences beyond the static endpoint and removes hardcoded dependencies in this project (STATIC_PORTABLE_SCENES_S3_BUCKET_URL)

```suggestion
// import {resolveContentUrl} from '@dcl/urn-resolver'
const mappingsUrl = await resolveContentUrl(pe)
const baseUrl: string = (new URL('..', mappingsUrl)).toString()
const mappingsFetch = await resolveContentUrl(mappingsUrl)
```

kernel/packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -0,0 +1,162 @@
...
+    currentPortableExperiences.delete(sceneId)
+    unityInterface.UnloadScene(sceneId)
+    return true
+  } else {
+    return false
+  }
+}
+
+export async function getPortableExperienceFromS3Bucket(pe: OffChainAsset) {
+  const peId: string = pe.id
```
menduz: 👾 peId should be the whole URN

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -464,9 +465,10 @@ export class BrowserInterface {
     }
   }
 
-  public KillPortableExperience(data: { portableExperienceId: string }) {
-    stopPortableExperienceWorker(data.portableExperienceId)
-    unityInterface.UnloadScene(data.portableExperienceId)
+  public async KillPortableExperience(data: { portableExperienceId: string }): Promise<void> {
+    const portableExperienceUrn: PortableExperienceUrn = `urn:decentraland:off-chain:static-portable-experiences:${data.portableExperienceId}`
```
menduz: - the `portableExperienceId` naming is inconsistent in other parts of the code, is it the same as peId?
- I think the "peId" should be the whole URN 👾

kernel/packages/shared/apis/PortableExperiences.ts
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Stops a portable experience from the current running portable scene.
+   *
+   * Returns true if was able to kill the portable experience, false if not.
+   */
+  @exposeMethod
+  async exit(): Promise<boolean> {
+    const parcelIdentity: ParcelIdentity = this.options.getAPIInstance(ParcelIdentity)
+
+    const portableExperienceUrn: PortableExperienceUrn = `urn:decentraland:off-chain:static-portable-experiences:${parcelIdentity.cid}`
```
menduz: 👾 same here

kernel/packages/decentraland-ecs/types/@decentraland/PortableExperiences/index.d.ts
```diff
@@ -0,0 +1,16 @@
+declare module '@decentraland/PortableExperiences' {
+  type PortableExperienceIdentifier = string
+  type PortableExperienceHandle = {
+    pid: PortableExperienceIdentifier
```
menduz: should this pid be the sceneUrn instead? I think there is a greater concept here, will use this 👾 emoji to refer to the same topic in the review

menduz (changes requested):
This pr will work as it is if we merge it now, although, I think we are veeery close to have a very good foundation for the greater objective of URNs and to make some names more consistent. 

I left some comments regarding the opportunity to consolidate URNs in the scene ids for portable experiences with this emoji 👾 to know that all of them are talking about the same topic

kernel/packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -69,39 +69,25 @@ export async function getPortableExperience(pid: string): Promise<PortableExperi
...
   }
-  const baseUrl: string = new URL('..', mappingsUrl).toString()
-  const mappingsFetch = await fetch(`${baseUrl}mappings`)
+  const mappingsFetch = await fetch(mappingsUrl)
   const mappingsResponse = (await mappingsFetch.json()) as MappingsResponse
 
   const sceneJsonMapping = mappingsResponse.contents.find(($) => $.file === 'scene.json')
 
   if (sceneJsonMapping) {
+    const baseUrl: string = mappingsUrl.slice(0, -1 * 'mappings'.length)
```
menduz: Why did you drop the url resolution? that approach sounds hackish and assumes it always returns a url ending in "mappings"

```suggestion
    const baseUrl: string = `${new URL('..', mappingsUrl)}`
```

kernel/packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -69,39 +69,25 @@ export async function getPortableExperience(pid: string): Promise<PortableExperi
...
     const sceneResponse = await fetch(`${baseUrl}${sceneJsonMapping.hash}`)
 
     if (sceneResponse.ok) {
       const scene = (await sceneResponse.json()) as SceneJsonData
       return getLoadablePortableExperience({
         sceneUrn: sceneUrn,
-        baseUrl,
+        baseUrl: mappingsUrl,
```
menduz: Was this on purpose?

kernel/packages/unity-interface/portableExperiencesUtils.ts
agusaldasoro: It is a hack that I don't expect to last, I was just testing. Anyway, the problem I faced was that the resolver returns only: `https://static-pe.decentraland.io/${asset.id}/mappings`, so if I slice only the baseUrl, then I'll have to manually add the assetId. Any idea?

kernel/packages/unity-interface/portableExperiencesUtils.ts
agusaldasoro: my bad :)

kernel/packages/unity-interface/portableExperiencesUtils.ts
menduz: I don't understand the problem, is it related to this line?

agusaldasoro: Yes, the solution that I found using the url resolution is something like:


```
const baseUrl: string = `${new URL('..', mappingsUrl)}`
const sceneUrl: string = `${baseUrl}/${portableExperienceId}/${sceneHash}`
```

I don't really love it, what do you think?

menduz: Oh I understand, I think it is easier than it looks since the baseUrl already contains the portableExperienceId

```ts
// this resolves .. from the /mappings url
const baseUrl: string = new URL('..', mappingsUrl).toString()

// this resolves the scene.json
const sceneUrl: string = new URL(sceneHash, baseUrl).toString()
// this also works, and it sounds better from a security point of view
const sceneUrl: string = `${baseUrl}/${sceneHash}`

```

agusaldasoro: No, it doesn't :( the value for `sceneUrl` after running that is `https://static-pe.decentraland.io/scene.json`

menduz: Omg, something is wrong, thanks for clarifying. I will check the library

menduz: I checked, it was my mistake, I accidentally added two dots instead of one in the URL('..')

This one should work

```ts
// this resolves .. from the /mappings url
const baseUrl: string = new URL('.', mappingsUrl).toString()

// this resolves the scene.json
const sceneUrl: string = new URL(sceneHash, baseUrl).toString()
// this also works, and it sounds better from a security point of view
const sceneUrl: string = `${baseUrl}/${sceneHash}`
```

```js
Welcome to Node.js v14.15.0.
Type ".help" for more information.
> r = require('@dcl/urn-resolver')
{
  LandUtils: [Getter],
  resolveContentUrl: [Getter],
  parseUrn: [AsyncFunction: parseUrn],
  resolveUrlFromUrn: [AsyncFunction: resolveUrlFromUrn]
}
> r.resolveUrlFromUrn('urn:decentraland:off-chain:static-portable-experiences:quest-1').then($ => console.log(new URL('aHash', new URL('.', $)).toString()))
Promise { <pending> }
> https://static-pe.decentraland.io/quest-1/aHash
```

kernel/packages/shared/apis/PortableExperiences.ts
```diff
@@ -0,0 +1,69 @@
...
+    const parcelIdentity: ParcelIdentity = this.options.getAPIInstance(ParcelIdentity)
+
+    const portableExperienceUrn: string = buildPortableExperienceUrn(parcelIdentity)
+    await killPortableExperienceScene(portableExperienceUrn)
+    return true
+  }
+}
+
+function buildPortableExperienceUrn(parcelIdentity: ParcelIdentity): string {
+  return `urn:decentraland:off-chain:static-portable-experiences:${parcelIdentity.cid}`
```
menduz: `parcelIdentity.cid` for portable experiences _should be_ the whole URN

kernel/packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -0,0 +1,138 @@
...
+export async function spawnPortableExperienceScene(
+  sceneUrn: string,
+  parentCid: string
+): Promise<PortableExperienceHandle> {
+  const scene = new UnityPortableExperienceScene(await getPortableExperienceFromS3Bucket(sceneUrn))
+
+  loadParcelScene(scene, undefined, true)
+  const parcelSceneId = getParcelSceneID(scene)
+  unityInterface.CreateUIScene({
+    id: parcelSceneId,
```
menduz: `id` should be the URN

kernel/packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -0,0 +1,138 @@
...
+  parentCid: string
+}
+
+let currentPortableExperiences: Map<string, string> = new Map()
+
+export async function spawnPortableExperienceScene(
+  sceneUrn: string,
+  parentCid: string
+): Promise<PortableExperienceHandle> {
+  const scene = new UnityPortableExperienceScene(await getPortableExperienceFromS3Bucket(sceneUrn))
```
menduz, line 34: what would happen if I execute this call twice?

```ts
spawnPortableExperienceScene(urn, parentUrn)
spawnPortableExperienceScene(urn, parentUrn)
```

Will it create two copies of the same experience?

kernel/packages/unity-interface/portableExperiencesUtils.ts
agusaldasoro: Ohh, thanks!

kernel/packages/unity-interface/portableExperiencesUtils.ts
agusaldasoro: No, it creates only one per id.

menduz (approved):

PR #1972, opened 2021-01-29
refactor: make wearables & inventory work on demand

nchamo:
Since wearables will be opened to the community, we can't continue with the current "pre-load everything" approach. So what we are doing here is changing how all this works. This will get us a step closer to stop using the legacy wearable-api. We can divide the changes in two:

**Making wearables on-demand**
* Now the catalog won't be sent to the renderer by default
* The renderer will ask Kernel for wearable descriptions as they are needed
* Kernel will then (and only then) fetch that wearable description and send it to the renderer

**Making inventory on-demand**
* Profiles won't have a list of owned wearables (or inventory) by default
* Unity will ask Kernel which are the owned wearables
* Kernel will then (and only then) fetch this data and sent it to the renderer

**Protocol**
In order to make this work, we defined a little "protocol" for requesting wearables. 
* Unity will request wearables from Kernel using different filters, and it will provide a `context` (an arbitrary string)
* Kernel will then apply the filters, collect the wearable definitions and send them to Unity, with the same `context`


kernel/packages/shared/catalogs/sagas.ts
```diff
@@ -126,21 +121,88 @@ function* initialLoad() {
...
+
+  defaultLogger.error(`Failed to fetch wearables for context '${context}'`, error)
+
+  yield call(ensureRenderer)
+  yield call(informRequestFailure, error, context)
+}
+
+function areFiltersValid(filters: WearablesRequestFilters) {
+  let filtersSet = 0
+  let ok = true
```
menduz, line 189: best variable name ever

menduz (approved):
It looks good from my side. You must take into account that I am not familiar with the sagas code, I love the amount of "de-hardcoding" that is present in this PR.

marcosnc (approved):
Great work!

PR #1974, opened 2021-02-01
feat: add builder-projects-panel scene card view

pbosio:
+ add SceneCardView

screenshot for reference:
Screen Shot 2021-02-01 at 12 30 51

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Scripts/ScenesController/SceneData.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Scripts/ScenesController/SceneData.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using UnityEngine;
+
+internal interface ISceneData
+{
+    Vector2Int coords { get; }
+    Vector2Int size { get; }
+    string id { get; }
+    string name { get; }
+    string thumbnailUrl { get; }
+    bool isOwner { get; }
+    bool isOperator { get; }
+    bool isContributor { get; }
+    bool isDeployed { get; }
+}
+
+internal class SceneData : ISceneData
+{
+    public Vector2Int coords;
+    public Vector2Int size;
+    public string id;
+    public string name;
+    public string thumbnailUrl;
+    public bool isOwner;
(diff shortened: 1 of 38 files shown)


pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Scripts/ScenesController/SceneData.cs
```diff
@@ -0,0 +1,38 @@
...
+    Vector2Int ISceneData.coords => coords;
+    Vector2Int ISceneData.size => size;
+    string ISceneData.id => id;
+    string ISceneData.name => name;
+    string ISceneData.thumbnailUrl => thumbnailUrl;
+    bool ISceneData.isOwner => isOwner;
+    bool ISceneData.isOperator => isOperator;
+    bool ISceneData.isContributor => isContributor;
+    bool ISceneData.isDeployed => isDeployed;
+}
```
BrianAmadori, line 38: I'm thinking that we may not need to extract an interface of this class, because data classes are generally used as stubs as opposed to mocks. 

To put into different terms: generally you are going to use the same class with mocked contents instead of extracting a mocked class that doesn't need any particular implementation.

BrianAmadori (approved):
lgtm!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Scripts/ScenesController/SceneData.cs
pbosio: I'm using this way mostly to handle access restrictions to avoid changing some property by mistake. Every class that is expecting this data is using the interface that only has getters.
Plus, since we don't have fully defined the structure of the data I prefer using the interface since the actual class might change

pbosio: In any case, I think the future will tell us if we need a refactor to remove that interface, I would like to leave that decision for the future Pato :P

PR #1976, opened 2021-02-01
feat: add builder-projects-panel menu sections controller

pbosio:
+ add controller for menu sections (to handle which menu screen should be open/close)

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Tests/SectionsControllerShould.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Tests/SectionsControllerShould.cs
@@ -1,0 +1,136 @@
+﻿using NUnit.Framework;
+using UnityEngine;
+
+namespace Tests
+{
+    public class SectionsControllerShould
+    {
+        private SectionsController controller;
+        private SectionFactory sectionFactory;
+
+        [SetUp]
+        public void SetUp()
+        {
+            sectionFactory = new SectionFactory();
+            controller = new SectionsController(sectionFactory, null);
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            controller.Dispose();
+        }
+
+        [Test]
+        public void OpenSection()
+        {
+            bool openCallbackCalled = false;
+            SectionBase sectionOpened = null;
+
+            void OnSectionOpen(SectionBase section)
+            {
+                openCallbackCalled = true;
+                sectionOpened = section;
+            }
+
+            controller.OnSectionShow += OnSectionOpen;
+            controller.OpenSection(SectionsController.SectionId.SCENES_MAIN);
+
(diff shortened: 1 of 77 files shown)


pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Tests/SectionsControllerShould.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            Assert.IsFalse(sectionFactory.sectionScenesMain.isVisible);
+            Assert.IsTrue(sectionFactory.sectionScenesProjects.isVisible);
+
+            Assert.AreEqual(sectionFactory.sectionScenesProjects, openSection);
+            Assert.AreEqual(sectionFactory.sectionScenesMain, hiddenSection);
+        }
+    }
+
+    class SectionFactory : ISectionFactory
```
BrianAmadori: This factory can't be mocked using `NSubstitute`? In any case we could rename it to have the `Mock` moniker to avoid confusions here.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Tests/SectionsControllerShould.cs
```diff
@@ -0,0 +1,136 @@
...
+    {
+        public SectionScenesMain sectionScenesMain;
+        public SectionScenesProjects sectionScenesProjects;
+
+        public SectionFactory()
+        {
+            sectionScenesMain = new SectionScenesMain();
+            sectionScenesProjects = new SectionScenesProjects();
+        }
```
BrianAmadori: Same comment about mocking it through `NSubstitute`.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Tests/SectionsControllerShould.cs
pbosio: I don't think that I can use NSubstitute here cause I'm adding two fields to the mocked factory to use for reference comparison on the test. I will look into it, otherwise I will change the name of the class as suggested

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Tests/SectionsControllerShould.cs
pbosio: I ended up renaming the classes. I think using NSubstitute is possible, but while trying to do so the code was becoming less easy to understand so I decide to keep the mock classes

AjimenezDCL (approved):
LGTM!

PR #1987, opened 2021-02-05
fix: assign currentShape before OnLoadCompleted

D4rWiNSS:
# What? 
This PR fixes the issue with SceneBoundarieChecker not being able to make invisible some items outside the parcel


BrianAmadori (approved):
nice fix!

PR #2067, opened 2021-02-09
chore: remove feature flags for voice chat

nchamo:
The voice chat feature is live, and there is no going back. So we are removing all feature flag and config related to it

### DIFF
--- a/kernel/packages/config/index.ts
+++ b/kernel/packages/config/index.ts
@@ -100,10 +100,6 @@ export const RESIZE_SERVICE = qs.RESIZE_SERVICE
 export const HOTSCENES_SERVICE = qs.HOTSCENES_SERVICE
 export const REALM = qs.realm
 
-export const VOICE_CHAT_DISABLED_FLAG = location.search.includes('VOICE_CHAT_DISABLED')
-
-export const VOICE_CHAT_ENABLED_FLAG = location.search.includes('VOICE_CHAT_ENABLED')
-
 export const ENABLE_BUILDER_IN_WORLD = location.search.includes('ENABLE_BUILDER_IN_WORLD')
 
 export const AUTO_CHANGE_REALM = location.search.includes('AUTO_CHANGE_REALM')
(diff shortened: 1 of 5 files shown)


kernel/packages/config/index.ts
```diff
@@ -100,10 +100,6 @@ export const RESIZE_SERVICE = qs.RESIZE_SERVICE
 export const HOTSCENES_SERVICE = qs.HOTSCENES_SERVICE
 export const REALM = qs.realm
 
-export const VOICE_CHAT_DISABLED_FLAG = location.search.includes('VOICE_CHAT_DISABLED')
```
pablitar, line 103: I'd leave this URL flag. It doesn't hurt anyone.

kernel/packages/shared/meta/selectors.ts
```diff
@@ -33,16 +33,5 @@ export const isMOTDInitialized = (store: RootMetaState): boolean =>
...
   store.meta.config.world ? store.meta.config.world.messageOfTheDay || null : null
 
-export const isVoiceChatEnabled = (store: RootMetaState): boolean => {
-  if (!WORLD_EXPLORER || VOICE_CHAT_DISABLED_FLAG) return false
-  return !!getCommsConfig(store).voiceChatEnabled || VOICE_CHAT_ENABLED_FLAG
-}
-
-export const getVoiceChatAllowlist = (store: RootMetaState): string[] => getCommsConfig(store).voiceChatAllowlist ?? []
-
-export const isVoiceChatEnabledFor = (store: RootMetaState, userId: string): boolean =>
```
pablitar, line 43: I'd also leave this method. If for any reason we need to have an enable/disable functionality for voice chat in the future, we can use this as hook.

nchamo: Done!

kernel/packages/config/index.ts
nchamo: Done!

pablitar (approved):
LGTM!

PR #2071, opened 2021-02-09
feat: adding support for feature flags

nchamo:
We are now adding support for feature flags inside the explorer. 

We also added a selector called `isFeatureEnabled(featureName: string, ifNotSet: boolean)` that accepts url overrides. To explain how it works, let's go with the feature flag called `wearables_v2`. 
1. If the url has the following on the query `DISABLE_WEARABLES_V2`, then the flag is considered as off, regardless of what the server says
1. If the url has the following on the query `ENABLE_WEARABLES_V2`, then the flag is considered as on, regardless of what the server says
1. If the flag is configured on the server, then that is the value that stands
1. If it wasn't present on the server, then `ifNotSet` is used as default

This will allow us to easily add new flags with little to no effort


pablitar (approved):
LGTM!

menduz (approved):

PR #2072, opened 2021-02-10
fix: add avatar culling behaviour

BrianAmadori:
This PR adds avatar shadow culling behaviour to `CullingController`. This will improve performance for events and crowded locations.

- Avatars that are far away will have its shadow turned off.
- The threshold distance is coupled to the `Detail Object Culling` slider in quality settings (15m for slider in 50%).


pravusjif (approved):
The code looks neat and tested it working with live users in genesis plaza with no problems

sandrade-dcl (approved):

PR #2076, opened 2021-02-10
feat: ab converter visual test

BrianAmadori:
# Why? 
- We need a more consistent way of detecting an Asset Bundle conversion failure as many visual defects have been reported on converted assets that get uploaded to the content server.
- We need to avoid uploading the visual imperfect ABs and consider them as failed

# What? 
- Implement visual tests to determine if an asset has been successfully converted
- Add visual tests check to the conversion pipeline
- Avoid uploading failed ABs


unity-client/Assets/ABConverter/VisualTests.cs
```diff
@@ -0,0 +1,220 @@
...
+            var pathList = Directory.GetDirectories(Application.dataPath + "/" + workingFolderName);
+
+            List<string> dependencyAbs = new List<string>();
+            List<string> mainAbs = new List<string>();
+
+            foreach (var paths in pathList)
+            {
+                var hash = new DirectoryInfo(paths).Name;
+                var path = "Assets/" + workingFolderName + "/" + hash;
+                // var guids = AssetDatabase.FindAssets("t:GameObject", new[] {path});
```
BrianAmadori:
```suggestion
```

unity-client/Assets/ABConverter/VisualTests.cs
```diff
@@ -0,0 +1,220 @@
...
+
+            foreach (GameObject go in gltfs)
+            {
+                go.SetActive(false);
+            }
+
+            foreach (GameObject go in gltfs)
+            {
+                go.SetActive(true);
+                yield return VisualTestHelpers.TakeSnapshot($"ABConverter_{go.name}.png", Camera.main, new Vector3(7, 7, 7), Vector3.zero);
```
BrianAmadori: Always positioning the camera at `7,7,7` wouldn't cause issues with big models? Maybe we'd need to position the camera in a dynamic way?

unity-client/Assets/ABConverter/VisualTests.cs
```diff
@@ -0,0 +1,220 @@
...
+                    string filePath = abPath + go.name;
+                    if (env.file.Exists(filePath))
+                    {
+                        env.file.Delete(filePath);
+
+                        string depMapPath = filePath + ".depmap";
+                        env.file.Delete(depMapPath);
+                    }
+
+                    if (core != null)
```
BrianAmadori: The need to access the whole `Core` object just for error passing is a bit overkill. I think we can avoid this pretty easily.

I propose the following:
- Make `TestConvertedAssets` have its own `System.Action<int>` callback that pass along the number of skipped assets.
- From `Core` we can do something like:

```
EditorCoroutineUtility.StartCoroutineOwnerless(
        VisualTests.TestConvertedAssets(
            env: env, 
            OnFinish: (skippedAssets) => 
                {
                    this.skippedAssets = skippedAssets;
                    OnFinish?.Invoke(state.lastErrorCode);
                })
    );
```

BrianAmadori: Well... Just figured this looks like I'm leaving review comments in my own code.

unity-client/Assets/ABConverter/Core.cs
```diff
@@ -83,9 +83,14 @@ public Core(Environment env, Client.Settings settings = null)
...
             {
                 if (OnFinish == null)
+                {
                     OnFinish = CleanAndExit;
+                }
                 else
+                {
+                    OnFinish -= CleanAndExit;
                     OnFinish += CleanAndExit;
+                }
```
D4rWiNSS: This doesn't work if we just do 

` OnFinish += CleanAndExit;`?

unity-client/Assets/ABConverter/VisualTests.cs
```diff
@@ -0,0 +1,220 @@
...
+                var req = UnityWebRequestAssetBundle.GetAssetBundle(path);
+
+                if(SystemInfo.operatingSystemFamily == OperatingSystemFamily.MacOSX)
+                    req.url = req.url.Replace("http://localhost", "file:///");
+
+                req.SendWebRequest();
+
+                while (!req.isDone)
+                {
+                }
```
D4rWiNSS, line 167: Can this be move to a Courutine?

unity-client/Assets/ABConverter/VisualTests.cs
```diff
@@ -0,0 +1,220 @@
...
```
D4rWiNSS, line 194: Same as above, we could move the whole call to a coroutine if possible

D4rWiNSS (approved):
I don't know much about this matter but it looks good to me, I left two comments that maybe can't be addresed

unity-client/Assets/ABConverter/VisualTests.cs
pravusjif: I updated the code to take into account the renderers merged bounds and center, to calculate the camera offset position. I think the snapshots look much better now (the small objects were super small in the snapshot before, not anymore)

unity-client/Assets/ABConverter/VisualTests.cs
pravusjif: Comment removed

unity-client/Assets/ABConverter/VisualTests.cs
pravusjif: Great suggestion, it bothered me as well that I needed the Core object to mark the "skipped assets".

Just implemented the change.

unity-client/Assets/ABConverter/Core.cs
pravusjif: You are right! I just tested it just in case, but this works fine with just that one line.

unity-client/Assets/ABConverter/VisualTests.cs
pravusjif: The problem with that is that the `LoadAndInstantiateAllAssetBundles()` needs to return the collection of instantiated gameobjects (`GameObject[]`), to make the change you suggest we should change that method to be a coroutine, thus returning an `IEnumerator`, and have the array pass as a reference parameter or something similar.
Since all the Asset Bundle Conversor code runs in editor more (in a service we have running and converting assets) and not in production, I wouldn't make this change you suggest, otherwise i think you are right and we should turn these `while(true)` waits into coroutines to avoid stopping the main thread.

BrianAmadori (commented):
Can't approve because I opened this PR, but approved in spirit. 🔥 

Also, +666 lines is a good omen.

D4rWiNSS (approved):

PR #2077, opened 2021-02-11
refactor: instanced data store

AjimenezDCL:
The static approach for the `DataStore` makes mocking and testing difficult. The static instance is not being cleaned-up between tests and this lead to errors, leaks from variables not unsubscribing to events, etc...

Ideally we would want this `DataStore` to be in the `Environment` class, and certainly we are aiming for that, but currently is not doable. `Environment` is still in the `MainScripts` assembly and putting the `DataStore` on it would force every class that uses it to reference it. This goes against our current goal of decoupling that humongous assembly from the rest and leads to annoying circular references.

So until `Environment` is on its own assembly we are going to have the `DataStore` as an auto-instanced singleton with a `Clear()` method. I know we are trying to avoid them, but it's temporary and it's the only solution I came up with to keep the system decoupled.

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
@@ -180,7 +180,7 @@ public class InputController : MonoBehaviour
                     InputProcessor.FromKey(action, KeyCode.U, modifiers: InputProcessor.Modifier.None);
                     break;
                 case DCLAction_Trigger.CloseWindow:
-                    if (allUIHidden || DataStore.isSignUpFlow.Get()) break;
+                    if (allUIHidden || DataStore.i.isSignUpFlow.Get()) break;
                     InputProcessor.FromKey(action, KeyCode.Escape, modifiers: InputProcessor.Modifier.None);
                     break;
                 case DCLAction_Trigger.OpenExpressions:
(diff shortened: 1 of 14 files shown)


sandrade-dcl (approved):
LGTM as a temporal solution 👍

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
```diff
@@ -180,7 +180,7 @@ public void Update_Trigger(InputAction_Trigger[] triggerTimeActions)
                     InputProcessor.FromKey(action, KeyCode.U, modifiers: InputProcessor.Modifier.None);
                     break;
                 case DCLAction_Trigger.CloseWindow:
-                    if (allUIHidden || DataStore.isSignUpFlow.Get()) break;
+                    if (allUIHidden || DataStore.i.isSignUpFlow.Get()) break;
```
D4rWiNSS: if (allUIHidden || DataStore.i.isSignUpFlow.Get())
                      break;

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/MouseCatcher/MouseCatcher.cs
```diff
@@ -41,7 +41,7 @@ void Update()
 
         public void LockCursor()
         {
-            if (!renderingEnabled || DataStore.isSignUpFlow.Get()) return;
+            if (!renderingEnabled || DataStore.i.isSignUpFlow.Get()) return;
```
D4rWiNSS: if (!renderingEnabled || DataStore.i.isSignUpFlow.Get())
     return;

D4rWiNSS (approved):
LGTM!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
AjimenezDCL: I dont think we have a convention against one-liners but will change it anyway! Thx!

PR #2078, opened 2021-02-11
feat: builder in world smart item implementation and refactor

### LINKED ISSUE #1998: Smart items support on stateful scenes
We need to start supporting smart items in stateful scenes. This will allow us to:
- Communicate changes in smart items between Kernel & Unity
- Save smart items in the content server

### LINKED ISSUE #2039: Refactor SceneObject Catalog to have his custom data structure to have better consistency (Instead of work with sceneObjects it should work with CatalogObject that have extra data)

### LINKED ISSUE #2041: Entity should have a relation somehow to SceneObject since their data are related

### LINKED ISSUE #2043: Add SetModel to base component classes

D4rWiNSS:
closes #2043 
closes #2041 
closes #2039 
closes #1998

# What? 

This PR includes the functionality of the smart Items implementation, the catalog refactor, and some bug fixes.

This PR also have some changes in the Builder in World Tests, so they have more consistency


unity-client/Assets/Scripts/MainScripts/DCL/Components/SmartItem/SmartItemComponent.cs
```diff
@@ -19,29 +19,17 @@ public override object GetModel()
...
-                return true;
-
-            return false;
-        }
-
         public override IEnumerator ApplyChanges(string newJson)
         {
-            UpdateModel(newJson);
-            yield return null;
+            Model newModel = Utils.SafeFromJson<Model>(newJson);
```
BrianAmadori: `SafeFromJson` doesn't work with dictionaries, nor classes without `Serializable` attribute.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
```diff
@@ -32,6 +32,7 @@ public enum EditModeState
 
     [Header("Activation of Feature")]
     public bool activeFeature = false;
+    public bool byPassLandOwnershipCheck = false;
```
BrianAmadori: `bypass` is a single word.

```suggestion
    public bool bypassLandOwnershipCheck = false;
```

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
```diff
@@ -55,7 +55,7 @@ public void ForceSetNewName(string value)
     public override IEnumerator ApplyChanges(string newJson)
     {
         ApplyNewValue(newJson);
-        return null;
+        yield break;
```
AjimenezDCL, line 58: Aren't these equivalents? Not requesting a change, just curious

unity-client/Assets/Scripts/MainScripts/DCL/Components/SmartItem/SmartItemCommon/SmartItemCommon.cs
```diff
@@ -30,5 +54,12 @@ public class OptionsParameter
             public string label;
             public string value;
         }
+
+        public ParameterType GetParameterType()
+        {
+            if (!Enum.TryParse(type.ToUpper(), out ParameterType myStatus))
+                Debug.Log($"Error parsing the smart item parameter type: {type}, The parameter doesn't exist!");
+            return myStatus;
```
AjimenezDCL: Maybe we could cache this to avoid parsing it on each access.

You could even try to make the `type` a property with a setter that automatically parses it into a `ParameterType` field, I'm not sure if it will work though, Unity Serialization system is special.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
AjimenezDCL:
```suggestion
    public bool bypassLandOwnershipCheck = false;
```

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/BIWCatalogManager.cs
```diff
@@ -0,0 +1,205 @@
...
+
+    public static void Init()
+    {
+        if (i == null)
+        {
+            i = new BIWCatalogManager();
+
+            BuilderInWorldNFTController.i.OnNftsFetched += ConvertCollectiblesPack;
+            AssetCatalogBridge.OnSceneObjectAdded += AddSceneObject;
+            AssetCatalogBridge.OnSceneAssetPackAdded += AddSceneAssetPack;
```
AjimenezDCL, line 22: Besides these listeners to events, everything here is static, maybe we can avoid the singleton, make this class static and renaming it to `BIWCatalogHelpers`.

If I'm not mistaken `ConvertCollectiblesPack`, `AddSceneObject` and `AddSceneAssetPack` are just putting info in the `DataStore`. This could be done in the `BuilderInWorldNFTController` and the `AssetCatalogBridge` when raising these events or we could mainting these methods here as wrappers if we dont want to overload the other classes.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/CatalogData/Model/CatalogItem.cs
```diff
@@ -0,0 +1,74 @@
...
+    public SmartItemAction[] actions;
+    public bool isVoxel = false;
+    public string thumbnailURL;
+
+    private bool isFavorite = false;
+
+    public string GetThumbnailUrl()
+    {
+        return thumbnailURL;
+    }
```
AjimenezDCL: Our code convention allowes the use of expresion bodies:
```
    public string GetThumbnailUrl() => thumbnailURL;
```

Not requesting a change, I prefer them because they reduce the amount of code but it's completly subjetive.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/EntityInformation/SmartItems/Parameters/Actions/SmartItemActionEventAdapter.cs
```diff
@@ -11,17 +11,54 @@ public class SmartItemActionEventAdapter : MonoBehaviour
...
+    {
+        switch (index)
+        {
+            case 0:
+                ResetActionable();
+                break;
+            case 1:
+                RemoveActionable();
+                break;
+        }
```
AjimenezDCL, line 43: Not a huge fan of this since a change in the order of the dropdown would require code changes. Buuuut, I havent worked with the `TMP_Dropdown` and maybe there's no alternative.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/EntityInformation/SmartItems/SmartItemUIParameterAdapter.cs
```diff
@@ -12,21 +12,50 @@ public class SmartItemUIParameterAdapter : MonoBehaviour
...
+    }
+
+    protected virtual object GetParameterValue()
+    {
+        if (currentValues.ContainsKey(KEY_NAME))
+            return currentValues[KEY_NAME];
+        else if (!string.IsNullOrEmpty(currentParameter.@default))
+            return currentParameter.@default;
+        else
+            return null;
```
AjimenezDCL:
```suggestion
        if (currentValues.ContainsKey(KEY_NAME))
            return currentValues[KEY_NAME];

	if (!string.IsNullOrEmpty(currentParameter.@default))
            return currentParameter.@default;

        return null;
```

AjimenezDCL (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
D4rWiNSS: I am not totally certain about this ( you can take it with a grain of salt), but I think the difference is that return null will wait for a frame to complete the IEnumerator while yield break will exit inmediatelly

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/EntityInformation/SmartItems/Parameters/Actions/SmartItemActionEventAdapter.cs
D4rWiNSS: Since this Dropdown is not generated by code the only thing you can do is to add an enum but you will need to change it anyway if you change the order. 

This approach is not "Perfect" since we could create custom actions and it will be better, however, since this won't change (most likely) I think we should be good

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
AjimenezDCL: I dont think so since you are not yielding. I dont think `yield return null` == `return null`. 

Anyway, we can leave it, it's completly fine.

pravusjif:
Tested using a smart item in my land with no problem. The "publishing" UI never goes away when you publish a scene but that's an issue for another PR.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioStream.cs
```diff
@@ -147,5 +143,20 @@ public void UpdateOutOfBoundariesState(bool enable)
...
+            ApplyCurrentModel(prevModel);
+        }
+
+        private void ApplyCurrentModel(Model prevModel)
+        {
+            bool forceUpdate = prevModel.volume != model.volume;
+            settingsVolume = Settings.i.generalSettings.sfxVolume;
+
+            UpdatePlayingState(forceUpdate);
+        }
```
pravusjif: I find it a bit confusing that we feed the `prevModel` to the `ApplyCurrentModel()` and also having the parameter called `prevModel` inside the `ApplyCurrentModel()` implementation.
Shouldn't we just apply the current model as we do in the other classes `SetModel()` ?

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
```diff
@@ -189,5 +188,12 @@ public override void Cleanup()
             avatarUserInfo.userId = model.id;
             MinimapMetadataController.i?.UpdateMinimapUserInformation(avatarUserInfo, true);
         }
+
+        public override void SetModel(object model)
+        {
+            this.model = (AvatarModel)model;
+
+            //TODO (Adrian): This should handle the change of the model. This will required a major refactor so if you really need it, feel free to implement it!
+        }
```
pravusjif, line 197: shouldn't we call this new `SetModel()` inside this class `ApplyChanges()` instead of the old `model = Utils.SafeFromJson<AvatarModel>(newJson);` ? it would be something like `SetModel(Utils.SafeFromJson<AvatarModel>(newJson));` I think.

unity-client/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
```diff
@@ -196,5 +180,38 @@ public static TextAlignmentOptions GetAlignment(string vTextAlign, string hTextA
                     }
             }
         }
+
+        public override void SetModel(object model)
+        {
+            this.model = (Model)model;
+            PrepareRectTransform();
+            ApplyCurrentModel();
+        }
```
pravusjif, line 189: I think `SetModel()` and `ApplyChanges()` responsibilities are colliding in some cases like in this class.

Normally a component would have its `ApplyChanges()` called with a new JSON representation of the model (sent by the Kernel) and, inside that method, we parse the model and effectively apply the new changes.
If there is a new case now, with the builder in-world functionalities, in which we modify the component client-side without receiving a new JSON from kernel, then it's probably better if we move the current `ApplyChanges()` logic from all components to their new `SetModel()` method. Leaving only the model JSON parsing and a call to its own `SetModel()` inside the `ApplyChanges()` of every component.

What do you think?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/DecentralandEntity.cs
```diff
@@ -180,14 +181,27 @@ public void RemoveSharedComponent(System.Type targetType, bool triggerDettaching
...
+            T component = components.Values.FirstOrDefault(x => x is T) as T;
+            if (component != null)
+                return component;
+
+            component = sharedComponents.Values.FirstOrDefault(x => x is T) as T;
+            if (component != null)
+                return component;
+
+            return null;
+        }
```
pravusjif: I see that `TryGetSharedComponent()` now uses this new method.
Besides the fact that LinQ may affect performance if called too frequently (not sure how frequent are calls to `TryGetSharedComponent()`), if this new `TryGetComponent()` method is normally used more for shared components instead of normal components, can we please check the shared components collection first instead of normal components? to avoid that extra LinQ check on most cases.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioStream.cs
D4rWiNSS: I have changed it to be more clear! The original thought to make the call `SetModel `was to have a way to set the model from unity without the need to always wait for a frame in the `ApplyChanges`. This way we can work with updated data if we create the component from Unity and set its data. 

Since `ApplyChanges ` required a JSON It doesn't have much sense to encode the data in Unity to work with Unity.  I was just searching for an easy way to make it work in all components without a major refactor in the code that supports the JSON and data.

If this doesn't work for you, you have a better idea or you think that we need a better naming, feel free to share it with me so we can take a look!

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
D4rWiNSS: Same as before, the `SetModel` call is designed to update the model component in the same frame so call to ApplyChanges won't work for that

unity-client/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
D4rWiNSS: That sounds good to me but there is a big problem with that approach that we need to solve.

In builder in world, we create/update components client-side and send an update to the kernel that the component has changed. Some components will need some frames to apply the model that is the reason `ApplyChanges  ` returning an `IEnumerator ` and update its data. At the moment of creation of an entity, we tell kernel something like "New entity with X,Y,Z components and their values". In this situation we need components X,Y,Z to have their data updated in order to sent the correct data to Kernel.

However, since the components may need some frames to apply the changes, we could end in a situation where we are updating the components with old values. 

The correct way to handle this should be to separate the current data from the changes applies but since this is a big refactor and we don't know if this can affect heavily in the current system, we ended with this approach.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/DecentralandEntity.cs
D4rWiNSS: `TryGetSharedComponent()` still have his current implementation so this shouldn't affect it, This method is used for components that implement interfaces. Right now it is called frequently, however, I will add some notes and information to make its use more clear!

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
BrianAmadori: Just FYI: returning just `null` causes the `IEnumerator` to be executed like a normal method instead a coroutine. `yield break` effectively causes a frame skip and we should avoid it if we don't need it.

It's a bit of a hack: https://github.com/decentraland/explorer/blob/43770b126b0f802431cf94534310f56a977d46de/unity-client/Assets/Scripts/MainScripts/DCL/Components/ComponentUpdateHandler.cs#L105

pravusjif (approved):

AjimenezDCL (approved):

Issue #2068, opened 2021-02-09
In preview using next, objects wrongly appear out of bounds


nearnshaw:
If you run genesis plaza as a local preview using @latest it looks fine
But if you install @next and then run it again, about a 1/3 of the plaza is marked in red as being out of bounds

![Screen Shot 2021-02-09 at 13.49.18.png](https://images.zenhubusercontent.com/5d9b940e491c060001c8647b/ae168d3c-8217-43e9-b925-16282a0c561e)

This is the repo for genesis plaza to test:

https://github.com/decentraland-scenes/Genesis-Plaza

PR #2080, opened 2021-02-11
fix: In preview using next, objects wrongly appear out of bounds

sandrade-dcl:
Fix related to this [BUG](https://github.com/decentraland/explorer/issues/2068).


pravusjif (approved):
tested with local scenes in preview mode and the bug is fixed

PR #2084, opened 2021-02-12
fix: support estates with more than 100 parcels

nchamo:
TheGraph has a 100 default limit for queries. We are increasing that limit up to the max (1000)


D4rWiNSS (approved):
Nice catch! LGTM!

PR #2087, opened 2021-02-12
fix: ensure disable culling for avatar preview

pbosio:


sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/CharacterPreviewController.cs
```diff
@@ -191,4 +201,13 @@ public void Rotate(float rotationVelocity)
     {
         avatarRenderer.transform.Rotate(Time.deltaTime * rotationVelocity * Vector3.up);
     }
+
+    private void EnforceDisableAvatarCulling()
```
BrianAmadori, line 205: We can't just stop the `CullingController` before taking the snapshot and start it again afterwards? When stopping it, it ensures all the renderers go back to the initial state. I think `SetCullingForRenderer` shouldn't be exposed from outside like this as its breaks encapsulation.

BrianAmadori (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/CharacterPreviewController.cs
pbosio: I think that currently stopping CullingController doesn't reset the renderers to the initial state.
I can add that, but there are two possible scenarios there:
1- resetting the all renderers without using a coroutine means that we might have a hiccup there when taking snapshot
2- resetting renderers inside a coroutine means that clicking "save" in avatar editor will take longer to process and close the editor (which I think that already takes too long)

pravusjif (approved):

pravusjif:
Tested with a guest user and saw no problem when generating the local snapshots

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/CharacterPreviewController.cs
BrianAmadori: Resetting all the renderers will involve a pretty negligible time compared to taking the snapshot and sending it over to kernel, so I think we can do it safely.

pbosio:
I'm changing the approach on another PR

PR #2093, opened 2021-02-17
fix: stop culling when taking avatar snapshot in avatar editor

pbosio:


unity-client/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -330,17 +331,17 @@ internal void ResetObjects()
             var renderers = objectsTracker.GetRenderers();
             var animations = objectsTracker.GetAnimations();
 
-            for (var i = 0; i < skinnedRenderers.Length; i++)
+            for (var i = 0; i < skinnedRenderers?.Length; i++)
```
sandrade-dcl, line 334: Only for curiossity: why did you need to add the `?` operator? Were we receiving sometimes the `skinnedRenderers` as null and it maked the function flow fail?

sandrade-dcl (approved):
LGTM!

unity-client/Assets/Rendering/Culling/CullingController.cs
pbosio: I don't think those array can be null in runtime, but there were throwing an annoying error when stopping unity editor :)

AjimenezDCL (approved):

pravusjif (approved):
Tested snapshots with a guest user with no problems.
Also saw no problems dealing with other rendered avatars around me.

PR #2095, opened 2021-02-18
fix: better wearables error handling and disable avatar snapshot on error

pbosio:
- fix situation where wearables were not correctly loaded and success event was triggered anyway
- handle avatar load error to avoid taking a snapshot of the avatar if it's not loaded properly


pravusjif (approved):
Tested snapshots with a guest user with no problems.
Also saw no problems dealing with other rendered avatars around me.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/CharacterPreviewController.cs
```diff
@@ -45,6 +45,8 @@ public enum CameraFocus
 
     private Coroutine updateModelRoutine;
 
+    private bool avatarLoadFailed = false;
+
     private void Awake()
     {
         cameraFocusLookUp = new System.Collections.Generic.Dictionary<CameraFocus, Transform>()
```
BrianAmadori: Maybe we can do some `using` here to avoid verbosity?

sandrade-dcl (approved):

BrianAmadori (approved):
I'm a bit wary that we don't have those cases of failure handling covered by tests, but the fixes LGTM.

PR #2097, opened 2021-02-19
fix: scene integrity checker fail

BrianAmadori:
## What?

* Scene integrity checker was not taking into account scene `Component` objects for disabled gameObjects. This was causing some false positives and false negatives in editor test runs.

* After this fix was applied, fixed some tests that correctly triggered the consistency fail.

* Scene integrity checker was encapsulated and moved to a single file in order to introduce it to the new `IntegrationTestSuite` later. 

* Tests were failing in the `WSSController` suite. Fixed it by updating the parent class to `IntegrationTestSuite`


unity-client/Assets/Scripts/Tests/IntegrationTestSuite_Legacy.cs
```diff
@@ -49,6 +51,14 @@ protected virtual IEnumerator SetUp()
 
         runtimeGameObjectsRoot = new GameObject("_RuntimeGameObjectsRoot");
 
+        testSceneIntegrityChecker = new TestSceneIntegrityChecker();
+
+        //NOTE(Brian): integrity checker is disabled in batch mode to make it run faster in CI
+        if (Application.isBatchMode || !enableSceneIntegrityChecker)
+        {
+            testSceneIntegrityChecker.enabled = false;
+        }
```
pravusjif, line 60: If we don't use the integrity checker in the CI when do we use it? XD

pravusjif (approved):

unity-client/Assets/Scripts/Tests/IntegrationTestSuite_Legacy.cs
BrianAmadori: I suppose we still use local test runs 😥

BrianAmadori: I'm thinking that maybe we can try to enable it in CI right now, as we managed to reduce the test run time.

pbosio (approved):

sandrade-dcl (approved):

D4rWiNSS (approved):

PR #2104, opened 2021-02-22
refactor: component new structure

### LINKED ISSUE #2042: Add SetModel usage to name and lock components

### LINKED ISSUE #2094: Components model have to be able to be updated in a single frame. Model should be updated instantly, visual change can be deferred.
```
        public void UpdateFromJSON(string json)
        {
            var model = GetModelFromJSON(json);
            UpdateFromObject(model);
        }

        public void UpdateFromObject(object model)
        {
            SetModel(model);
            updateHandler.ApplyChangesIfModified(null);
        }
```

D4rWiNSS:
# What?
The model has been separated from the applied changes, now the data will exists since the first frame and it can then apply the changes in futures frames without problems.

This way the model has always the correct data even if the changes are applied later. 

Also, the `OnEnable` `ApplyChanges `call has been removed

# Why? 
Since now unity can create components and send them to the kernel, we needed to have the data from the beginning to send the correct information to kernel even if the changes were applied later.

We designed a more robust way to handle the components.

closes #2094 
closes #2042


unity-client/Assets/Scripts/MainScripts/DCL/Components/BaseComponent.cs
```diff
@@ -12,10 +12,14 @@ public interface IComponent : ICleanable
...
         string componentName { get; }
         void UpdateFromJSON(string json);
-        IEnumerator ApplyChanges(string newJson);
+        void UpdateFromObject(BaseModel model);
+        IEnumerator ApplyChanges(BaseModel model);
         void RaiseOnAppliedChanges();
         ComponentUpdateHandler CreateUpdateHandler();
         bool IsValid();
+        BaseModel GetModel();
+        void SetModel(BaseModel model);
```
BrianAmadori: I'm just wondering if `SetModel` is useful now, as we have `UpdateFromObject` that should serve the same purpose. Maybe it's confusing to have both from an interface design perspective.

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
```diff
@@ -14,72 +14,73 @@
...
+            {
+                return false;
+            }
+
+            return value.Equals(item.value);
+        }
+
+        public override BaseModel GetDataFromJSON(string json)
+        {
+            return JsonUtility.FromJson<Model>(json);
```
BrianAmadori: We shouldn't be using `Utils.SafeFromJson` here?

unity-client/Assets/Scripts/MainScripts/DCL/Components/Gizmos/DCLGizmos.cs
```diff
@@ -14,32 +16,52 @@ public static class Gizmo
...
+                hashCode = hashCode * -1521134295 + localReference.GetHashCode();
+                return hashCode;
+            }
         }
 
-        public override IEnumerator ApplyChanges(string newJson)
+        public override IEnumerator ApplyChanges(BaseModel baseModel)
         {
-            model = Utils.SafeFromJson<Model>(newJson);
             yield return null;
```
BrianAmadori:
```suggestion
            return null;
```

BrianAmadori (commented):
Phew! Looking good!

unity-client/Assets/Scripts/MainScripts/DCL/Components/Animator/DCLAnimator.cs
```diff
@@ -33,9 +34,23 @@ public DCLAnimationState Clone()
...
+
+            public override bool Equals(object obj)
+            {
+                return obj is Model model &&
+                       EqualityComparer<DCLAnimationState[]>.Default.Equals(states, model.states);
+            }
+
+            public override int GetHashCode()
+            {
+                return 282409271 + EqualityComparer<DCLAnimationState[]>.Default.GetHashCode(states);
```
AjimenezDCL: what is this magic number?

AjimenezDCL (approved):
I'm only worried about these maigc numbers in the hash calculation, the rest looks good! I'm approving preemptively

unity-client/Assets/Scripts/MainScripts/DCL/Components/Animator/DCLAnimator.cs
D4rWiNSS: This magic number is a random constant, The function GetHashCode is used to generate a unique int identifier for Dictionaries keys (this way you only compare an int instead of the whole objects, that is the reason why dictionaries are more  efficient)

So for each class you need a random intenger to start the hash to be sure that you are comparing a correct key so you don't have problem with parent/children..etcs

AjimenezDCL: How are we preventing collisions?

pbosio: @Suduck can we create a static class with all those magic numbers for the hashing as constants?

pbosio (commented):
I'm worry about having to create a custom equals and hashing for every component's model.
It will increase the development time of creating new components and extending/modifying old ones.
It will increase the review time cause we will have to check if there are possible collisions/issues with the hashing and equals implementation.
We can't test that cause combinations for each component are infinite.
This might end in really difficult issues to debug as result of errors occurring in very specific situations where the hashing or equals were not perfectly implemented.
I find it kind of a big error prone solution.
On the other hand, I'm not entirely sure why are we checking if the model has changed since kernel already does this for every component before sending a new model json

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs
```diff
@@ -2,23 +2,64 @@
...
+                if (item == null)
+                {
+                    return false;
+                }
+
+                bool clipIdIsEqual = false;
+                if (string.IsNullOrEmpty(audioClipId) && string.IsNullOrEmpty(audioClipId))
+                    clipIdIsEqual = true;
+                else if (string.IsNullOrEmpty(audioClipId) || string.IsNullOrEmpty(audioClipId))
+                    clipIdIsEqual = false;
```
pravusjif: in these 2 conditionals you missed using `item.audioClipId`.

Either way, Can't we compare these with 1 line like `bool clipIdIsEqual = audioClipId == item.audioClipId;` ? or the line that you used below `audioClipId.Equals(item.audioClipId);`

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs
```diff
@@ -2,23 +2,64 @@
...
+                       playing.Equals(item.playing) &&
+                       volume.Equals(item.volume) &&
+                       loop.Equals(item.loop) &&
+                       pitch.Equals(item.pitch);
+            }
+
+            public override int GetHashCode()
+            {
+                int hashCode = 1859268890;
+                hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(audioClipId);
```
pravusjif: I preffer if we don't do these `GetHash()` overrides but if we end up using them, can we move the `-1521134295` to a static int in  `Assets/Scripts/MainScripts/DCL/Configuration/Configuration.cs` or similar? because I see we use it in many components

unity-client/Assets/Scripts/MainScripts/DCL/Components/Animator/DCLAnimator.cs
pravusjif: I'm not really following why we need these `GetHash()` overrides, are they automatically used when comparing even though we override the `Equals()` ?, can't we just determine that 2 models are equal if they have the same values regardless of their instance?, and we can remove all the GetHash() overrides.

pravusjif (changes requested):

pravusjif:
Testing the build I found a problem when the UI text is updated in the [Salmonomicon game](https://play.decentraland.zone/branch/refactor/component-model-update/index.html?ENV=org&position=-53%2C4), at the beginning of the game when you pick up the pages, the UI normally updates on every page picked but in this branch build it udpates after 2 pickups, weird

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs
D4rWiNSS: `audioClipId.Equals(item.audioClipId);` Doesn't work if the audio clip field is null, but seems like `==` works! So I changed it

unity-client/Assets/Scripts/MainScripts/DCL/Components/Animator/DCLAnimator.cs
sandrade-dcl: I would like to understand better the use of the `GetHashCode()` function and this random unique number, but if we have to use it, as Pato said, I think we should create some centralized place where we can store all these numbers. This will help us to avoid to repeat them in future components.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioClip.cs
```diff
@@ -42,6 +66,21 @@ public DCLAudioClip(IParcelScene scene) : base(scene)
...
+
+        public bool GetIsLoop()
+        {
+            return ((Model)model).loop;
+        }
+
+        public bool GetShouldTryLoad()
+        {
+            return ((Model)model).shouldTryToLoad;
+        }
```
sandrade-dcl: Only as suggestion: Would not it be cleaner to use properties for giving these values?

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLLockedOnEdit.cs
```diff
@@ -14,12 +14,26 @@
...
-    public Model model;
+        public override bool Equals(object obj)
+        {
+            return obj is Model model &&
+                   isLocked == model.isLocked;
+        }
+
+        public override BaseModel GetDataFromJSON(string json)
+        {
+            throw new System.NotImplementedException();
```
sandrade-dcl: Is this temporal?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/EntityInformation/EntityInformationController.cs
```diff
@@ -125,7 +125,7 @@ public void SetEntity(DCLBuilderInWorldEntity entity, ParcelScene currentScene)
         if (entity.HasSmartItemComponent())
         {
             if(entity.rootEntity.TryGetBaseComponent(CLASS_ID_COMPONENT.SMART_ITEM, out BaseComponent baseComponent))
-                smartItemListView.SetSmartItemParameters(entity.GetSmartItemParameters(), ((SmartItemComponent) baseComponent).model.values);
+                smartItemListView.SetSmartItemParameters(entity.GetSmartItemParameters(), ((SmartItemComponent) baseComponent).GetValues());
```
sandrade-dcl, line 128: Reminder for Santi: It will provoke conflicts on your current branch -> `chore/BuildModeHUD-refactorization`

sandrade-dcl (changes requested):

sandrade-dcl (approved):
As the `GetHashCode()` topic is already talked by others reviewers and, apart of this, I only put some suggestions, I'm going to approve.

unity-client/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLLockedOnEdit.cs
```diff
@@ -14,13 +14,16 @@
...
 {
     [System.Serializable]
-    public class Model
+    public class Model : BaseModel
     {
         public bool isLocked;
+        
+        public override BaseModel GetDataFromJSON(string json)
+        {
+            return Utils.SafeFromJson < Model > (json);
```
BrianAmadori:
```suggestion
            return Utils.SafeFromJson<Model>(json);
```

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioClip.cs
```diff
@@ -42,6 +47,12 @@ public DCLAudioClip(IParcelScene scene) : base(scene)
             loadingState = LoadState.IDLE;
         }
 
+        public double Volume => ((Model)model).volume;
+
+        public bool IsLoop => ((Model)model).loop;
+        
+        public bool ShouldTryLoad => ((Model)model).shouldTryToLoad;
```
BrianAmadori: This naming doesn't match with the style on the rest of our codebase, we should switch to lower camelCase here.

BrianAmadori (approved):
LGTM codewise. Approving with some small remarks.

pravusjif (approved):
Tested the build a lot and didn't see any difference with master's build

sandrade-dcl (approved):
I have tested the build and I have not found any problem :)

PR #2105, opened 2021-02-22
feat: upgrade decentraland-connect

2fd:
# What?

Drop custom implementation of Formatic and replace it with `decentraland-connect@2`

# Why? 

In order to maintain a coherent experience when users login into our products

> There is a change of dependencies, you will need to run `npm install` in `kernel` and `website` directories


pravusjif (changes requested):
I tested this branch in local and running the `make watch` for the kernel triggers the following error
Screen Shot 2021-02-24 at 18 33 37

Then, after the server builds, trying to access the local build shows the following error
Screen Shot 2021-02-24 at 19 06 41

After these issues are fixed we should also test in preview mode with the `ENABLE_WEB3` URL param

pablitar:
> I tested this branch in local and running the `make watch` for the kernel triggers the following error
> Screen Shot 2021-02-24 at 18 33 37
> 
> Then, after the server builds, trying to access the local build shows the following error
> Screen Shot 2021-02-24 at 19 06 41
> 
> After these issues are fixed we should also test in preview mode with the `ENABLE_WEB3` URL param

Have you run `npm install` in the `website` folder? This error may be caused by that.

pablitar:
@2fd This PR changes the behaviour so that when a user already has a session, they don't have to click on "Play" to go to the loading screen. Instead, it goes there directly. I'm not sure we want to do that. We actually required the users to click on play in order for them to need to interact with the page, allowing us to bypass some chrome throttling.

Was this intentional? Do we want to not require the returning users to click play anymore?

2fd:
@pablitar yes, it was intentional and it the expected behaviour, but if there is any problem with some browser I can reverted, let me know if it is the case

2fd:
@pravusjif try to `npm install` in `kernel` and `website` directories as @pablitar said

pablitar (approved):
> @pablitar yes, it was intentional and it the expected behaviour, but if there is any problem with some browser I can reverted, let me know if it is the case

Alright. I think is OK. Chrome works in mysterious ways, but I think we only need the "click" for the first time the user logs in.

LGTM!

pravusjif:
After the `npm i` you guys suggested, now the server compiles fine but I get this crash as soon as I enter the local website

Screen Shot 2021-03-01 at 16 23 03

(I can't test the implementation in preview mode until these initializing crashes are fixed =S)

pravusjif:
> After the `npm i` you guys suggested, now the server compiles fine but I get this crash as soon as I enter the local website
> 
> Screen Shot 2021-03-01 at 16 23 03
> 
> Tested also updating the local renderer (`make update-renderer` + `make update-renderer-branch`), also in incognito mode
> 
> (I can't test the implementation in preview mode until these initializing crashes are fixed =S)

apparently this is happening in master as well, I'll let the team know.

pravusjif:
running `npm i` inside 'kernel/' and another inside 'website/' solved my problem.

Tested accessing prod with `ENV=org` and also using a scene in preview mode with and without `ENABLE_WEB3` url param.

pravusjif (approved):

PR #2117, opened 2021-02-25
fix: Show existing session message correctly

pablitar:
We are currently not showing the "existing session" message when it happens, and instead, we are showing an error message.

We can reduce the amount of mails we receive in developers@decentraland.org with this fix.


agusaldasoro (approved):

pravusjif (approved):
tested multi-session logging and the error shown is working as expected.

PR #2118, opened 2021-02-25
feat: changed error message in fatal error screen.

marcosnc:
We changed the fatal error message as follows:
image

To raise the fatal error page you can execute the following command in the JS console:
`globalThis.globalStore.dispatch({type:'fatal error', payload: {type:'fatal'}})`


kernel/static/index.html
```diff
@@ -664,15 +737,13 @@
...
-          feel free to reach out to us at <br /> <a
-            href="mailto:developers@decentraland.org">developers@decentraland.org</a><br /><br />Thank you for helping
-          us
-          improve!
+        <div class="error">Oops!</div>
+        <div class="communicationslink">Something went wrong</div>
+        <div class="givesomedetailof">If you have any ad blocking extensions, try turning them off for this
+          site.<br /><br />
+          Loading should not take any longer than 2-3 minutes. <br />
+          If you seem to be stuck, make sure hardware acceleration is on.
```
BrianAmadori: We have evidence of this being the cause of loading hang-ups? We have an specific message for video cards that do not support WebGL 2.0, and AFAIK this is not fixed by turning hardware acceleration on.

I think we should suggest to reload the page too, as a lot of times the client crashes due to profile loading failures or failures on some of our services.

marcosnc: @RominaMarchetti What do you think? Should we change the message?

RominaMarchetti: Yes, it's fine.

RominaMarchetti (approved):

PR #2123, opened 2021-02-26
chore: AB conversor dump single asset

pravusjif:
Implementation to be able to dump single assets instead of being forced to dump a whole scene, for debugging purposes.


D4rWiNSS (approved):

unity-client/Assets/ABConverter/Core.cs
```diff
@@ -106,6 +111,44 @@ void UpdateLoop()
 
                         env.assetDatabase.Refresh();
 
+                        if (assetsAlreadyDumped && DataStore.i.ABConversorGLTFMissingDependencies.Count > 0 && sceneCid != null)
```
BrianAmadori: Please we can move the `DataStore` variables to a `ABConverter` local storage class? We should aim to have the `ABConverter` as decoupled as possible for the rest of our codebase, as it should deserve its own repo in the future. (the only reason we can't do this right now is the GLTF importer dependency).

unity-client/Assets/UnityGLTF/Scripts/Loader/FileLoader.cs
```diff
@@ -38,6 +38,10 @@ private IEnumerator LoadFileStream(string rootPath, string fileToLoad)
             if (!File.Exists(pathToLoad))
             {
                 Debug.LogError($"Buffer file not found ({pathToLoad}) -- {fileToLoad}");
+
+                if(DCL.DataStore.i.ABConversorRunning && !DCL.DataStore.i.ABConversorGLTFMissingDependencies.Contains(fileToLoad))
```
BrianAmadori: GLTF importer shouldn't know about `ABConverter`. We should expose this behaviour through an additional setting on `GLTFImporter` or something like that.

BrianAmadori (commented):
It looks good! But we should address the encapsulation issues because these will produce technical debt that it will be hard to fix in the future.

unity-client/Assets/UnityGLTF/Scripts/Loader/FileLoader.cs
BrianAmadori: To add a more solid proposal:

* We can add a new parameter to the Editor GLTF Importer (just add another public field to `GLTFImporter` class)
* On `ABConverter.Core`, after copying the files, we can retrieve the importer distance using `AssetImporter.GetAtPath` ([docs](https://docs.unity3d.com/ScriptReference/AssetImporter.GetAtPath.html)). We can then set the parameter of the importer. On some cases, `AssetImporter.GetAtPath` wouldn't do the trick because we have to wait a frame, if this is the case, we can just try to make the setting static. This second approach is a bit messy though.
* On `GLTFImporter` we can set an additional parameter to the created `GLTFSceneImporter` using our new parameter ([this code](https://github.com/decentraland/explorer/blob/f05cb55cc33c717923f3545c3ee3bff84288a1e6/unity-client/Assets/UnityGLTF/Scripts/Editor/GLTFImporter.cs#L499-L504))

unity-client/Assets/ABConverter/Core.cs
pravusjif: The refactor to detect dependencies first and add them to the rawContents to be downloaded made this change unneeded so I reverted it.

unity-client/Assets/UnityGLTF/Scripts/Loader/FileLoader.cs
pravusjif: The refactor to detect dependencies first and add them to the rawContents to be downloaded made this change unneeded so I reverted it.

D4rWiNSS (approved):

unity-client/Assets/ABConverter/Client.cs
```diff
@@ -196,6 +196,60 @@ public static Core.State ConvertScenesToAssetBundles(List<string> sceneCidsList,
...
+
+            List<ContentServerUtils.MappingPair> rawContents = new List<ContentServerUtils.MappingPair>();
+            rawContents.Add(new ContentServerUtils.MappingPair
+            {
+                file = assetFilename,
+                hash = assetHash
+            });
+
+            // If the asset is a GLTF we add the dependencies to the rawContents to be downloaded
+            if (assetFilename.Contains(".glb") || assetFilename.Contains(".gltf") || assetFilename.Contains(".GLB") || assetFilename.Contains(".GLTF"))
```
sandrade-dcl: What do you think about reducing the number of conditions using `toLower()` in this way?

```suggestion
            if (assetFilename.toLower().Contains(".glb") || assetFilename.toLower().Contains(".gltf"))
```

sandrade-dcl (approved):
LGTM! I only left a suggestion 👍

AjimenezDCL (approved):

unity-client/Assets/ABConverter/Client.cs
pbosio: Is `assetFilename` expected to have another extension while being a GLB?
I mean, Is there any particular reason of using `Contains` instead of `EndsWith`?
ie: `assetFilename.EndsWith(".gltf", StringComparison.OrdinalIgnoreCase);`

pravusjif: true that, `EndsWith()` should work as well, I'll try it out

unity-client/Assets/ABConverter/Core.cs
```diff
@@ -160,13 +163,92 @@ void UpdateLoop()
                         OnFinish: (skippedAssetsCount) =>
                         {
                             this.skippedAssets = skippedAssetsCount;
+
+                            if(this.skippedAssets > 0)
+                                state.lastErrorCode = ErrorCodes.ASSET_BUNDLE_BUILD_FAIL;
```
BrianAmadori: This is not entirely true as we might have converted most assets successfully. In the worst case, can we have an alternative error code for this case?

unity-client/Assets/UnityGLTF/Scripts/Editor/GLTFImporter.cs
```diff
@@ -128,9 +131,10 @@ public override void OnImportAsset(AssetImportContext ctx)
                 var meshHash = new HashSet<UnityEngine.Mesh>();
                 var meshFilters = gltfScene.GetComponentsInChildren<MeshFilter>();
                 var vertexBuffer = new List<Vector3>();
-                meshes = meshFilters.Select(mf =>
+                meshes = meshFilters.Where(mf => mf.sharedMesh != null).Select(mf =>
```
BrianAmadori, line 134: Why we do need this additional check?

BrianAmadori (approved):
Awesome work, approving anyway, but left a small request and some questions.

unity-client/Assets/UnityGLTF/Scripts/Editor/GLTFImporter.cs
pravusjif: one time i saw a null ref right there

unity-client/Assets/ABConverter/Core.cs
pravusjif: Updated it with a new error code, tell me what you think

BrianAmadori (approved):

pbosio (approved):

PR #2124, opened 2021-03-01
chore: update test scenes readme

pravusjif:
Clarified test-scenes usage instructions

### DIFF
--- a/docs/how-to-test-parcels-and-preview-scenes.md
+++ b/docs/how-to-test-parcels-and-preview-scenes.md
@@ -4,22 +4,25 @@
 
 You can test and debug components and SDK functionalities with many test-scenes located at `kernel/public/test-scenes/`, running the client in `debug` mode.
 
-1. Test scenes must be built once locally before using them by running:
+1. Start the regular kernel server in a terminal at '/kernel' and leave it watching for changes
 
-    make test-scenes
+    `make watch`
 
-Note that the test-scenes building may take a while.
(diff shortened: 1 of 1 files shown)


docs/how-to-test-parcels-and-preview-scenes.md
```diff
@@ -1,27 +1,30 @@
...
 
     http://localhost:3000/?DEBUG_MODE
 
-To spawn in a specific set of coordinates append the following query paramter:
+3. To spawn at the specific coordinates of a test scene append the position query paramter:
 
     http://localhost:3000/?DEBUG_MODE&position=10,10
 
 
+4. (optional) If you want the test scenes to be updated dynamically after the first build (to be able to modify the scenes code and see the changes by reloading the browser), you should run `make watch-only-test-scenes` in another terminal (as you'd also need a terminal running the regular `make watch` in parallel) and keep it watching for changes.
```
pbosio: just a suggestion/discussion starter (maybe?)
this option seems to be the simpler and probably the most used, also there is not much difference using this option instead of the previous one.
so... can we put it as the default option instead of optional?

pbosio (approved):

BrianAmadori (approved):

docs/how-to-test-parcels-and-preview-scenes.md
pravusjif: Good call, I think that's better too, take a look now at 
https://github.com/decentraland/explorer/blob/5eb95de5d6791ca2ea0e855609a7b0778c9164a4/docs/how-to-test-parcels-and-preview-scenes.md

pbosio (approved):

PR #2133, opened 2021-03-02
fix: Names are no longer being temporarily overridden when updating profile

pablitar:
Foregoing of removing some server properties, since we need them for local profile repo


nchamo (approved):
Great work!

PR #2134, opened 2021-03-02
[WIP] feat: add support for v2 wearables

nchamo:
We are now adding support for v2 wearables. It will be disabled by default, but there is a feature flag that can turn it on

## Feature flag
**Profile ownership sanitization** (this means removing wearables that are not longer owned by the user from their profile):
If the feature flag is off:
* Sanitization will be done on the explorer, by querying wearables-api
* Profile fetching is done by querying `/profile` lambda

If feature flag is on:
Profile fetching will be done by querying `/profiles` lambda, that already filters wearables by ownership

**Wearable fetching** 
If the feature flag is off:
Fetching will be done from the catalog, that is pre-loaded with all wearables from the wearable-api

If the feature flag is on:
Fetching is done from the catalyst server, on-demand

**Wearable ownership** 
If the feature flag is off:
Owned wearables are calculated by the wearable-api

If the feature flag is on:
Owned wearables are calculated by the catalyst

## Mappings
The new v2 wearables are different both in id and payload from the current v1 wearables. So the approach will be the following:
1. The renderer will continue to get both the legacy id and payload. We will translate the ids both in the wearable and profiles
2. When fetching by wearable id on the catalysts, we will translate from legacy id to urns
3. In the future, when we delete the feature flag, we can update the renderer and remove the mappings

## Testing
Link V1: https://play.decentraland.zone/branch/feat/add-support-for-v2-wearables?CATALYST=peer-testing.decentraland.org&ENV=org

Link V2: https://play.decentraland.zone/branch/feat/add-support-for-v2-wearables?CATALYST=peer-testing.decentraland.org&ENV=org&ENABLE_WEARABLES_V2


kernel/test/unit/catalog.saga.test.tsx
```diff
@@ -6,20 +6,20 @@ import { baseCatalogsLoaded, getExclusiveCatalog, getPlatformCatalog } from 'sha
 import { ensureRenderer } from 'shared/renderer/sagas'
 import { throwError } from 'redux-saga-test-plan/providers'
 
-const wearableId1 = 'WearableId1'
+const wearableId1 = 'dcl://base-avatars/name'
```
nchamo, line 9: This change was done because now, when a legacyId is requested, it is translated into urn before being sent to the catalyst

kernel/packages/shared/catalogs/types.ts
```diff
@@ -16,6 +16,8 @@ export type Wearable = {
   replaces?: string[]
   rarity: RarityEnum
   representations: BodyShapeRepresentation[]
+  i18n: { code: string; text: string }[]
```
nchamo, line 19: These were already present in all wearables, but they weren't explicit on the type definition

kernel/packages/shared/catalogs/sagas.ts
```diff
@@ -168,6 +152,168 @@ export function* handleWearablesRequest(action: WearablesRequest) {
...
 }
 
+function* fetchWearablesV2(filters: WearablesRequestFilters) {
+  const catalystUrl = yield select(getCatalystServer)
+  const downloadUrl = yield select(getFetchContentServer)
+  const client: CatalystClient = new CatalystClient(catalystUrl, 'EXPLORER')
+
+  const result: any[] = []
+  if (filters.ownedByUser) {
+    // TODO: What about ALL_WEARABLES?
```
nchamo, line 162: The equivalent of this flag in v2 will be tackled in another PR

kernel/packages/shared/catalogs/sagas.ts
```diff
@@ -168,6 +152,168 @@ export function* handleWearablesRequest(action: WearablesRequest) {
...
+      return `dcl://${result.registry}/${result.id}`
+    } else if (result?.type === 'blockchain-collection-v1') {
+      return `dcl://${result.collectionName}/${result.id}`
+    }
+  } catch {
+    return undefined
+  }
+}
+
+function* fetchWearablesV1(filters: WearablesRequestFilters) {
```
nchamo, line 283: All this was just moved

nchamo:
Replaced by https://github.com/decentraland/explorer/pull/2169

PR #2142, opened 2021-03-03
fix: Increment wearables requests timeout

sandrade-dcl:
Increment wearables requests timeout (`REQUESTS_TIME_OUT`) to 5 minutes.


pbosio (approved):
LGTM! 
just leaving an opinion here, I personally prefer when declaring a constant for seconds to use Minutes * 60 so I avoid the people reading the code to do the math on their heads :P
like `private const float REQUESTS_TIME_OUT = 5 * 60;`

BrianAmadori (approved):

PR #2163, opened 2021-03-04
fix: canvas size adjustments

BrianAmadori:
The canvas used for rendering on "match 1080p" and "match 720p" settings was 40% too big because it didn't take into account the browser bar.

Changed the calculations to take this into account. This should give a performance boost.


kernel/packages/unity-interface/UnityInterface.ts
```diff
@@ -60,12 +60,15 @@ function fillMouseEventDataWrapper(eventStruct: any, e: any, target: any) {
...
 
 function resizeCanvas(module: any) {
+  let desiredHeight = 1080
+
   if (targetHeight > 2000) {
     targetHeight = window.innerHeight * devicePixelRatio
+    desiredHeight = targetHeight
+  } else {
+    desiredHeight = targetHeight - (window.outerHeight - window.innerHeight) * devicePixelRatio
   }
```
pbosio: since `desiredHeight` will be always modified, can we just set one of the possibilities at the beginning to reduce the if/else?

```suggestion
function resizeCanvas(module: any) {
  let desiredHeight = targetHeight - (window.outerHeight - window.innerHeight) * devicePixelRatio

  if (targetHeight > 2000) {
    targetHeight = window.innerHeight * devicePixelRatio
    desiredHeight = targetHeight
  }
```

pbosio (approved):
small non-important suggestion that you can dismiss if you want to :)

pravusjif (approved):

kernel/packages/unity-interface/UnityInterface.ts
BrianAmadori: Changed the code but not exactly what you suggested. Didn't want to put the expensive operation as the default one and also had some doubts about the desired/target height as the names weren't exactly friendly, but in any case, the code should be way more readable now.

pbosio (approved):

PR #2166, opened 2021-03-05
feat: particle effects for avatar

AjimenezDCL:
Particle effect for avatar spawning and wearable changes.


unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
```diff
@@ -302,12 +308,19 @@ private IEnumerator LoadAvatar()
...
             eyesController.Load(bodyShapeController, model.eyeColor);
             eyebrowsController.Load(bodyShapeController, model.hairColor);
             mouthController.Load(bodyShapeController, model.skinColor);
 
             yield return new WaitUntil(() => eyebrowsController.isReady && eyesController.isReady && mouthController.isReady);
 
+            if (useFx && (bodyIsDirty || wearablesIsDirty))
+            {
+                Instantiate(fxSpawnPrefab).transform.position += transform.position;
+                Instantiate(fxChangeWearablePrefab).transform.position += transform.position;
```
sandrade-dcl: Only by curiossity (maybe it is designed in this way): Are both effects (spawn and wearable change) always run together? Even if an user is already loaded into the world and only change the wearable? In this case, shouldn't we see only the wearable change fx?

sandrade-dcl (approved):
Nice effects :)

pravusjif (approved):
Looks neat, I'd decrease a bit the particle area radius to cover only the own avatar personal space but that's just me. Nice job!

unity-client/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
AjimenezDCL: Yes, it is by design. Maybe we need some renaming there to make it more explicit.

pravusjif (approved):
Ran a last test on the build and everything was great, awsome job Alex and Shibu

PR #2169, opened 2021-03-05
[WIP] feat: add support for v2 wearables

nchamo:
We are now adding support for v2 wearables. It will be disabled by default, but there is a feature flag that can turn it on

## Feature flag
**Profile ownership sanitization** (this means removing wearables that are not longer owned by the user from their profile):
If the feature flag is off:
* Sanitization will be done on the explorer, by querying wearables-api
* Profile fetching is done by querying `/profile` lambda

If feature flag is on:
Profile fetching will be done by querying `/profiles` lambda, that already filters wearables by ownership

**Wearable fetching** 
If the feature flag is off:
Fetching will be done from the catalog, that is pre-loaded with all wearables from the wearable-api

If the feature flag is on:
Fetching is done from the catalyst server, on-demand

**Wearable ownership** 
If the feature flag is off:
Owned wearables are calculated by the wearable-api

If the feature flag is on:
Owned wearables are calculated by the catalyst

## Mappings
The new v2 wearables are different both in id and payload from the current v1 wearables. So the approach will be the following:
1. The renderer will continue to get both the legacy id and payload. We will translate the ids both in the wearable and profiles
2. When fetching by wearable id on the catalysts, we will translate from legacy id to urns
3. In the future, when we delete the feature flag, we can update the renderer and remove the mappings

## Testing
Link V1: https://play.decentraland.zone/branch/feat/wearables-v2?CATALYST=peer-testing.decentraland.org&ENV=org

Link V2: https://play.decentraland.zone/branch/feat/wearables-v2?CATALYST=peer-testing.decentraland.org&ENV=org&ENABLE_WEARABLES_V2

### DIFF
--- a/kernel/package-lock.json
+++ b/kernel/package-lock.json
@@ -216,6 +216,85 @@
       "resolved": "https://registry.npmjs.org/@dcl/urn-resolver/-/urn-resolver-1.0.1.tgz",
       "integrity": "sha512-FueQGxAZzh9EBltms0w/jaLng1bdt04OMozWX1dXcQ5B7tnS33n9J55rK8NpcFhRElrfG0DSf5namR5BfyCevA=="
     },
+    "@eslint/eslintrc": {
+      "version": "0.4.0",
+      "resolved": "https://registry.npmjs.org/@eslint/eslintrc/-/eslintrc-0.4.0.tgz",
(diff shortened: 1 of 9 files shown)


nchamo:
Replaced by https://github.com/decentraland/explorer/pull/2172

PR #2178, opened 2021-03-08
feat: add helper for searching users by owned names

pbosio:


kernel/packages/shared/web3.ts
```diff
@@ -138,6 +138,27 @@ export async function fetchOwner(url: string, name: string) {
...
+        owner{
+          address
+        }
+      }
+    }`
+
+  const variables = { name: name.toLowerCase() }
+
+  try {
+    const resp = await queryGraph(url, query, variables)
```
BrianAmadori:
```suggestion
    const response = await queryGraph(url, query, variables)
```

unity-client/Assets/Scripts/MainScripts/DCL/Bridges/UsersSearchBridge/UsersSearchBridge.cs
```diff
@@ -0,0 +1,42 @@
...
+        if (i != null)
+        {
+            Utils.SafeDestroy(this);
+            return;
+        }
+
+        i = this;
+    }
+
+    public void SetENSOwnerQueryResult(string payload)
```
BrianAmadori, line 33: I'd be nice to have this method documented somehow.

kernel/packages/shared/web3.ts
```diff
@@ -138,6 +138,27 @@ export async function fetchOwner(url: string, name: string) {
   }
 }
 
+export async function fetchENSOwnersContains(url: string, name: string, maxResults: number) {
+  const query = `
+    query GetOwner($name: String!) {
```
BrianAmadori: We query the graph with this and that's it? We have our username search? 🎉 

Maybe it's nice to have this method documented as well? As `fetchENSOwnersContains` name its a bit opaque.

BrianAmadori (approved):
Left approve with some minor remarks. Great job!

kernel/packages/shared/web3.ts
```diff
@@ -138,6 +138,33 @@ export async function fetchOwner(url: string, name: string) {
...
+          address
+        }
+      }
+    }`
+
+  const variables = { name: name.toLowerCase(), maxResults }
+
+  try {
+    const response = await queryGraph(url, query, variables)
+    return response.nfts.length > 0 ? response.nfts.map((nft: any) => nft.owner.address as string) : []
```
nchamo:
```suggestion
    return response.nfts.map((nft: any) => nft.owner.address as string)
```

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -377,6 +378,45 @@ export class BrowserInterface {
...
 
+  public SearchENSOwner(data: { name: string; maxResults: number }) {
+    let profilesPromise
+
+    if (/^0x[a-fA-F0-9]{40}$/.test(data.name)) {
+      profilesPromise = ProfileAsPromise(data.name, undefined, ProfileType.DEPLOYED).then((profile) => {
+        return [profile]
+      })
+    } else {
+      profilesPromise = new Promise((resolve, reject) => {
```
pablitar:
```suggestion
      profilesPromise = new Promise(async (resolve, reject) => {
```
I think that if you do this you can avoid the anonymous function below

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -377,6 +378,45 @@ export class BrowserInterface {
     globalThis.globalStore.dispatch(updateFriendship(action, userId.toLowerCase(), false))
   }
 
+  public SearchENSOwner(data: { name: string; maxResults: number }) {
+    let profilesPromise
+
+    if (/^0x[a-fA-F0-9]{40}$/.test(data.name)) {
```
nchamo: Could you make this test into a different function with a name like `isStringEthAddress` or similar. So that it is clearer what you are testing

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -377,6 +378,45 @@ export class BrowserInterface {
...
+                profiles.push(profile)
+              })
+            }
+            resolve(profiles)
+          } catch (error) {
+            reject(error)
+          }
+        })()
+        .catch(error => defaultLogger.error(error))
+      })
```
nchamo: What about something like this? I think it would be a little bit more clear

```
let userIds: Promise<string[]> 

if (/^0x[a-fA-F0-9]{40}$/.test(data.name)) {
  addresses = Promise.resolve([data.name])
} else {
  addresses = getAppNetwork()
    .then(net => fetchENSOwnersContains(ethereumConfigurations[net].names, data.name, data.maxResults))
}

const profilesPromise = userIds
  .then(userIds => userIds.map(userId => ProfileAsPromise(userId, undefined, ProfileType.DEPLOYED)))
  .catch(error => defaultLogger.error(error))
```

Disclaimer: haven't tested it

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -377,6 +378,45 @@ export class BrowserInterface {
...
+        return [profile]
+      })
+    } else {
+      profilesPromise = new Promise((resolve, reject) => {
+        (async () => {
+          try {
+            const net = await getAppNetwork()
+            const owners = await fetchENSOwnersContains(ethereumConfigurations[net].names, data.name, data.maxResults)
+            const profiles: Profile[] = []
+            for (let userId of owners) {
```
pablitar:
```suggestion
            const profiles = await Promise.all(owners.map(userId => ProfileAsPromise(userId, undefined, ProfileType.DEPLOYED))
```
If you do this, requests execute in parallel. Otherwise, they'll go one after the other.

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -377,6 +378,45 @@ export class BrowserInterface {
     globalThis.globalStore.dispatch(updateFriendship(action, userId.toLowerCase(), false))
   }
 
+  public SearchENSOwner(data: { name: string; maxResults: number }) {
```
nchamo: Could we set up a max here? What default value are you using right now?

kernel/packages/unity-interface/BrowserInterface.ts
menduz: `import { isAddress } from 'eth-connect'`

kernel/packages/unity-interface/BrowserInterface.ts
menduz: Another approach to this, is to do 

```ts

async function getProfiles() {

// your code with awaits and throws

}



profilesPromise = getProfiles()


```

that would make the code easier to maintain also

kernel/packages/shared/web3.ts
pbosio: done

kernel/packages/unity-interface/BrowserInterface.ts
pbosio: changed this logic for Mr. Chamo suggestion + Promise.all

kernel/packages/unity-interface/BrowserInterface.ts
pbosio: done. thx menduz!

kernel/packages/unity-interface/BrowserInterface.ts
pbosio: Thx for this suggestion! I implemented it with little changes

kernel/packages/unity-interface/BrowserInterface.ts
pbosio: moved the whole logic to it own file `fetchENSOwnerProfile.ts` to make it easier to mantain

kernel/packages/unity-interface/BrowserInterface.ts
pbosio: I made `maxResults` as optional

nchamo (approved):
Great work!

kernel/packages/unity-interface/fetchENSOwnerProfile.ts
```diff
@@ -0,0 +1,19 @@
+import { isAddress } from 'eth-connect'
+import { ethereumConfigurations } from 'config'
+import { fetchENSOwnersContains, getAppNetwork } from 'shared/web3'
+import { Profile, ProfileType } from 'shared/profiles/types'
+import { ProfileAsPromise } from 'shared/profiles/ProfileAsPromise'
+
+export function fetchENSOwnerProfile(name: string, maxResults: number = 1): Promise<Profile[]> {
```
pablitar, line 7: I'd think making this an async function would allow for easier to read code. Then you can use await below, and you wouldn't need to call then or Promise.resolve. I think.

pablitar (approved):
LGTM! One minor opinion that you are free to ignore.

PR #2191, opened 2021-03-10
fix: Teleport not working in portable experiences

sandrade-dcl:
This PR fix the bug: https://github.com/decentraland/explorer/issues/2171

### Testing
Go to https://play.decentraland.zone/branch/fix/Teleport-not-working-in-portable-experiences/index.html?ENV=org and execute the next command in the Chrome console: `spawnPortableExperienceScene("urn:decentraland:off-chain:static-portable-experiences:portal-experience", "main")`


kernel/packages/unity-interface/UnityParcelScene.ts
```diff
@@ -59,7 +60,15 @@ export class UnityPortableExperienceScene extends UnityScene<LoadablePortableExp
     this.worker
       .getAPIInstance(ParcelIdentity)
       .then((parcelIdentity) => {
+        parcelIdentity.land = {
+          baseUrl: this.data.data.land!.baseUrl,
+          mappingsResponse: this.data.data.land!.mappingsResponse,
+          baseUrlBundles: this.data.data.land!.baseUrlBundles,
+          sceneId: this.data.data.land!.cid,
+          sceneJsonData: this.data.data.land!.sceneJsonData
+        }
```
menduz: We should find another mechanism to do this, since PX are not tied to any LAND

kernel/packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -128,7 +130,14 @@ export async function getLoadablePortableExperience(data: {
...
       contents: mappings,
-      icon: sceneJsonData.display?.favicon
+      icon: sceneJsonData.display?.favicon,
+      land: {
+        baseUrl: baseUrl,
+        baseUrlBundles: '',
+        cid: sceneUrn,
+        sceneJsonData: sceneJsonData,
+        mappingsResponse: mappingsResponse
+      }
```
menduz: why is this necessary?

menduz (changes requested):

kernel/packages/unity-interface/UnityParcelScene.ts
sandrade-dcl: Yeah, take a look to my latest commited changes. With the approach of creating a `isPortableExperience` flag that you suggested me, it is not already necessary to set a land for a PE parcel identity.

kernel/packages/unity-interface/portableExperiencesUtils.ts
sandrade-dcl: With my latest commits it is not already needed 👍 

The problem was that the `getSceneData()` function (in `RestrictedActions.ts`) needed to use `this.parcelIdentity.land.sceneJsonData` to check some conditionals. Now, with the `isPortableExperience` flag we don't need to call `getSceneData()` when the parcel identity is a PE.

Please take a look to the current changes, thanks!

menduz (approved):

kernel/packages/shared/apis/ParcelIdentity.ts
```diff
@@ -10,6 +10,7 @@ export interface IParcelIdentity {
 export class ParcelIdentity extends ExposableAPI implements IParcelIdentity {
   land!: ILand
   cid!: string
+  isPortableExperience!: boolean
```
menduz:
```suggestion
  isPortableExperience: boolean = false
```

menduz: Avoid ! whenever possible

sandrade-dcl: Done! Thanks Agus :)

PR #2193, opened 2021-03-11
fix: BIW UI scene catalog button sometimes doesn't activate

sandrade-dcl:
Fix for the issue: https://github.com/decentraland/unity-renderer/issues/48


D4rWiNSS (approved):
Unable to reproduce GJ!

PR #2196, opened 2021-03-11
feat: helper for fetching profiles by id from renderer

pbosio:
Currently kernel is only reporting profiles for users nearby or friends.
In some occasions we might need to get a user's profile before friends are initialized, in this situation profiles aren't available on renderer's catalog yet.
In other occasions we might need to get the user profile of somebody that is not a friend nor is nearby.
This PR adds a helper for fetching users profile by user id, first it will check if profile is available on renderer's catalog and if it's not available it will request it to kernel.


pravusjif (approved):
awsome!

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/UserProfile/Helpers/UserProfileFetcher.cs
```diff
@@ -0,0 +1,82 @@
+﻿using System;
+using System.Collections.Generic;
+using DCL.Helpers;
+using DCL.Interface;
+
+public class UserProfileFetcher : IDisposable
+{
+    private readonly Dictionary<string, List<Promise<UserProfile>>> pendingPromises = new Dictionary<string, List<Promise<UserProfile>>>();
+    private bool isSubscribedToCatalog = false;
```
AjimenezDCL: We can avoid this flag by always unsubscribing before subscribing:
```
UserProfileController.userProfilesCatalog.OnAdded -= OnProfileAddedToCatalog;
UserProfileController.userProfilesCatalog.OnAdded += OnProfileAddedToCatalog;
```

This way we only subscribe once, regardless of how many times the code gets executed.

AjimenezDCL (approved):
Good job, just left a small comment.

unity-client/Assets/Scripts/MainScripts/DCL/UserProfile/Helpers/UserProfileFetcher.cs
pbosio: great suggestion, I've changed it accordingly. thx!

unity-client/Assets/Scripts/MainScripts/DCL/UserProfile/Helpers/UserProfileFetcher.cs
```diff
@@ -0,0 +1,65 @@
...
+        if (UserProfileController.userProfilesCatalog.TryGetValue(userId, out UserProfile profile))
+        {
+            promise.Resolve(profile);
+            return promise;
+        }
+
+        if (!pendingPromises.TryGetValue(userId, out List<Promise<UserProfile>> promisesForUserId))
+        {
+            UserProfileController.userProfilesCatalog.OnAdded -= OnProfileAddedToCatalog;
+            UserProfileController.userProfilesCatalog.OnAdded += OnProfileAddedToCatalog;
```
BrianAmadori: It wouldn't be more simpler to just subscribe to this in the class constructor and unsubscribe on Dispose (and remove all the other subscription code)? This event is only unsubscribed when the `pendingPromises.Count == 0`. This means the subscription will be re-triggered on each `Fetch` call. 

The logic works but looks like error prone if the maintainer doesn't know exactly what is doing.

BrianAmadori (approved):
LGTM! Added a minor remark.

unity-client/Assets/Scripts/MainScripts/DCL/UserProfile/Helpers/UserProfileFetcher.cs
pbosio: I didn't wanted to add the subscription to event inside the constructor cause, imagine this scenario, which I think is the most probable:
1- instantiate this helper
2- want to fetch a couple of profiles
3- I get the profiles I needed and there is no more pending profiles to fetch
4- I don't dispose the helpers cause I will need it later
Result: the helper will keep listening and trying to get ids from the profile catalog every time a profile is added

The subscription is not always retriggered on each Fetch call, It will only be subscribed if the profile is not currently on the catalog...
So unsubscription is done either on dispose or either if there is no pending profiles to wait for, so I'm kind of failing to figure out what you mean to be error prone, could you please elaborate a little bit on that?

PR #2197, opened 2021-03-12
fix: plane shape uv issue

BrianAmadori:
`PlaneShape.ShouldGenerateNewMesh` was returning always false due to the component `model` being set in advance after the refactor.

image


BrianAmadori:
This is a nightly hotfix, will merge without review.

PR #2204, opened 2021-03-15
fix: The scene catalog is not grouping the assets correctly

sandrade-dcl:
This PR fix the issue: https://github.com/decentraland/unity-renderer/issues/22


D4rWiNSS (approved):
I didn't check this bug, but it was pretty simple, sorry about that!

I was wondering if it will make sense to have the data two times in the Datastore, one grouped by category and 1 grouped by asset pack. It will increase memory usage but it will give us more flexibility (I was thinking in the search function). 
Just a thought.

Anyway we can tackle it in another PR

Good Job!

pravusjif (approved):

PR #2205, opened 2021-03-15
refactor: runtime component factory

BrianAmadori:
This PR adds an unified factory for all SDK components. The main idea is:

- Decouple all references to components from `ParcelScene`.
- Remove the limitation of having all `BaseComponent` as `MonoBehaviour` via interfaces and the unified factory.
- After this, we can have an assemblies reference chain that goes `MainScripts -> Factory -> Components`, instead of `MainScripts -> Components`. Removing dependencies burden on `MainScripts` (the main dependency burden comes from components, so we are effectively attacking the bottleneck).

Changes guide:

- Renamed our `BaseComponent` factory to `PoolableComponentFactory` ([link](https://github.com/decentraland/explorer/pull/2205/files#diff-c8fa7c1364a82aefd0a0f4b556f54d2d5f370deb8c138cecd07335d07ec8be50 ))
- Created new `RuntimeComponentFactory` for all the components. `PoolableComponentFactory` is contained inside it.
- Remove `ParcelScene` dependency injection from all components to enable creation via generics.

Test link
https://play.decentraland.zone/branch/refactor/runtime-component-factory/index.html?ENV=org


unity-client/Assets/Scripts/MainScripts/DCL/Components/UI/UIScreenSpace/UIScreenSpace.cs
```diff
@@ -185,6 +179,7 @@ void InitializeCanvas()
             if (!scene.isPersistent)
             {
                 UpdateCanvasVisibility();
+                CommonScriptableObjects.allUIHidden.OnChange += AllUIHidden_OnChange;
```
BrianAmadori, line 182: Had to move this line here because `scene` is still not populated in the constructor now.

unity-client/Assets/Scripts/MainScripts/DCL/Components/BaseDisposable.cs
```diff
@@ -11,7 +11,7 @@ public abstract class BaseDisposable : IComponent
     {
         public virtual string componentName => GetType().Name;
         public string id;
-        public IParcelScene scene { get; protected set; }
+        public IParcelScene scene { get; set; }
```
BrianAmadori, line 14: The setter had to be made public because the `scene` is now injected from outside in `ParcelScene`. The best implementation is to have an `Initialize(scene, id)`. I'm having this implemented in a continuation PR.

pbosio (approved):
LGTM!

pravusjif (approved):
awsome!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/RuntimeComponentFactory.cs
```diff
@@ -0,0 +1,148 @@
...
+using DCL.Controllers;
+using DCL.Helpers;
+using DCL.Models;
+using UnityEngine;
+
+namespace DCL
+{
+    public interface IRuntimeComponentFactory
+    {
+        IComponent CreateComponent(int classId, object model);
```
D4rWiNSS, line 14: this model shouldn't be a `BaseModel` instead of an `object`?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/RuntimeComponentFactory.cs
```diff
@@ -0,0 +1,148 @@
...
+        void Initialize();
+    }
+
+    public class RuntimeComponentFactory : IRuntimeComponentFactory
+    {
+        private delegate IComponent ComponentBuilder(int classId, object model);
```
D4rWiNSS, line 20: Same here

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/RuntimeComponentFactory.cs
```diff
@@ -0,0 +1,148 @@
...
+            return poolableComponentFactory.CreateItemFromId<BaseComponent>((CLASS_ID_COMPONENT) classId);
+        }
+
+        private T CreateComponent<T>(int classId, object model)
+            where T : IComponent, new()
+        {
+            return new T();
+        }
+
+        private UUIDComponent CreateUUIDComponent(int classId, object model)
```
D4rWiNSS, line 107: Same for this 3

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/RuntimeComponentFactory.cs
```diff
@@ -0,0 +1,148 @@
...
+                    break;
+                case OnPointerUp.NAME:
+                    newComponent = go.GetOrCreateComponent<OnPointerUp>();
+                    break;
+            }
+
+            return newComponent;
+        }
+
+        public IComponent CreateComponent(int classId, object model)
```
D4rWiNSS, line 135: Same for this (If it doesn't make sense, just ignore these comments)

unity-client/Assets/Scripts/MainScripts/DCL/Components/UI/UIScreenSpace/UIScreenSpace.cs
AjimenezDCL: I'm concerned about having multiple subscriptions to this event by mistake. Maybe we can use the `initialized` flag or unsubscribe preemptively

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/RuntimeComponentFactory.cs
```diff
@@ -0,0 +1,148 @@
...
+        void Initialize();
+    }
+
+    public class RuntimeComponentFactory : IRuntimeComponentFactory
+    {
+        private delegate IComponent ComponentBuilder(int classId, object model);
+
+        private Dictionary<int, ComponentBuilder> builders = new Dictionary<int, ComponentBuilder>();
+
+        public IPoolableComponentFactory poolableComponentFactory { get; private set; }
```
sandrade-dcl, line 24: Question: Should we add this property in the interface definition? I guess it would ease the testing of the entities that use the `IRuntimeComponentFactory` as dependency.

AjimenezDCL (approved):
Left a comment with a concern, approving anyway, good job!

sandrade-dcl (approved):
Good job! I only left a suggestion 👍

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/RuntimeComponentFactory.cs
BrianAmadori: I removing the `model` parameter in the subsequent PR, so for now I'll leave this code temporarily so it doesn't conflict with that change, it will be fixed later anyways.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/Scene/RuntimeComponentFactory.cs
BrianAmadori: It's tempting to do that!. However, having the `PoolableComponentFactory` inside the general factory its an implementation detail. The runtime factory user wouldn't have to know the details of how the components are constructed, if they are poolable or what.

unity-client/Assets/Scripts/MainScripts/DCL/Components/UI/UIScreenSpace/UIScreenSpace.cs
BrianAmadori: `InitializeCanvas()` is only called if the already existing `initialized` flag is false. So there's no chance of multiple subscriptions in this case. If that was the case, we could be creating many canvases by mistake too, so I can safely bet on this.

PR #2214, opened 2021-03-17
fix: mouse events rare issue

BrianAmadori:
Tentative fix for https://github.com/decentraland/issues/issues/3

Couldn't reproduce the issue because Tominoya devs already found a workaround, so it's a defensive/theoretical fix.

In any case, this should improve the performance a bit because we aren't regenerating all the colliders with every shape parameter update.

### Review guidelines:
See the `OnPointerEventColliders` file. Everything else is auto-format done by the pre-commit hook.


pbosio (approved):

AjimenezDCL (approved):
Looks good!

Out of curiosity, how come the code style has been triggered for files that contains no changes?

sandrade-dcl (approved):

PR #2215, opened 2021-03-17
feat: stickers system

AjimenezDCL:
A system to play stickers on top of our avatars either throuh emotes or through a profile field.

Test-suite: To test this feature properly we need a refactored avatar renderer. In fact most of avatar renderer test (although disabled) are not passing right now. So, once that's tackled down and avatar renderer is easily mockable, we can trigger these animations in a simpler way. If we wanted to do that with the current architecture we would have to rely in the current AvatarRenderer test suite which is quite slow.

Documentation: Once the approach is validated and reviewed in this PR I will document the whole process of creating and implementing new stickers into the system.

Testing the PR: We haven't implemented any sticker currently so testing the system (without hardcoding stuff) is difficult.
I would suggest to check the code, ensure we have no regressions and check the feature working in the next PR once we have some stickers working.


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
```diff
@@ -0,0 +1,20 @@
...
+    private StickersFactory stickersFactory;
+
+    private void Awake() { stickersFactory = Resources.Load<StickersFactory>("StickersFactory"); }
+
+    public void PlayEmote(string id)
+    {
+        if (stickersFactory == null || !stickersFactory.TryGet(id, out GameObject prefab))
+            return;
+
+        GameObject emoteGameObject = Instantiate(prefab);
```
pravusjif, line 14: When/Where is this GO destroyed ?

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
AjimenezDCL: It's a condition of the particle itself to be destroyed automatically. You can do it through the `ParticleSystem` itself or with an util script https://github.com/decentraland/explorer/blob/master/unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/DestroyParticlesOnFinish.cs

D4rWiNSS, line 14: Should we make this instance in a pool? I am thinking about lots of users watching a concert and showing emotes to their artist. What do you think?

D4rWiNSS (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
pravusjif: Awsome, I didn't remember that, thanks!

AjimenezDCL: Yeah, I want to leverage the performance before implementing the pool and I need some stickers to be implemented before hand. What do you think about coming back to this topic once they are set?

sandrade-dcl (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
```diff
@@ -0,0 +1,20 @@
+﻿using UnityEngine;
+
+public class StickersController : MonoBehaviour
+{
+    private StickersFactory stickersFactory;
+
+    private void Awake() { stickersFactory = Resources.Load<StickersFactory>("StickersFactory"); }
```
BrianAmadori, line 7: I have this conundrum:

- If this is a `MonoBehaviour`, why use `Resources.Load`? We can just reference the factory.
- However, it does make sense to make this a `MonoBehaviour`? The `transform` reference can be injected and doesn't need many others or listening to component events.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
```diff
@@ -0,0 +1,20 @@
+﻿using UnityEngine;
+
+public class StickersController : MonoBehaviour
```
BrianAmadori, line 3: Why do we need a specialized stickers controller? Why we can't abstract this out with the rest of our expressions?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
AjimenezDCL: We could indeed reference the factory, but since this script goes in multiple prefabs, I didnt want to bother with nuisances like missing references on any of them. I dont mind using serialization though.

I like this controller to be a `MonoBehaviour` because our Avatar is a bit of a mess currently. There's no "main" controller that instances the `AvatarRenderer` or the `AvatarAnimator`. Them both are autonomous and working as components so to be able to make the `StickersController` a POCO I would need somebody to instantiate it. The `AvatarRenderer` might be a good candidate but I couldn't fit them properly since you can trigger stickers outside of the emotes flow as well.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
D4rWiNSS: Perfect!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
AjimenezDCL: Mostly because our expressions/emotes are not in a controller, they are just triggered by our `AvatarAnimator` script.
I also want to decouple this system from the expressions due to one of the requierements specified here: https://app.zenhub.com/workspaces/unity-6047bad476c3c0001942ee91/issues/decentraland/unity-renderer/9

We will need to trigger stickers outside of the emotes as reactions through the chat or a new UI, so decoupling this right now will help us in the future.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
BrianAmadori: Thanks for the explanation. I agree we need some sort of extra layer that handles the complexity of the avatar. I'd like to discuss later if it would be more convenient to have a controller for all the avatar systems and for the expressions in general as well. I'm worried that the highest layer of abstraction for emotes is the `AvatarAnimator`, as it doesn't sound enough for now.

Going to approve for now but I think we have to improve the design for this stuff.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
BrianAmadori: I think I still don't fully understand the implications of the "sticker" emotes. I see that issue and for me its just another emote but with triggered particles. Expressions can be triggered from chat and UI right now, so I don't see the true difference.

BrianAmadori (approved):
As I said in the comment, I'm approving this but I'd like to discuss the design further next week.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
AjimenezDCL: So we give more flexibility to the users to express themselves. They can wave or dance while triggering a heart or some clapping stickers. They could combine stickers and emotesin any way they want.
It has the advantage of not overriding the current animation  so if you want to show some stickers, you trigger them while running, or jumping or doing whatever. 

And also the system allows our emotes to throw stickers if we want through AnimationEvents so it doesnt collide with anything of the above.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
AjimenezDCL: Yeah the avatar refactoring has been on hold for too long IMHO

PR #2220, opened 2021-03-18
feat: add smart item filter and improve performance

D4rWiNSS:
# What? 

This PR adds the ability to filter smart items with the button in the search bar and improve the performance of the search bar in general

It also changes the back icon sprite on the SceneCatalog according to its state


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuildModeHUD/Scripts/Common/BIWSearchBarController.cs
```diff
@@ -0,0 +1,157 @@
...
+{
+    bool IsFilterActive();
+    void Dispose();
+    List<Dictionary<string, List<CatalogItem>>> FilterAssets(string nameToFilter);
+}
+
+public class BIWSearchBarController : IBIWSearchBarController
+{
+    public event Action<List<Dictionary<string, List<CatalogItem>>>> OnFilterChange;
+    public event Action OnFilterRemove;
```
sandrade-dcl, line 19: I would add these 2 events to the interface too. It will ease the testing.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuildModeHUD/Scripts/Common/BIWSearchBarController.cs
```diff
@@ -0,0 +1,157 @@
...
+    internal List<Dictionary<string, List<CatalogItem>>> filterObjects = new List<Dictionary<string, List<CatalogItem>>>();
+
+    private IBIWSearchBarView view;
+
+    private bool isSmartItemFilterActive = false;
+    private bool isFilterActive = false;
+    public BIWSearchBarController(ISceneCatalogView catalogView)
```
sandrade-dcl: Normally, for the cases where we need to inject dependencies into a class, we are using an `Initialize(...)` method instead of using directly the constructor. It allows us to have a more control (for example from tests).

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuildModeHUD/Scripts/Common/BIWSearchBarController.cs
```diff
@@ -0,0 +1,157 @@
...
+    {
+        this.view = catalogView.searchBarView;
+        if (this.view.searchInput != null)
+            this.view.searchInput.onValueChanged.AddListener(OnSearchInputChanged);
+
+        if (this.view.smartItemBtn != null)
+            this.view.smartItemBtn.onClick.AddListener(ChangeSmartItemFilter);
```
sandrade-dcl, line 35: Shouldn't we unsubscribe it in the `Dispose` method?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Tests/Helpers/BuilderInWorldTestHelper.cs
```diff
@@ -64,6 +64,18 @@ public static void CreateTestCatalogLocalSingleObject()
         }
     }
 
+    public static void CreateTestSmartItemCatalogLocalSingleObject()
+    {
+        AssetCatalogBridge.ClearCatalog();
```
AjimenezDCL: I find a bit strange that some stuff in the `AssetCatalogBridge` is `static` and some are accessed through the singleton. Is there any reasoning behind? can we standarize it?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuildModeHUD/Scripts/Common/BIWSearchBarController.cs
AjimenezDCL: This is a completely fine way of implementing dependency injection but it can raise some issues with cycle dependencies.

To avoid this we are trying to implement DI with an `Initialize` method. Would it be difficult to do it here as well?

sandrade-dcl (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuildModeHUD/Scripts/Common/BIWSearchBarController.cs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    public void Dispose()
+    {
+        if (view.searchInput != null)
+            view.searchInput?.onValueChanged.RemoveListener(OnSearchInputChanged);
```
AjimenezDCL: We dont need the `if` because we are using the ? check.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuildModeHUD/Scripts/Common/BIWSearchBarController.cs
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        if (currentSearchInput.Length <= 1)
+            return;
+        isFilterActive = true;
+        FilterAssets(currentSearchInput);
+        OnFilterChange?.Invoke(filterObjects);
+    }
+
+    public void FreeFilters()
```
AjimenezDCL: `ReleaseFilters` fits better

AjimenezDCL (approved):
Love the use of interfaces and the overall approach. Left a couple of small request, but I'm approving regardless.

sandrade-dcl (approved):
Thanks for apply the feedback! Good work!

PR #2225, opened 2021-03-20
ecs: export ComponentGroup

cruzdanilo:
# What? 
exports `ComponentGroup`

# Why? 
it was missing, apparently by mistake


menduz:
Hi there, thanks for this contribution!

Although we need `ComponentGroup` types, that class was never meant to be "instantiable" from the outside. 

What is the motivation behind this PR? the first or the former?

cruzdanilo:
i use a custom build (webpack + babel + typescript), and i import the types explicitly. one of my classes has a member property which is a reference to a `ComponentGroup`, so i need to import the type to declare it.

PR #2226, opened 2021-03-22
feat: rotation euler representantion

D4rWiNSS:
# What?

The entity information UI shows the current rotation of the entity that you can change through an InputField.

This PR track internally the changes to the euler angles of the entity so we can give a more user-friendly information

 # Why?
The problem is that Quaternions have multiple euler angles representation so when we are asking the euler angles to the unity transform, it gives us only 1 euler representation of the quaternion that can not match the user input even if it is the same rotation.

For example: The user inputs in our entity information Ui a rotation 180,0,0.
We set the rotation to the entity and ask the current transform to give us the result in euler angles.
The transform will give us 0,180,180 instead of 180,0,0. They are the same rotation but the euler representantion change


pravusjif (approved):
noice!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuildModeHUD/Scripts/Common/EntityInformationController.cs
```diff
@@ -86,7 +87,11 @@ public void Dispose()
 
     public void PositionChanged(Vector3 pos) { OnPositionChange?.Invoke(pos); }
 
-    public void RotationChanged(Vector3 rot) { OnRotationChange?.Invoke(rot); }
+    public void RotationChanged(Vector3 rot)
+    {
+        currentEntity?.SetRotation(rot);
+        OnRotationChange?.Invoke(rot);
```
sandrade-dcl, line 93: Please, update the `RotationChangedCorrectly` test in `EntityInformationControllerShould` taking on account this new line added.

sandrade-dcl (approved):

sandrade-dcl:
Hey @Suduck, is it normal that just when we drag a catlog item to the scene, the `Y` rotation is set to something different than 0?

image

PR #2227, opened 2021-03-22
fix: Implement the needed loading feedback for BiW

sandrade-dcl:
This PR fix the issue: https://github.com/decentraland/unity-renderer/issues/177

### Improvements included
- A new **loading panel** (`BuilderInWorldLoadingView / BuilderInWorldLoadingController`) that will be showed when the Builder In World be activated and hidden when all the editor has been loaded. This panel will show a carousel of phrases (random tips) during the loading process, and will let us to cancel and exit from the loading panel in any moment.

image

- A **loading spinner** for each catalog item to indicate when its thumbnail is being loaded.

image

### Testing
1. Go to: https://explorer.decentraland.zone/branch/fix/Implement-the-needed-loading-feedback-for-BiW/index.html?ENV=org&ENABLE_BUILDER_IN_WORLD&position=100,100&realm=fenrir-amber
2. Activate the Builder In World editor pressing the "**K**" key.
3. See the **loading panel** that will be showed.
4. Once the editor has been loaded, click on **Catalog** button and select any category.
5. See the loading spinners that will be showed on each item that is being loaded.


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/UI/Adapters/CatalogItemAdapter.cs
```diff
@@ -60,27 +57,50 @@ public void SetContent(CatalogItem catalogItem)
...
+    }
 
+    private void SetLoadingActive(bool isActive)
+    {
+        if (loadingGO == null)
+            return;
 
-        AssetPromiseKeeper_Texture.i.Forget(loadedThumbnailPromise);
-        loadedThumbnailPromise = newLoadedThumbnailPromise;
-        loadedThumbnailURL = newLoadedThumbnailURL;
```
D4rWiNSS, line 83: We should forget the promise, in case that the adapter get reused.

If we don't forget it , we can have some race conditions and can end it showing an incorrect icon

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuildModeHUD/Scripts/BuilderInWorldLoadingController.cs
```diff
@@ -0,0 +1,35 @@
...
+    event System.Action OnCancelLoading;
+
+    void Initialize(IBuilderInWorldLoadingView initialLoadingView);
+    void Dispose();
+    void Show(bool showTips = true);
+    void Hide(bool forzeHidding = false);
+    void CancelLoading();
+}
+
+public class BuilderInWorldLoadingController : IBuilderInWorldLoadingController
```
D4rWiNSS, line 12: What do you think about making it inherit from `BIWController` like other controllers, it will create an monobehaviour but there is some functionality in the BuilderinworldController that can be implemented here with the references.

This way BuilderInWorldController doesn't touch anything from the loading, and this class control everything. Just to have the same structure for the BuilderInWrold

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
```diff
@@ -429,10 +442,21 @@ public void EnterEditMode()
         RenderSettings.skybox = skyBoxMaterial;
     }
 
+    private void OnAllParcelsFloorLoaded()
+    {
+        biwFloorHandler.OnAllParcelsFloorLoaded -= OnAllParcelsFloorLoaded;
+        initialLoadingController.Hide();
+        inputController.isInputActive = true;
+    }
+
```
D4rWiNSS, line 451: This is the kind of functionality that can go into the controller (Just read all comments and you will understand this)

D4rWiNSS (approved):
Good Job! I have a request of how to handle the controller, but I will approve in case you think that is not neccesary

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/UI/Adapters/CatalogItemAdapter.cs
sandrade-dcl: I'm doing the forget in the `ClearThumbnailPromise()` function that I call just before the `new AssetPromise_Texture(url)`.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuildModeHUD/Scripts/BuilderInWorldLoadingController.cs
sandrade-dcl: I see this loading more as a HUD that have a view that is already a monobehaviour (`BuilderInWorldLoadingView`) and a controller that should only give orders to that view, and this controller is called by any consumer that needs it (in this case the `BuilderInWorldController` itself).

If we would want to extract logic from the `BuilderInWorldController` to the loading controller, for me the correct way would be to inject the dependency we need and then handle it as we want. But anyway, in terms of scalability, I still see that it makes more sense to treat the loading with the current philosophy of "it's just a HUD that is showed/hidden by any consumer that needs it", rather than putting specific BiW core logic in its controller.

Anyway, let see what the rest of the team think about it.

Thanks Adri!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
sandrade-dcl: I have already responded in the previous comment: https://github.com/decentraland/explorer/pull/2227#discussion_r600575647

Thanks for the review, Adri.

sandrade-dcl:
@RominaMarchetti this is the PR that contains the implementation of the loadings for Builder In World. Please take a look (you have the instructions to test in the PR description)! Thanks!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
```diff
@@ -222,7 +229,7 @@ public void Update_Trigger(InputAction_Trigger[] triggerTimeActions)
                     InputProcessor.FromKey(action, KeyCode.Alpha7, modifiers: InputProcessor.Modifier.FocusNotInInput);
                     break;
                 case DCLAction_Trigger.BuildEditModeChange:
-                    InputProcessor.FromKey(action, KeyCode.L, modifiers: InputProcessor.Modifier.FocusNotInInput);
+                    InputProcessor.FromKey(action, KeyCode.KeypadPlus, modifiers: InputProcessor.Modifier.FocusNotInInput);
```
pbosio: This won't work.
It will depend on the keyboard's layout.
We should avoid using any special character in here or upgrade to the new Unity's Input System
It actually doesn't work on my Mac :(

pbosio (changes requested):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
sandrade-dcl: Ups, I didn't know it, sorry! I changed the key associated (previously was the "L" key) because it was entering in conflict with the friends feature that used the same key. I have just change it to "K" key (it seems this key is not currently being used for nothing). Anyway this key is something temporal only for our testings, I guess in the future the BiW editor will be activated only by the panel you're acurrently implementing.

pbosio: thx ❤️ I can test it now :)

pbosio (approved):

PR #2228, opened 2021-03-22
feat: biw correct error handling while working with shapes

D4rWiNSS:
# What? 
This PR makes the loading objects of builder in world to correctly handle errors and removing loading items when necessary

# Why? 
Currently, we had a timeout of 10 seconds to delete the loading objects in case they didn't load, now it is related to the shape component so the loading object will be destroyed when loaded/error

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadableShape.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadableShape.cs
@@ -79,7 +79,7 @@ namespace DCL.Components
     {
         private bool isLoaded = false;
         private bool failed = false;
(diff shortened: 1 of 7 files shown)


pravusjif (approved):
The code looks good, althought i'm not sure if it is a problem of this PR but even if the GLTFs finish loading I still see the "loading" animated mesh

Screen Shot 2021-03-22 at 21 05 37

sandrade-dcl (changes requested):
The code looks fine, but I'm getting the same bug that Pravus said: the 3D loading model keeps on the screen when you drag any item to the scene.

image

sandrade-dcl (approved):
Tested again and the bug has been solved. Goog job!

pravusjif (approved):
good job!

PR #2231, opened 2021-03-22
fix: long textshape crash

AjimenezDCL:
This PR addressed two topics at once:

1) It makes the TextShape compatible with the scene boundaries checker so they are hidden when stepping out of the scene.
2) Fixes an issue where a big enough update component message was able to crash the explorer by setting a maximum size.

### DIFF
--- a/kernel/packages/scene-system/sdk/SceneRuntime.ts
+++ b/kernel/packages/scene-system/sdk/SceneRuntime.ts
@@ -241,6 +241,11 @@ export abstract class SceneRuntime extends Script {
         },
 
         openExternalUrl(url: string) {
+          if (JSON.stringify(url).length > 49000) {
+            console.error("URL payload cannot exceed 49.000 bytes")
+            return;
+          }
+
           if (that.allowOpenExternalUrl) {
             that.events.push({
               type: 'OpenExternalUrl',
@@ -254,14 +259,17 @@ export abstract class SceneRuntime extends Script {
 
         openNFTDialog(assetContractAddress: string, tokenId: string, comment: string | null) {
           if (that.allowOpenExternalUrl) {
+            let payload = { assetContractAddress, tokenId, comment}
+
+            if (JSON.stringify(payload).length > 49000) {
+              console.error("OpenNFT payload cannot exceed 49.000 bytes")
+              return;
+            }
+
             that.events.push({
               type: 'OpenNFTDialog',
               tag: '',
-              payload: {
-                assetContractAddress,
-                tokenId,
-                comment
-              } as OpenNFTDialogPayload
+              payload: payload as OpenNFTDialogPayload
             })
           } else {
             this.error('openNFTDialog can only be used inside a pointerEvent')
@@ -306,6 +314,11 @@ export abstract class SceneRuntime extends Script {
 
         /** called after adding a component to the entity or after updating a component */
         updateEntityComponent(entityId: string, componentName: string, classId: number, json: string): void {
+          if (json.length > 49000) {
+            console.error("Component payload cannot exceed 49.000 bytes")
+            return;
+          }
+
           if (componentNameRE.test(componentName)) {
             that.events.push({
               type: 'UpdateEntityComponent',
--- a/kernel/packages/scene-system/sdk/SceneRuntime.ts
+++ b/kernel/packages/scene-system/sdk/SceneRuntime.ts
@@ -242,7 +242,7 @@ export abstract class SceneRuntime extends Script {
 
         openExternalUrl(url: string) {
           if (JSON.stringify(url).length > 49000) {
-            console.error("URL payload cannot exceed 49.000 bytes")
+            that.onError(new Error("URL payload cannot exceed 49.000 bytes"))
             return;
           }
 
@@ -262,7 +262,7 @@ export abstract class SceneRuntime extends Script {
             let payload = { assetContractAddress, tokenId, comment}
 
             if (JSON.stringify(payload).length > 49000) {
-              console.error("OpenNFT payload cannot exceed 49.000 bytes")
+              that.onError(new Error("OpenNFT payload cannot exceed 49.000 bytes"))
               return;
             }
 
@@ -315,7 +315,7 @@ export abstract class SceneRuntime extends Script {
         /** called after adding a component to the entity or after updating a component */
         updateEntityComponent(entityId: string, componentName: string, classId: number, json: string): void {
           if (json.length > 49000) {
-            console.error("Component payload cannot exceed 49.000 bytes")
+            that.onError(new Error("Component payload cannot exceed 49.000 bytes"))
             return;
           }
 
@@ -428,6 +428,11 @@ export abstract class SceneRuntime extends Script {
         },
 
         componentUpdated(id: string, json: string) {
+          if (json.length > 49000) {
+            that.onError(new Error("Component payload cannot exceed 49.000 bytes"))
+            return;
+          }
+
           that.events.push({
             type: 'ComponentUpdated',
             tag: id,
--- a/kernel/packages/unity-interface/UnityScene.ts
+++ b/kernel/packages/unity-interface/UnityScene.ts
@@ -72,7 +72,7 @@ export class UnityScene<T> implements ParcelSceneAPI {
       const action = actions[i]
       try {
         nativeMsgBridge.SendNativeMessage(sceneId, action)
-      } catch(e) {
+      } catch (e) {
         this.logger.error(e)
       }
     }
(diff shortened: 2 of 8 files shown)


pravusjif (commented):
the code looks good, I'll approve it once I test a build

pbosio (approved):

kernel/packages/unity-interface/UnityScene.ts
```diff
@@ -70,7 +70,11 @@ export class UnityScene<T> implements ParcelSceneAPI {
     const sceneId = getParcelSceneID(this)
     for (let i = 0; i < actions.length; i++) {
       const action = actions[i]
-      nativeMsgBridge.SendNativeMessage(sceneId, action)
+      try {
+        nativeMsgBridge.SendNativeMessage(sceneId, action)
+      } catch (e) {
+        this.logger.error(e)
```
menduz: I think we should throw this error all the way to the scene and then shut it down. Keeping other scenes running

AjimenezDCL: That's a really good point but I need some of the platform guys to help me with this. If they cannot do it in this PR, is it okay to merge it  as it is and create an issue to be taken down whenever time allows?

sandrade-dcl (approved):

kernel/packages/unity-interface/UnityScene.ts
BrianAmadori: Why shut it down? Why not just ignore the message?

kernel/packages/unity-interface/nativeMessagesBridge.ts
```diff
@@ -212,6 +212,9 @@ export class NativeMessagesBridge {
   }
 
   entityComponentCreateOrUpdate(payload: UpdateEntityComponentPayload) {
+    if (payload.json.length > 49000) {
```
BrianAmadori: This check shouldn't be for any `json` payload that has creator inputs?

This error may be triggered by:
- `callSharedComponentUpdate`
- `callOpenNftDialog`
- `callOpenExternalUrl`

BrianAmadori: Also, this check should be on `SceneRuntime.ts` for improved performance (as it would run in the scene worker).

We can implement it like this:

```
        updateEntityComponent(entityId: string, componentName: string, classId: number, json: string): void {
          if (json.length > 49000) {
              return;
          }
          
          if (componentNameRE.test(componentName)) {
            that.events.push({
              type: 'UpdateEntityComponent',
              tag: sceneId + '_' + entityId + '_' + classId,
              payload: {
                entityId,
                classId,
                name: componentName.replace(componentNameRE, ''),
                json: generatePBObject(classId, json)
              } as UpdateEntityComponentPayload
            })
          }
```

And so on for the remaining message types.

kernel/packages/unity-interface/UnityScene.ts
AjimenezDCL: As a dev I would prefer the whole scene to be shut down. You cannot really foresee the consequences of ignoring a single message in the scene flow and it could ruin the experience, cause buggy behaviour or even cost you money.

kernel/packages/unity-interface/nativeMessagesBridge.ts
AjimenezDCL: I applied this, would you mind taking a look again? I also couldn't find anything related to `callSharedComponentUpdate` in `SceneRuntime`. Could you point me in the right direction here?

BrianAmadori: `callSharedComponentUpdate` is `componentUpdated`. The event names are messed up from kernel side.

kernel/packages/unity-interface/UnityScene.ts
D4rWiNSS: I am concerned about the final-user experience thinking that decentraland explorer is broken when the fault is from the scene dev, I think that we should kill the scene and give some feedback to the user rather than just ignore the message and making some strange behaviours

pravusjif: If we shut down the scene, we should display something in the scene itself (in 3D) telling that we had to shut it down or that it had fatal errors.

I preffer ignoring the problematic message and trigger an error in browser console telling exactly **which scene violated the kernel-client messages max-size** and telling that the scene **may display unwanted behaviours** (because of the skipped kernel-client message).

kernel/packages/scene-system/sdk/SceneRuntime.ts
```diff
@@ -306,6 +314,11 @@ export abstract class SceneRuntime extends Script {
 
         /** called after adding a component to the entity or after updating a component */
         updateEntityComponent(entityId: string, componentName: string, classId: number, json: string): void {
+          if (json.length > 49000) {
+            console.error("Component payload cannot exceed 49.000 bytes")
```
menduz:
```suggestion
            that.onError(new Error("Component payload cannot exceed 49.000 bytes"))
```

kernel/packages/scene-system/sdk/SceneRuntime.ts
```diff
@@ -254,14 +259,17 @@ export abstract class SceneRuntime extends Script {
 
         openNFTDialog(assetContractAddress: string, tokenId: string, comment: string | null) {
           if (that.allowOpenExternalUrl) {
+            let payload = { assetContractAddress, tokenId, comment}
+
+            if (JSON.stringify(payload).length > 49000) {
+              console.error("OpenNFT payload cannot exceed 49.000 bytes")
```
menduz:
```suggestion
              that.onError(new Error("OpenNFT payload cannot exceed 49.000 bytes"))
```

kernel/packages/scene-system/sdk/SceneRuntime.ts
```diff
@@ -241,6 +241,11 @@ export abstract class SceneRuntime extends Script {
         },
 
         openExternalUrl(url: string) {
+          if (JSON.stringify(url).length > 49000) {
+            console.error("URL payload cannot exceed 49.000 bytes")
```
menduz:
```suggestion
            that.onError(new Error("URL payload cannot exceed 49.000 bytes"))
```

menduz (changes requested):

menduz:
`this.onError` should be used in the scene to properly track errors and shut down scenes

kernel/packages/unity-interface/UnityScene.ts
sandrade-dcl: I agree with Pravus:
- If we go with the "shut down" option, I think it is completely needed to design any visual feedback that indicates to the final user what is happening. If this feedback does not exist, the experience for the user would be weird and not beneficial for us.
- If we go with the "ignoring message" option, I like what Pravus suggest about simply throw an error message in the console warning about possible unexpected behaviours.

pravusjif (approved):
Unity side looks good to me. Also tested the build with no problems in scenes with textshapes like The Salmonomicon

kernel/packages/unity-interface/UnityScene.ts
BrianAmadori: Let's explore the problem space a bit:

Currently this can only happen in components where the payload contains a user-made string OR maybe with some kind of hack. Let's assume the first.

As far as I'm looking, the components with user made strings in their payloads are:
- TextShape (text)
- LoadableShapes (url)
- NFTShapes (url)
- UIShapes (string alignment, string positions, text)

(Note: All entity ids pass as ints now instead of strings due to the mapping wrapper in `SceneRuntime.ts`, so no way of hacking ids here I think.)

The components that would have most issues with a message missing due to interruptions in the scene interaction would be the UUIDComponents. But as I see so far, UUIDComponents have no user string fields, so no risk here. 

I think we can safely ignore the message, with the only risk being a visual component not being updated in that cycle. The state lies in the worker so no risk of state corruption either.

I see no good reason for shutting down the scene because is not the convention on how we handle issues so far on the world, this will come across as confusing for all the personas in decentraland. Also, its a bit funny that we aren't shutting down scenes for performance reasons but we do so for this microscopic edge case. The better judgement should be to only shut down scenes in extreme cases.

In any case we should not only terminate the worker but send a scene remove message, or the UX will suffer. For the feedback, we can use our already existing notification system (top middle popup message).

kernel/packages/scene-system/sdk/SceneRuntime.ts
AjimenezDCL: done!

kernel/packages/scene-system/sdk/SceneRuntime.ts
AjimenezDCL: done!

kernel/packages/scene-system/sdk/SceneRuntime.ts
AjimenezDCL: done!

menduz (approved):

PR #2234, opened 2021-03-23
fix: remove need for yarn

nchamo:
By updating the `dcl-social-client` library to the latest version, we can now remove the need for yarn


pablitar (approved):
Great work!

PR #2237, opened 2021-03-24
fix: UI Builder In World general feedback

sandrade-dcl:
- [x] Review UI positions.
- [x] Review UI Shrinking behavior.
- [x] Increase icons resolution.

### Testing
https://explorer.decentraland.zone/branch/fix/UI-Builder-In-World-general-feedback/index.html?ENV=org&ENABLE_BUILDER_IN_WORLD&position=100,100&realm=fenrir-amber


D4rWiNSS (changes requested):
Good job! A couple of issues that I have seen:

-I can't see the change camera icon in first person mode (on the top)
-The scene limits icon (the circle with color that represents the amounts of entities) is now very small ( The color part should have more width) and it is like there is a transparency there

sandrade-dcl:
> -I can't see the change camera icon in first person mode (on the top)
> -The scene limits icon (the circle with color that represents the amounts of entities) is now very small ( The color part should have more width) and it is like there is a transparency there

I have fixed it! Thanks Adri.

D4rWiNSS (approved):

PR #2238, opened 2021-03-24
feat: enable external renderer

menduz:
# What? 
Enables external renderers to help in the repository separation of unity

# Why? 
To reduce complexity and decrease coupling

# Testing

To test this feature, go to https://explorer.decentraland.zone/branch/feat/enable-external-renderer/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:master

It will use the renderer from https://github.com/decentraland/unity-renderer/ (master branch)

Other branches can be tested changing the URL https://explorer.decentraland.zone/branch/feat/enable-external-renderer/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:feat/publish-renderer

# Migration plan

1) Validation (this PR): Enable remote loading and validate use case in the ecs@next
2) Migration: 
   1) Publish npm package in every unity master pipeline
   2)
      1) Change this CI to fetch that package (embeded artifacts for ECS & explorer CDN)
      2) Remove pipeline and code from unity from this repository

At every stage of the plan, changes will take into account developer experience, to not loose and enhance the testing of features between components.


menduz:
Screen Shot 2021-03-24 at 14 39 10

 ✅ Preview working

kernel/packages/unity-interface/loader.ts
```diff
@@ -0,0 +1,105 @@
...
+    throw new Error('RENDERER_ARTIFACTS_ROOT is undefined')
+  } else {
+    return new URL(globalThis.RENDERER_ARTIFACTS_ROOT, document.location.toString()).toString()
+  }
+}
+
+// TODO: return type DclRenderer
+async function injectRenderer(baseUrl: string): Promise<LoadRendererResult> {
+  const scriptUrl = new URL('index.js', baseUrl).toString()
+  const buildConfigPath = new URL('unity.json', baseUrl).toString()
```
BrianAmadori, line 29: Since Unity 2020 we have to provide the loader configuration manually. If loading `unity.json` is the only way to provide the loader config we may have to port this code over.

kernel/packages/unity-interface/loader.ts
```diff
@@ -0,0 +1,105 @@
...
+      DclRenderer: globalThis.DclRenderer,
+      UnityLoader: globalThis.UnityLoader,
+      buildConfigPath
+    }
+  }
+
+  // {
+  //   // Load the renderer from the artifacts root folder
+  //   return await injectRenderer(getRendererArtifactsRoot())
+  // }
```
BrianAmadori, line 67:
```suggestion
```

kernel/packages/unity-interface/loader.ts
```diff
@@ -0,0 +1,105 @@
...
+    // urn:decentraland:off-chain:renderer-artifacts:${branch}
+    if (parsedUrn.type == 'off-chain' && parsedUrn.registry == 'renderer-artifacts') {
+      return loadRendererByBranch(parsedUrn.id)
+    }
+
+    throw new Error('It was impossible to resolve a renderer for the URN "' + urn + '"')
+  }
+}
+
+async function injectScript(url: string) {
```
BrianAmadori, line 95: 🌈 🧙

BrianAmadori:
Is the `urn:decentraland:off-chain:renderer-artifacts` absolutely necessary on the URL param? We can't make the build system just infer it?

pbosio:
> Is the `urn:decentraland:off-chain:renderer-artifacts` absolutely necessary on the URL param? We can't make the build system just infer it?

I think it is already inferring it right here https://github.com/decentraland/explorer/blob/48b31daa06dc31c177a187a210b1dad555a646e2/kernel/packages/unity-interface/loader.ts#L77-L79

kernel/packages/unity-interface/loader.ts
menduz: This should remain, read the TODO a couple lines above

kernel/packages/unity-interface/loader.ts
menduz: Can we move that a build with that version into a unity-renderer repo branch? It will be easier to test if we already have it there, and we could do this PR one time only

kernel/packages/unity-interface/loader.ts
```diff
@@ -0,0 +1,105 @@
...
+    UnityLoader: globalThis.UnityLoader,
+    buildConfigPath
+  }
+}
+// TODO: return type DclRenderer
+async function loadDefaultRenderer(): Promise<LoadRendererResult> {
+  // TODO: after we remove unity from this folder, remove the following
+  // block and uncomment the block next to it
+  {
+    const scriptUrl = new URL('DCLUnityLoader.js', getRendererArtifactsRoot()).toString()
```
pbosio, line 51: This will prompt an error in Builder since we are using another loader
(https://github.com/decentraland/builder/blob/master/public/UnityLoader.js)
Otherwise these changes are working fine!

menduz: do you know what changes does that unityloader contains from the original one? or the reason behind that local copy?

pbosio: I have no idea. I could check.
I remember trying once, a lot time ago, to change it to use` DCLUnityLoader.js` instead but I had no success

BrianAmadori:
> I think it is already inferring it right here

@pbosio I mean infer it when we want custom branches, so the URL can work with just: `https://explorer.decentraland.zone/branch/feat/enable-external-renderer/index.html?renderer=feat/publish-renderer`

kernel/packages/unity-interface/loader.ts
pbosio: @menduz So it seems the only important difference between the two loaders is here:
https://github.com/decentraland/explorer/blob/master/kernel/static/unity/Build/DCLUnityLoader.js#L5199-L5200

For builder to work we would need to replace these two lines:
`document.getElementById('gameContainer').classList.add('loaded')`
`document.querySelector('.progress').classList.add('loaded')`
For this one:
`currentProgress = 100 * t + '%'`

menduz:
@BrianAmadori this is so to enable multiple sources without full URL, eventually we may want to use URN to have the most "trustable" renderer on-chain or similar mechanism, to do so we need the URN

kernel/packages/unity-interface/loader.ts
menduz: I verified, it is not being used anywhere else

This will suffice to make the builder work with the new release, after we merge to master we can create a PR in builder to test `@next`:

```ts
// inside builder bundle
UnityLoader.Progress.handler = function(){}
```

kernel/packages/unity-interface/loader.ts
BrianAmadori: But this code will keep living on this repo, right? This is to cover the case of having kernel with a custom renderer build. So we'll have to do this PR only once in any case. (and patch it anyways after the 2020 renderer is live).

I think we can just merge this as-is and patch it afterwards. Bear in mind that we are somehow going to lose compatibility with pre-2020 PRs after we do that. However, the version change will be around next week, so this should be a non-issue.

nchamo (approved):
Great work!

kernel/packages/unity-interface/loader.ts
menduz: OK, please point me to somewhere where I can learn about this change, will merge without it

menduz:
can we merge now? @pbosio @BrianAmadori

pbosio:
> can we merge now? @pbosio @BrianAmadori

yes sir!

pbosio (approved):

PR #2239, opened 2021-03-24
feat: add builder-projects-panel deployed scenes menu section

pbosio:
* add Builder Projects Panel deployed scenes section


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Scripts/SectionController/MenuSections/SectionDeployedScenesController.cs
```diff
@@ -0,0 +1,98 @@
...
+    private readonly SceneSearchHandler sceneSearchHandler = new SceneSearchHandler();
+    private Dictionary<string, SceneCardView> scenesViews;
+
+    public SectionDeployedScenesController(): this(
+        Object.Instantiate(Resources.Load<SectionDeployedScenesView>(VIEW_PREFAB_PATH))
+    )
+    {
+    }
+    
+    public SectionDeployedScenesController(SectionDeployedScenesView view)
```
sandrade-dcl, line 27: Only as suggestion: It would be a good practice to use here an `ISectionDeployedScenesView` interface in order to be able to pass a mocked view and ease the tests for the `SectionDeployedScenesController` controller.

It would let you to completely separate the tests for `SectionDeployedScenesController` and for `SectionDeployedScenesView`.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Scripts/Views/MenuSections/SectionDeployedScenesView.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using DCL.Helpers;
+using UnityEngine;
+using UnityEngine.UI;
+
+internal class SectionDeployedScenesView : MonoBehaviour
```
sandrade-dcl: Related to the pervious suggestion: I think doing an interface for this class would ease the testing for its consummers (that would use NSubstitute for the assertion).

sandrade-dcl (approved):
Very clean code, good job! I only left a suggestion related to "good practices", but it is not something mandatory to apply in this case, I commented it only for sharing my thoughts :)

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Scripts/SectionController/MenuSections/SectionDeployedScenesController.cs
pbosio: I use SectionDeployedScenesView with methods and events of the class without accessing directly any method or property from the monobehavior to make it easy to replace it by an interface if needed. Since I'm not using NSubstitute to mock the view because I was aiming for a more integrated tests I feel is better not to add unnecessarily an interface until we feel that we need it, to extend the tests in the future for example

PR #2241, opened 2021-03-24
feat: Adding realm and guest condition to signed fetch metadata

pablitar:
We are adding more metadata to signed fetch for it to be used with quests. In particular:
- The current realm
- If the user is guest or not


kernel/packages/shared/apis/SignedFetch.ts
```diff
@@ -50,7 +60,13 @@ export class SignedFetch extends ExposableAPI {
         ...getAuthHeaders(
           init?.method ?? 'get',
           path,
-          { sceneId: this.parcelIdentity.cid, parcel: this.getSceneData().scene.base, tld: getTLD() },
+          {
+            sceneId: this.parcelIdentity.cid,
+            parcel: this.getSceneData().scene.base,
+            tld: getTLD(),
+            isGuest: isGuest(),
```
menduz, line 67: just to doublecheck, does this method add the timestamp to the signature? to prevent "double spend" attacks

kernel/packages/shared/apis/SignedFetch.ts
```diff
@@ -50,7 +60,13 @@ export class SignedFetch extends ExposableAPI {
         ...getAuthHeaders(
           init?.method ?? 'get',
           path,
-          { sceneId: this.parcelIdentity.cid, parcel: this.getSceneData().scene.base, tld: getTLD() },
+          {
+            sceneId: this.parcelIdentity.cid,
+            parcel: this.getSceneData().scene.base,
+            tld: getTLD(),
```
menduz, line 66: i'd consider adding the origin of the url along with the tld, to offer more information about the actual deployment we are using, and to enable future `peer-ap1.decentraland.org/play` or `catalyst.myserver.tk/play`

nchamo (approved):
Great work!

Please take a look at @menduz 's comment about adding a timestamp or nonce to detect a replay attack

kernel/packages/shared/apis/SignedFetch.ts
pablitar: Ok. Will add that.

kernel/packages/shared/apis/SignedFetch.ts
pablitar: It is added above: https://github.com/decentraland/explorer/blob/ab068854fe8739a6705ca7e6a547025ab00e302b/kernel/packages/shared/apis/SignedFetch.ts#L29

menduz (approved):

PR #2245, opened 2021-03-25
feat: support min catalyst version

nchamo:
We are now adding the possibility to configure a minimum catalyst version. This is so that the client doesn't connect to catalysts with a really old version that might be incompatible

### Technical details
We basically made these changes:
1. We know read and store the version each catalyst reports
2. We read an optional `minCatalystVersion` from the config
3. When selecting a cached realm, we check its version to see if it is valid
4. All candidates are filtered so that we are sure that their version is valid

### How to test
You can try on zone catalysts. Only peer.decentraland.zone has the correct version.

However, if you add `CATALYST=peer-ue-2.decentraland.zone` to the url, you will be able to connect to that server, regardless of the version


pablitar (approved):
Great work!!

PR #2246, opened 2021-03-25
feat: allow fake ownership of collections

nchamo:
As of now, we have the `ALL_WEARABLES` flag that allows users to simulate as if the owned all existing wearables. 

With the changes made to support decentralized wearables, this is no longer an option. So the idea is to have a new query parameter called `WITH_COLLECTIONS` what would be used like this:
```
&WITH_COLLECTIONS=collectionId1,collectionId2,...
```
This parameter will only work for non-productive environments

### How to test it
Check this link to see how it looks like you own the wearables
https://play.decentraland.zone/branch/feat/allow-fake-ownership-of-collections/index.html?CATALYST=peer-testing.decentraland.org&ENABLE_WEARABLES_V2&WITH_COLLECTIONS=0x102daabd1e9d294d4436ec4c521dce7b1f15499e%2C0x30d3387ff3de2a21bef7032f82d00ff7739e403c

But in a prod environment, it doesn't work
https://play.decentraland.zone/branch/feat/allow-fake-ownership-of-collections/index.html?CATALYST=peer-testing.decentraland.org&ENABLE_WEARABLES_V2&WITH_COLLECTIONS=0x102daabd1e9d294d4436ec4c521dce7b1f15499e%2C0x30d3387ff3de2a21bef7032f82d00ff7739e403c&ENV=org

### DIFF
--- a/kernel/packages/config/index.ts
+++ b/kernel/packages/config/index.ts
@@ -65,9 +65,8 @@ export namespace parcelLimits {
   ]
 
   export const descriptiveValidWorldRanges = validWorldRanges
-    .map(range => `(X from ${range.x.from} to ${range.x.to}, and Y from ${range.y.from} to ${range.y.to})`)
+    .map((range) => `(X from ${range.x.from} to ${range.x.to}, and Y from ${range.y.from} to ${range.y.to})`)
     .join(' or ')
-
 }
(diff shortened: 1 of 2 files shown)


pablitar (approved):
LGTM!

PR #2248, opened 2021-03-26
chore: Implement a custom "WebRequest" that includes a retry system

sandrade-dcl:
## Why?
Our current systems for getting external data (from url's) don't have anything implemented that let us set a number of attemps to retry the request in case it fails.

So far, when we have needed to do this in some specific request, we have had to implement it manually in the consummer code side. We think that it would be a feature very useful to be added in the request system side (for example creating a custom WebRequest class) in order to let us to configure it in a generic way from any place where we send a request.

This would make all our requests more robust.

#### `WebRequestController` : `IWebRequestController`
This class implements all the core logic. It will manage the main lyfecicle of the different requests, including the retry system. It contains these methods:
- `void Initialize(genericWebRequest, assetBundleWebRequest, textureWebRequest, audioWebRequest)`
- `WebRequestAsyncOperation Get(url, OnSuccess, OnFail, requestAttemps, timeout)`: We will use it when we need to requests generic data from any url.
- `WebRequestAsyncOperation GetAssetBundle(url, OnSuccess, OnFail, requestAttemps, timeout)`: We will use it when we need to request an asset bundle.
- `WebRequestAsyncOperation GetTexture(url, OnSuccess, OnFail, requestAttemps, timeout)`: We will use it when we need to request a texture.
- `WebRequestAsyncOperation GetAudioClip(url, audioType, OnSuccess, OnFail, requestAttemps, timeout)`: We will use it when we need to request an audio.

#### `WebRequest` : `IWebRequest`
Our custom implementation of the UnityWebRequest. It implements:
- `UnityWebRequest CreateWebRequest(url)`

#### `WebRequestAssetBundle` : `IWebRequestAssetBundle`
Our custom implementation of the UnityWebRequestAssetBundle. It implements:
- `UnityWebRequest CreateWebRequest(url)`

#### `WebRequestTexture` : `IWebRequestTexture`
Our custom implementation of the UnityWebRequestTexture. It implements:
- `UnityWebRequest CreateWebRequest(url)`

#### `WebRequestAudio` : `IWebRequestAudio`
Our custom implementation of the UnityWebRequestMultimedia (Audio Clip). It implements:
- `void SetAudioType(audioType)`
- `UnityWebRequest CreateWebRequest(url)`

#### `WebRequestAsyncOperation` : `CustomYieldInstruction`
Our custom request async operation to be used with the WebRequestController.

## Using the new WebRequest system
To use our new WebRequest with retry system, you will only have to access to the `WebRequestController` through our `Environment` instance in this way:

`Environment.i.platform.webRequest.GetAsset("some_url")`
`Environment.i.platform.webRequest.GetAssetBundle("some_url")`
`Environment.i.platform.webRequest.GetTexture("some_url")`
`Environment.i.platform.webRequest.GetAudioClip("some_url", audioType)`

**NOTE:** Temporally, until the `Environment` be totally deacoupled and to avoid cyclic references, we can also access to the web request controller through its own static instance in this way:

`WebRequestController.i.GetAsset("some_url")`
`WebRequestController.i.GetAssetBundle("some_url")`
`WebRequestController.i.GetTexture("some_url")`
`WebRequestController.i.GetAudioClip("some_url", audioType)`

## Next steps
Once we have already implemented this custom WebRequest system, we will use it in the entire project. This will be applied in this second PR: https://github.com/decentraland/explorer/pull/2256


unity-client/Assets/Scripts/MainScripts/DCL/Environment/PlatformContextFactory.cs
```diff
@@ -1,18 +1,26 @@
...
+using DCL.Rendering;
 
 namespace DCL
 {
     public static class PlatformContextFactory
     {
         public static PlatformContext CreateDefault()
         {
+            WebRequestController newWebRequestController = new WebRequestController();
+            newWebRequestController.Initialize(
```
BrianAmadori: I'd put this inside a Builder method inside `WebRequestController` to avoid having this code here.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs
```diff
@@ -0,0 +1,58 @@
...
+        void Initialize(
+            IWebRequest webRequest,
+            IWebRequestAssetBundle webRequestAB,
+            IWebRequestTexture webRequestTexture,
+            IWebRequestAudio webRequestAudio);
+    }
+
+    public class WebRequestController : IWebRequestController
+    {
+        public static WebRequestController i { get; private set; }
```
BrianAmadori, line 97:
```suggestion
```

If the access can be retrieved using `Environment` we should remove the `i` to ensure it doesn't get used by mistake.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestBase.cs
```diff
@@ -0,0 +1,96 @@
...
+            UnityWebRequest request;
+            int remainingAttemps = Mathf.Clamp(requestAttemps, 1, requestAttemps);
+
+            do
+            {
+                try
+                {
+                    request = CreateWebRequest(url);
+                    var requestOperation = request.SendWebRequest();
+                    while (!requestOperation.isDone && requestOperation.progress < 1) { }
```
BrianAmadori:
```suggestion
                    while (!requestOperation.isDone && requestOperation.progress < 1) { }
```

This is OK for editor mode requests, but should never happen in runtime. We should always make the requests async by default. I suggest we remove this `Get` method altogether and rename the `GetAsync` to `Get` to avoid this case entirely.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestBase.cs
```diff
@@ -0,0 +1,96 @@
...
+                }
+                catch (HttpRequestException e)
+                {
+                    throw new HttpRequestException($"{e.Message} -- ({url})", e);
```
BrianAmadori: This can't break the logic for failed requests? I think we should just output a `Log.Error` here.

unity-client/Assets/Scripts/MainScripts/DCL/Environment/PlatformContext/PlatformContext.cs
```diff
@@ -14,20 +14,23 @@ public class PlatformContext : System.IDisposable
         public readonly IClipboard clipboard;
         public readonly IPhysicsSyncController physicsSyncController;
         public readonly IDebugController debugController;
+        public readonly IWebRequestController webRequestController;
```
BrianAmadori: This should generate the following syntax that I think is a bit inconvenient:

`Environment.i.platform.webRequestController.webRequestAB.Get();`

I think it can be improved to one of the following:

`Environment.i.platform.fetch.assetBundles.Get();`

`Environment.i.platform.request.assetBundle.Get();`

`Environment.i.platform.webRequest.assetBundle.Get();`

you get the drift.

BrianAmadori (changes requested):

pbosio:
There are a couple of things I'm failing to understand.
Using as example the new `WebRequest` class (https://github.com/decentraland/explorer/blob/1bf79dca46b0435651f89cc745237bc01874dc6d/unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequest.cs)

1- `WebRequest` inherits from `WebRequestBase` and implements interface `IWebRequest` which inherits from interface `IWebRequestBase` BUT `WebRequestBase` is already implementing `IWebRequestBase`?
This spider web of interfaces I find it kind of confusing (It's not straight forward enough for something that should be as simple as a web request), Maybe can we remove the use of those interfaces?
2- It also confuses me a lot that `WebRequestBase` methods are returning unity's `UnityWebRequest` so it feels like an hybrid of our custom request and unity's request... The main problem I find with this is that, as developer, when you receive the `UnityWebRequest` you think that you have total control over the request, but that's not true at all, both `Get` and `GetAsync` methods calls the `SendWebRequest` internally... so if I having a method returning a `UnityWebRequest` I would expect to send the request whenever I want for example.
3- There is no `Dispose` method
4- Why are we having a controller for this? I feel that the only reason for the controller is because those classes should be static instead?
5- Can `GetAsync` use unity's `UnityWebRequestAsyncOperation.completed` event instead of a starting `Coroutine`?
6- Other no-minor issue I can see is that `UnityWebRequest.Abort()` is not being handled, request will be retried even tho the request was explicitly aborted

Maybe a better approach could be to create a custom `UnityWebRequestAsyncOperation` and adding an extension method like `UnityWebRequest.SendWebRequestWithRetries` that returns the custom `UnityWebRequestAsyncOperation`?

unity-client/Assets/Scripts/MainScripts/DCL/Environment/PlatformContextFactory.cs
sandrade-dcl: I agree, what you suggest is cleaner. Done!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestBase.cs
sandrade-dcl: Yes, I implemented the sync GET method thinking in it could be useful in some situation, but you're right, we are not going to use it in any context. I have removed it and renamed the `GetAsync` by `Get`. Thanks Brian.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestBase.cs
sandrade-dcl: As the sync GET method has been removed, this no longer applies. Thanks!

unity-client/Assets/Scripts/MainScripts/DCL/Environment/PlatformContext/PlatformContext.cs
sandrade-dcl: I like you suggestion. Finally I have decided to follow the next way:
`Environment.i.platform.webRequest.generic.Get()`
`Environment.i.platform.webRequest.assetBundle.Get()`
`Environment.i.platform.webRequest.texture.Get()`
`Environment.i.platform.webRequest.audio.Get()`

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs
sandrade-dcl: I had to add the static accessor because of a lot of circular references that I got when I tried to use the `Environment` from several places where we will need to call this new web request (`Environment.platform.webRequest...`). Tomorrow I will take a look in deep in order to see how we could remove these assemby dependencies.

sandrade-dcl:
@pbosio I answer your comments here:
> 1- `WebRequest` inherits from `WebRequestBase` and implements interface `IWebRequest` which inherits from interface `IWebRequestBase` BUT `WebRequestBase` is already implementing `IWebRequestBase`?
> This spider web of interfaces I find it kind of confusing (It's not straight forward enough for something that should be as simple as a web request), Maybe can we remove the use of those interfaces?

First of all, I would like to differentiate between these 2 concepts:
- The classes that we inherit from (`WebRequestBase`)
- The interfaces that we implement (`IWebRequest`, `IWebRequestAssetBundle`, `IWebRequestTexture`, `IWebRequestAudio`). 

The first ones allow us to have all the common implementation in only one place (in this case most of the implementation is included in the base clase), and the second ones are ONLY defining a design structure (in this case it would helping us to mock each class for testing purposes).

That said, in our specific case you're right in that we could remove the `IWebRequest`, `IWebRequestAssetBundle` and `IWebRequestTexture` (because the fact of inherits from `WebRequestBase` already make it to implement the `IWebRequestBase` interface), however we could not remove `IWebRequestAudio` because we have needed to add a new method (`SetAudioType(...)`) not included in `IWebRequestBase`, and if we want to keep mockeable this class just like the other ones, we need to create its own interface.

I don't know the explained reasons are undertandable enough (sorry, it is not easy to explain it here and in english), but we can discuss a bit more about it with the rest of the team.

> 2- It also confuses me a lot that `WebRequestBase` methods are returning unity's `UnityWebRequest` so it feels like an hybrid of our custom request and unity's request... The main problem I find with this is that, as developer, when you receive the `UnityWebRequest` you think that you have total control over the request, but that's not true at all, both `Get` and `GetAsync` methods calls the `SendWebRequest` internally... so if I having a method returning a `UnityWebRequest` I would expect to send the request whenever I want for example.

I understand what you mean, it is a good point. Maybe it would be a better approach to return a `UnityWebRequestAsyncOperation` instead of the own `UnityWebRequest`. I'm going to see it, thanks! [**DONE!**]

> 3- There is no `Dispose` method

You're right, it would help to abort ongoing requests if we destroy the class, I'm going to implement it. Thanks! [**DONE!**]

> 4- Why are we having a controller for this? I feel that the only reason for the controller is because those classes should be static instead?

The reason of having the `WebRequestController` is to group the different types of requests in a single class that ease to access them from a single place. This controller is being injected in the `PlatformContext` class that will be accessed from our `Environment` instance, then we would use each type of request on the next ways:

`Environment.i.platform.webRequest.generic.Get(...)`
`Environment.i.platform.webRequest.assetBundle.Get(...)`
`Environment.i.platform.webRequest.texture.Get(...)`
`Environment.i.platform.webRequest.audio.Get(...)`

> 5- Can `GetAsync` use unity's `UnityWebRequestAsyncOperation.completed` event instead of a starting `Coroutine`?

Relating with the point 2, yes, we could use the `UnityWebRequestAsyncOperation` returned by the `SendWebRequest()` to avoid the coroutine and manage the retries system with the `completed` event. I will take a look, thanks! [**DONE!**]

> 6- Other no-minor issue I can see is that `UnityWebRequest.Abort()` is not being handled, request will be retried even tho the request was explicitly aborted

You're right, I have to handle this kind of situations. Thanks! [**DONE!**]

> Maybe a better approach could be to create a custom `UnityWebRequestAsyncOperation` and adding an extension method like `UnityWebRequest.SendWebRequestWithRetries` that returns the custom `UnityWebRequestAsyncOperation`?

I already thinked about this kind of solution of creating a `SendWebRequestWithRetries()` extension method but discussing it with Brian, I don't remember the specific reason now, we discard it. Anyway we can discuss it with the resto of the team 👍

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs
sandrade-dcl: As we have agreed, we are temporally going with the `WebRequestController.i` approach and we will change it when we have the `Environment` totally decoupled to avoid cyclic references.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestBase.cs
```diff
@@ -0,0 +1,95 @@
...
+                    remainingAttemps--;
+                    if (remainingAttemps > 0)
+                    {
+                        TrySendWebRequest(url, OnSuccess, OnFail, remainingAttemps);
+                    }
+                    else
+                    {
+                        OnFail?.Invoke(request.error);
+                    }
+                }
```
D4rWiNSS: Maybe we should check the error code to filter some of the more common uses casess like 404 for example.

I would recommend to retry in all cases that the server had an error ( codes 500-600)

D4rWiNSS: What I mean is , there is no point in retry a call that the server has gives us a 404, if the server didn't found the URL, it doesn't make sense to retry again, since our catalyst are full at the moment, we can increase the load with the asset bundles for example

D4rWiNSS (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestBase.cs
sandrade-dcl: Good point Adri! I have added this check to avoid making uneeded retries. Thanks!

sandrade-dcl:
@pbosio @BrianAmadori I have already applied the refactor that we were discussing yesterday. Could you please take a look to confirm that everything is just like we agreed? Thanks!

pbosio (approved):
Love it ❤️

pravusjif (approved):
Looks good, nice job, I want to see it in action in the next PR

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs
AjimenezDCL: We are doing the same with the `QuestController`, so I agree as well that temporarily we are using this approach to be ported into environment in the future.

sandrade-dcl:
Finally I have implemented a custom `WebRequestAsyncOperation` to be returned by our request system, instead of using the `UnityWebRequestAsyncOperation` provided by Unity. I realize that I will need it in the next PR.

BrianAmadori (approved):
Can't wait to see this in action!

PR #2250, opened 2021-03-29
fix: Handling error when fetching parcels with access

pablitar:
So it is not a critical error and it doesn't crash


nchamo (approved):
Great work!

BrianAmadori (approved):
❤️

PR #2255, opened 2021-03-31
fix: null check for currentShape in SceneBoundariesChecker

AjimenezDCL:
With the addition of TextMeshPro meshes to the SceneBoundariesChecker we were missing an edge case where an entity doesn't have a `currentShape` set but it triggers the SceneBoundariesChecker due to the renderers.


pbosio (approved):

sandrade-dcl (approved):
Tested and it is fixed now, thanks Alex!

pravusjif (approved):

PR #2256, opened 2021-03-31
chore: Replace the current UnityWebRequest calls with our new WebRequest (with retries system)

sandrade-dcl:
This is the second step for the PR: https://github.com/decentraland/explorer/pull/2248

Once we have implemented our custom WebRequest system, we're going to use it in the entire project, overriding the current Unity request calls. With this we will gain the retry system in all our requests.

### In addition: New improvements added to our WebRequest system
- New method `SetHash(Hash128 hash)` added to the `WebRequestAssetBundle` class in order to let the user configure the AB requests using a hash.
- Improvements on the web requests disposing.
- Add the new property `isSucceded` to the `WebRequestAsyncOperation` class that will tell us if a request was successfully finished or not.
- Add the new optional param `downloadHandler` to the `WebRequestController.i.Get(...)` signature that let the user to override the download handler used for the request.
- Move all the `UnityWebRequest` extension methods from `Utils` to its own `WebRequestExtensions` class.

### Testing
- Link: https://explorer.decentraland.zone/branch/chore/Replace-the-current-UnityWebRequest-calls-with-our-new-WebRequest/index.html?ENV=org

### Considerations
Please note that the requests will only be retried if the they are failed by server errors (responseCode between 500 and 600). For example, if a request fails by a 404 error, it will not be retried.


unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/Materials/Frame_Wood.mat
```diff
@@ -22,7 +22,7 @@ Material:
   m_PrefabAsset: {fileID: 0}
   m_Name: Frame_Wood
   m_Shader: {fileID: 4800000, guid: 933532a4fcc9baf4fa0491de14d08ed7, type: 3}
-  m_ShaderKeywords: _SPECULAR_SETUP
+  m_ShaderKeywords: 
   m_LightmapFlags: 4
   m_EnableInstancingVariants: 0
   m_DoubleSidedGI: 0
```
pbosio: Is there any reason to change a material in this PR or it's mistake?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
```diff
@@ -83,15 +79,11 @@ protected override void OnCancelLoading()
...
-        }
+        protected override void OnAfterLoadOrReuse() { }
 
-        protected override void OnBeforeLoadOrReuse()
-        {
-        }
+        protected override void OnBeforeLoadOrReuse() { }
 
-        private UnityWebRequestAsyncOperation asyncOp;
+        private WebRequestAsyncOperation asyncOp;
```
pbosio: can we move this on top of the class? not a big fan of leaving fields and/or properties between methods definition

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Texture/AssetPromise_Texture.cs
```diff
@@ -64,29 +54,33 @@ protected override void OnLoad(Action OnSuccess, Action OnFail)
...
+                    OnFail: (errorMsg) =>
                     {
                         OnFail?.Invoke();
-                    }
-                    webRequest?.Dispose();
-                    webRequest = null;
-                };
+                    });
+
+                webRequestOp.disposeOnCompleted = true;
```
pbosio: Just out of curiosity. You will know better. Maybe the default for `disposeOnCompleted` should be `true`?
Do you feel that you needed to add it in more or less scripts while creating this PR?

pbosio (approved):
Tested. I found no problems with wearables or picture frames, everything looks like it's working fine! Great!

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Texture/AssetPromise_Texture.cs
D4rWiNSS: I agree with Pato, at first sight, it seems like `diposeOnComplete `should be `true `by default

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestAssetBundle.cs
```diff
@@ -1,12 +1,40 @@
+using UnityEngine;
 using UnityEngine.Networking;
 
 namespace DCL
 {
+    /// <summary>
+    /// Our custom implementation of the UnityWebRequestMultimedia (Audio Clip).
```
D4rWiNSS: This is for Audio Clip? I think this is a typo, it should be asset bundle right?

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea_Internal/Request.cs
```diff
@@ -141,61 +141,70 @@ IEnumerator Fetch(float delayRequest)
...
+                        SplitGroup();
+                    }
+
+                    asyncOp.Dispose();
+                }
+                else
+                {
+                    shouldRetry = retryCount < REQUEST_RETRIES;
+                    if (!shouldRetry)
+                    {
```
D4rWiNSS, line 187: now this is handle automatically by the web request right? This is doing double retry tries

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea_Internal/Request.cs
```diff
@@ -269,40 +278,46 @@ IEnumerator Fetch()
...
+                {
+                    shouldRetry = false;
+                    OnSuccess?.Invoke(response);
+                    requestOp.Dispose();
+                }
+                else
+                {
+                    shouldRetry = retryCount < REQUEST_RETRIES;
+                    if (!shouldRetry)
                     {
```
D4rWiNSS, line 308: Same here

unity-client/Assets/UnityGLTF/Scripts/Loader/WebRequestLoader.cs
```diff
@@ -50,10 +48,7 @@ public IEnumerator LoadStream(string filePath)
             yield return CreateHTTPRequest(_rootURI, filePath);
         }
 
-        public void LoadStreamSync(string jsonFilePath)
-        {
-            throw new NotImplementedException();
-        }
+        public void LoadStreamSync(string jsonFilePath) { throw new NotImplementedException(); }
```
D4rWiNSS, line 57: I know that you didn't make this change but do you know if this Is suppose to throw not implemented?

D4rWiNSS (approved):
Good job! Congratulation! We have now a more robust app!

unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/Materials/Frame_Wood.mat
sandrade-dcl: Mmmm I'm not sure why this change has been added in this PR. I have reverted it in https://github.com/decentraland/explorer/pull/2256/commits/d3c8124571aa84bde8f76d99d221567aadcab74c. Thanks!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
sandrade-dcl: Done! https://github.com/decentraland/explorer/pull/2256/commits/c46adada51c9713be23a6455d4bb3052a9da9d63

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Texture/AssetPromise_Texture.cs
sandrade-dcl: After take a look, I agree with you guys that it has more sense to set `disposeOnCompleted` as true as default. To ease the use of this flag, I have added it (as true by default) in the own GET signature. In this way, if we would want to set it as `false` for any request, we can do it directly on the call instead having to do a second step.

Chansges done in: https://github.com/decentraland/explorer/pull/2256/commits/6f3cfb7540359bffed48ec8b4a998205d7cb1081

Thanks for the feedback!

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea_Internal/Request.cs
sandrade-dcl: No, in this specific case, as I commnent above of this line, we are letting this class to implement the retries.
https://github.com/decentraland/explorer/blob/cbf1c53fe9f2b43bf96d06c84c6b23f18e3dda1b/unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea_Internal/Request.cs#L149

As it is implementing a very specific retries system (including delays), we use our custom WebRequest system without retries (`requestAttemps: 1`) and let the current code to apply the retries.

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea_Internal/Request.cs
sandrade-dcl: Same response as before :)

unity-client/Assets/UnityGLTF/Scripts/Loader/WebRequestLoader.cs
sandrade-dcl: The truth is that I don't know.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestAssetBundle.cs
sandrade-dcl: You're right! Thanks! Fixed in https://github.com/decentraland/explorer/pull/2256/commits/142347c71fa9f48bb2c13e48e06a209dca2f86d1

unity-client/Assets/ABConverter/VisualTests.cs
```diff
@@ -211,49 +211,49 @@ public static GameObject[] LoadAndInstantiateAllAssetBundles()
             foreach (var hash in dependencyAbs)
             {
                 string path = abPath + hash;
-                var req = UnityWebRequestAssetBundle.GetAssetBundle(path);
 
                 if (SystemInfo.operatingSystemFamily == OperatingSystemFamily.MacOSX)
-                    req.url = req.url.Replace("http://localhost", "file:///");
+                    path = path.Replace("http://localhost", "file:///");
 
-                req.SendWebRequest();
```
BrianAmadori: This is inside `ABConverter`. `ABConverter` has its own Environment with its own `IWebRequest` implementation with the editor mode concerns. We shouldn't be referencing the `WebRequestController` here. 

This is because we aim to move the `ABConverter` code in a separate repo in the future, so we should revert the changes of this file.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioClip.cs
```diff
@@ -95,8 +87,11 @@ IEnumerator TryToLoad()
                 Model model = (Model) this.model;
                 if (scene.contentProvider.HasContentsUrl(model.url))
                 {
-                    yield return Utils.FetchAudioClip(scene.contentProvider.GetContentsUrl(model.url),
-                        Utils.GetAudioTypeFromUrlName(model.url), OnComplete, OnFail);
+                    Utils.FetchAudioClip(
```
BrianAmadori: Please keep the yield logic to minimize regression chances by the logic change.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
```diff
@@ -164,12 +147,11 @@ protected IEnumerator LoadAssetBundleWithDeps(string baseUrl, string hash, Actio
                 yield return promise;
             }
 
-            AssetBundle assetBundle = DownloadHandlerAssetBundle.GetContent(assetBundleRequest);
+            AssetBundle assetBundle = DownloadHandlerAssetBundle.GetContent(asyncOp.webRequest);
+            asyncOp.Dispose();
 
             if (assetBundle == null || asset == null)
             {
-                assetBundleRequest.Abort();
```
BrianAmadori, line 171: Why this was removed and the dispose moved outside the if statement? Are you 100% that this can't introduce regressions?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/DependencyMapLoadHelper.cs
```diff
@@ -62,31 +62,28 @@ public static IEnumerator GetDepMap(string baseUrl, string hash)
...
 
-        using (UnityWebRequest depmapRequest = UnityWebRequest.Get(url))
-        {
-            downloadingDepmap.Add(hash);
-
-            yield return depmapRequest.SendWebRequest();
-
-            if (!depmapRequest.WebRequestSucceded())
+        downloadingDepmap.Add(hash);
+        WebRequestController.i.Get(
```
BrianAmadori: This request was yielded for a reason. If we don't yield, the method will return before the request is finished and this may be unexpected from the AssetPromise.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Gif/GifProcessor/GifProcessor.cs
```diff
@@ -48,9 +49,9 @@ public void DisposeGif()
         {
             DCL.GIFProcessingBridge.i.DeleteGIF(url);
         }
-        else if (!(webRequest is null))
+        else if (webRequestOp != null)
```
BrianAmadori, line 52: The previous logic nulls the webRequest to make sure it isn't running in this statement. This is an ongoing pattern in all the project, so please be extra careful here. We should keep the logic and null the webRequestOp below or check for !IsDisposed().

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
```diff
@@ -61,11 +60,9 @@ protected override void OnCancelLoading()
                 loadCoroutine = null;
             }
 
-            if (assetBundleRequest != null && !assetBundleRequest.isDone)
+            if (asyncOp != null)
```
BrianAmadori, line 63: This statement should check if the request is in progress. If we check for null it may not work as expected. (Check out the comment above about nullity and running requests)

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea_Internal/Request.cs
```diff
@@ -269,40 +278,46 @@ IEnumerator Fetch()
...
-                            OnError?.Invoke("Error " + request.downloadHandler.text);
-                        }
-                        else
-                        {
-                            retryCount++;
-                            if (OpenSeaRequestController.VERBOSE)
-                                Debug.Log($"RequestGroup: Request retrying {url}");
-                            yield return new WaitForSeconds(GetRetryDelay());
-                        }
+                        retryCount++;
```
BrianAmadori, line 314: We should remove the retryCount to favor our retry logic here.

unity-client/Assets/UnityGLTF/Scripts/Loader/WebRequestLoader.cs
```diff
@@ -66,53 +60,31 @@ private IEnumerator CreateHTTPRequest(string rootUri, string httpRequestPath)
...
-                www = new UnityWebRequest(finalUrl, "GET", new DownloadHandlerBuffer(), null);
 
-                www.timeout = 5000;
-#if UNITY_2017_2_OR_NEWER
-                yield return www.SendWebRequest();
-#else
-            yield return www.Send();
-#endif
-                if ((int)www.responseCode >= 400)
+            return DCL.WebRequestController.i.Get(
```
BrianAmadori: `UnityGLTF` should stand by itself because we will have to migrate it to an external repo in some point, so please let's use dependency injection to inject this implementation instead of using `WebRequestController.i`.

unity-client/Assets/UnityGLTF/Scripts/Loader/WebRequestLoader.cs
```diff
@@ -66,53 +60,31 @@ private IEnumerator CreateHTTPRequest(string rootUri, string httpRequestPath)
...
-                    //Note (Adrian): 500 to 600 codes are reserved codes to Server error responses, so if we have an error from server, we retry
-                    if (retryCont < RETRY_AMOUNTS && www.responseCode >= 500 &&  www.responseCode < 600)
+                    if (webRequestResult.downloadedBytes > int.MaxValue)
                     {
-                        retryCont++;
-                        retry = true;
+                        Debug.LogError("Stream is too big for a byte array");
                     }
-                    else
+                    else if (webRequestResult.downloadHandler.data != null)
```
BrianAmadori: This creates a whole copy of all the data of the request, we shouldn't call it twice. Please check the comment below (in the code).

BrianAmadori (changes requested):
After looking at this in depth, I see quite a few issues with this migration we should address. Left some requests.

unity-client/Assets/ABConverter/VisualTests.cs
sandrade-dcl: Ups, I didn't know about it. You're right it is better not to touch anything related to `ABConverter` if we plan to move it out of the repo in the future. Thanks for the warning! Done in https://github.com/decentraland/explorer/pull/2256/commits/c30f9b5faffb9a87ccc94aef77c601a80dabc8d1

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioClip.cs
sandrade-dcl: Reviewing the DCLAudioClip flow, I think the yield logic was uneeded in this case. Also, after several tests with audio, I didn't notice any problem. Anyway I will add it back (just in case). Done in https://github.com/decentraland/explorer/pull/2256/commits/d8a7c637eed3918d66ef0874dbd323b3d5cda549

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
sandrade-dcl: If you notice, right before this conditional we are getting the downloaded content of the request:

`AssetBundle assetBundle = DownloadHandlerAssetBundle.GetContent(asyncOp.webRequest);`

As the request operation will not be used anymore from this point, it is the perfect place where dispose it. The old code not only was not making a dispose of the request, but was aborting a request that had already finished... and that doesn't make sense.

The current code has been tested and for me is more performant in terms of memory felease.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/DependencyMapLoadHelper.cs
sandrade-dcl: My fault, I didn't notice when I applied the change of method. Thanks! Done in https://github.com/decentraland/explorer/pull/2256/commits/d24a7ebe1a73c92facab4dc8258aeba4f3a7368c

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Gif/GifProcessor/GifProcessor.cs
sandrade-dcl: Notice that the `Dispose()` we are using here belongs to our custom `WebRequestAsyncOperation` and it is already nulling the request:

image

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
sandrade-dcl: Following the previous explanation, our new `Dispose()` is already checking for the `isDone` param to avoid the abortation or not:

image

Apart of that, I think the old code was not disposing the request well. If the request was running, then it was aborted and disposed correctly, BUT if the request was already finished, it was not aborted it (OK, it is not needed) but it didn't dispose either (and it should be disposed always in any cancelation).

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea_Internal/Request.cs
sandrade-dcl: I left it that way on purpose. After a conversation with @pbosio about this specific case, we decided to keep the specific "retries" implementation there. Please take a loook to my comment here: https://github.com/decentraland/explorer/blob/f8bbae8c344d097818cbcded8b50dc3a201a8ed2/unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea_Internal/Request.cs#L147

pbosio: The reason is that requests for OpenSea handle a very specific retry behavior added by @AjimenezDCL We shouldn't change that. (after a request has failed the request is splitted in half and 2 request are send and so on and on...)

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
BrianAmadori: Amazing.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Gif/GifProcessor/GifProcessor.cs
BrianAmadori: The `if` checks against the `webRequestOp` to check if the request is in progress. If that's true then the if should be `if (webRequestOp?.webRequest != null)` or something like that. Maybe i'm missing something else here?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/AssetBundles/AB/AssetPromise_AB.cs
BrianAmadori: Cool! On a closer look, if the request is running `asyncOp` should be non-null. Going to resolve this one.

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/NFTHelper/NFTMarkets/OpenSea_Internal/Request.cs
BrianAmadori: Ok. Resolved.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioClip.cs
BrianAmadori: Ty!

unity-client/Assets/ABConverter/VisualTests.cs
BrianAmadori: Ty!

unity-client/Assets/UnityGLTF/Scripts/Loader/WebRequestLoader.cs
BrianAmadori: To do this, we should pass the interface implementation through `GLTFSceneImporter`. This is injected via `GLTFComponent` or `GLTFImporter` (for editor mode). 

I think the editor code path doesn't use the `WebRequestLoader` so we may escape the injection of that path, but we'd have to make sure.

BrianAmadori:
Please make sure to resolve the remaining discussions before merging. Awesome work so far!

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Gif/GifProcessor/GifProcessor.cs
sandrade-dcl: If we thing about this specific case, what it should happen if we reach this `if` is that, whether the request is in progress or not, we should dispose it (and abort it before if it is in progress in that moment), and our new `Dispose()` is already doing that:

image
image

Maybe in the old code, using directly the `UnityWebRequest`, we always had to check this kind of things before an abort or a dispose, but now we have a `WebRequestAsyncOperation` that acts as a "container" of the request and we delegate all these kind of checking responsabilities to its internal logic (in functions like `Abort()`, `Dispose()`, etc.). That is why we only make the null-check of the "container" now, the responsability of the rest of checkings has been moved to the own request operation. This allows us to avoid always having to think about what checks to make before execute an abort or a dispose, we simply order the specific operation to `WebRequestAsyncOperation` and he will do the rest.

Don't worry, this disposing case is under control.

unity-client/Assets/UnityGLTF/Scripts/Loader/WebRequestLoader.cs
sandrade-dcl: @BrianAmadori I have done the changes but I would like to review it with you before commit. Can we get a quick talk?

unity-client/Assets/UnityGLTF/Scripts/Loader/WebRequestLoader.cs
sandrade-dcl: You're right! I will adapt it, thanks!

unity-client/Assets/UnityGLTF/Scripts/Loader/WebRequestLoader.cs
sandrade-dcl: Done in https://github.com/decentraland/explorer/pull/2256/commits/8d8a0aa9218abdb7456ad945675d5c1a30215979

unity-client/Assets/UnityGLTF/Scripts/Loader/WebRequestLoader.cs
sandrade-dcl: Done in https://github.com/decentraland/explorer/pull/2256/commits/810b7caaa2a8c27b21c55bed7ca5c6768dd00aad

BrianAmadori (approved):
Everything has been sorted out! Thanks!

PR #2258, opened 2021-04-01
chore: update base avatars collection id

nchamo:
We are doing 2 changes in this PR:
* We are now moving collection ids to urns. So what used to be the id `base-avatars` will now be `urn:decentraland:off-chain:base-avatars`.
* We are updating the `urn-resolver` library, and now types that used to be `blockchain-collection-v1` are `blockchain-collection-v1-asset`

### DIFF
--- a/kernel/package-lock.json
+++ b/kernel/package-lock.json
@@ -435,9 +435,9 @@
       }
     },
     "@dcl/urn-resolver": {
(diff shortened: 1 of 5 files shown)


sandrade-dcl (approved):
LGTM! :)

PR #2259, opened 2021-04-02
feat: New event when a user enters or leaves the scene

kuruk-mm:
# What?
Since https://github.com/decentraland/explorer/pull/1768, there is now an event that is emitted every time the user changes scene. So in this PR we hook that event to the SDK to have access to the event of enter or left the scene.

And also I added two new scenes to the test-scenes, to test this. Those scenes have a floor in gray, and when you enter those the floor turns red, and when you left turns gray again.

Video:
https://user-images.githubusercontent.com/12563266/113425077-a01cde00-93a7-11eb-91ff-2134576bb145.mp4

When we left one scene, to a parcel without scene, the event is not triggered, but this only happens with the test-scenes, because in Decentraland World there are not parcels without scenes assigned.
I chated with @nchamo and this is the desired behavior.

# Why?
Allow the user use the SDK to get the event of enter/left scenes.

# How to test
Follow this guide (`Running Test Scenes` section):
https://github.com/decentraland/explorer/blob/master/docs/how-to-test-parcels-and-preview-scenes.md

And go to position 0,11 (quick link: http://localhost:3000/?DEBUG_MODE&position=0,11)

### DIFF
--- a/kernel/packages/decentraland-ecs/src/decentraland/Types.ts
+++ b/kernel/packages/decentraland-ecs/src/decentraland/Types.ts
@@ -1,5 +1,4 @@
 import { ReadOnlyVector3, ReadOnlyQuaternion, ReadOnlyColor4 } from './math'
-import { RaycastResponse } from './Events'
 
 /** @public */
 export type ModuleDescriptor = {
@@ -129,6 +128,13 @@ export type GlobalInputEventResult = InputEventResult & {
   type: InputEventType
 }
 
+/** @public */
+export type RaycastResponsePayload<T> = {
+  queryId: string
+  queryType: string
+  payload: T
+}
+
 /**
  * @public
  */
@@ -180,7 +186,7 @@ export interface IEvents {
   /**
    * `raycastResponse` is triggered in response to a raycast query
    */
-  raycastResponse: RaycastResponse<any>
+  raycastResponse: RaycastResponsePayload<any>
 
   /**
    * `chatMessage` is triggered when the user sends a message through chat entity.
--- a/kernel/packages/decentraland-ecs/src/decentraland/Types.ts
+++ b/kernel/packages/decentraland-ecs/src/decentraland/Types.ts
@@ -271,6 +271,20 @@ export interface IEvents {
   }
 
   /**
+   * This event gets triggered when the user enters the scene
+   */
+  onEnterScene: {
+    userId: string
+  }
+
+  /**
+   * This event gets triggered when the user leaves the scene
+   */
+  onLeftScene: {
+    userId: string
+  }
+
+  /**
    * This event gets triggered after receiving a comms message.
    */
   comms: {
--- a/kernel/public/test-scenes/0.11.subscribeToEnterLeaveScene1/game.ts
+++ b/kernel/public/test-scenes/0.11.subscribeToEnterLeaveScene1/game.ts
@@ -1,0 +1,29 @@
+import { log, engine, Entity, BoxShape, Material, Color3, Transform, Vector3, onEnterScene, onLeftScene } from 'decentraland-ecs/src'
+
+//Create entity and assign shape
+const box = new Entity()
+box.addComponent(new BoxShape())
+
+//Create material and configure its fields
+const material = new Material()
+material.albedoColor = Color3.Gray()
+
+//Assign the material to the entity
+box.addComponent(material)
+
+box.addComponent(new Transform({
+  position: new Vector3(8, 0, 8),
+  scale: new Vector3(16, 0.1, 16),
+}))
+
+engine.addEntity(box);
+
+onEnterScene.add(({ userId }) => {
+  material.albedoColor = Color3.Red()
+  log("onEnterScene: ", userId)
+})
+
+onLeftScene.add(({ userId }) => {
+  material.albedoColor = Color3.Gray()
+  log("onLeftScene: ", userId)
+})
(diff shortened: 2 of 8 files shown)


kernel/public/test-scenes/0.11.subscribeToEnterLeaveScene1/game.ts
```diff
@@ -0,0 +1,31 @@
...
+
+box.addComponent(new Transform({
+  position: new Vector3(8, 0, 8),
+  scale: new Vector3(16, 0.1, 16),
+}))
+
+engine.addEntity(box);
+
+
+declare var dcl: DecentralandInterface
```
nchamo: We try not to expose the `DecentralandInterface` in the SDK, so we should try to create an abstraction similar to the [camera](https://github.com/decentraland/explorer/blob/397acd24d5bdfd764a2d835ec55251332ca40110/kernel/packages/decentraland-ecs/src/decentraland/Camera.ts#L9).

We could create a `Scene` class with some events that could look like this:
```
onEnterScene(listener: ({ userId }) => void ) {
    // subscribe to `onEnterScene` if it hasn't subscribed yet
    dcl.onEvent((evt: any) => {
        if (evt.type === 'onEnterScene') {
            listener(evt.data)
        }
    }) 
}
```
@menduz , does the subscription model make sense like this?
 
And I have one question. If the user starts inside the scene, is the event triggered?

nchamo (commented):
Amazing work!

Just one comment about how to better expose this to the users

kernel/public/test-scenes/0.11.subscribeToEnterLeaveScene1/game.ts
menduz: I think adding observables in `kernel/packages/decentraland-ecs/src/decentraland/Events.ts` would be a good course of action to start decoupling the `dcl.onEvent` from many functions scattered over there. It also makes the code more testable.

```ts
// kernel/packages/decentraland-ecs/src/decentraland/Events.ts
import { Observable } from '../ecs/Observable'

...
...
...

/**
 * @internal
 * This function generates a callback that is passed to the Observable
 * constructor to subscribe to the events of the DecentralandInterface
 */
function createSubscriber(eventName: keyof IEvents) {
  return () => {
    if (internalDcl) {
      internalDcl.subscribe(eventName)
    }
  }
}

let internalDcl: DecentralandInterface | void

/**
 * @internal
 * This function adds _one_ listener to the onEvent event of dcl interface.
 * Leveraging a switch to route events to the Observable handlers.
 */
export function _initEventObservables(dcl: DecentralandInterface) {
  // store internal reference to dcl, it is going to be used to subscribe to the events
  internalDcl = dcl

  if (internalDcl) {
    internalDcl.onEvent((event) => {
      switch (event.type) {
        case 'onEnterScene': {
          onEnterScene.notifyObservers(event.data)
          return
        }
      }
    })
  }
}



/**
 * These events are triggered after your character enters the scene.
 * @public
 */
export const onEnterScene = new Observable<IEvents['onEnterScene']>(createSubscriber('onEnterScene'))

```

and then call `_initEventObservables(dcl)` from `kernel/packages/decentraland-ecs/src/index.ts`

```diff
/** @internal */
declare let dcl: DecentralandInterface | void
if (typeof dcl !== 'undefined') {
  engine.addSystem(new DecentralandSynchronizationSystem(dcl), Infinity)
+ _initEventObservables(dcl)
}

Comment (kuruk-mm) [reply kernel/public/test-scenes/0.11.subscribeToEnterLeaveScene1/game.ts]: About your question, yes, the event is triggered when you start inside the scene, or when you teleport into the scene.

**Review by menduz — CHANGES_REQUESTED** [review]
Very cool, I think we can work on a better long term solution for events in the SDK based on @nchamo and my comments.

Does this work properly for multi-parcel scenes?

**Comment by kuruk-mm** [comment]
> Very cool, I think we can work on a better long term solution for events in the SDK based on @nchamo and my comments.
> 
> Does this work properly for multi-parcel scenes?

It should work in multi-parcel scenes so I will expand some parcels in one of the scenes to test that.

So with what you said, this looks good for SDK user code?

```
onEnterScene((userId) => {
    log("Enter Scene")
})
onLeftScene((userId) => {
    log("Left Scene")
})
```
This only will trigger to the local user, but `userId` parameter is to extend the event in the future if we want to add a trigger when other users enter/left the scene (@nchamo idea)

**Comment by menduz** [comment]
If you go with the Observable approach, the event will be typed in the `IEvents` type.

The usage would be similar to this:

```ts
onEnterScene.add(data => {
  log("Enter Scene", data.userId)
})

// or this one , with sugar syntax

onEnterScene.add(( {userId} ) => {
  log("Enter Scene", userId)
})
```

The reason to not do `onLeftScene(callback)` is to not diverge on the implementation of the event listeners and move towards using Observable, since it allows not only registering typed callbacks. It also keeps track of the callbacks that were registered, allows to dispatch once, to execute sequentially, to dispatch and await promises, and to clear the observers. It is also tested and embeded in the SDK, removing duplication of code

kernel/packages/decentraland-ecs/src/decentraland/Types.ts
```diff
@@ -270,6 +270,16 @@ export interface IEvents {
     entityId: string
   }
 
+  /**
+   * This event gets triggered when the user enters the scene
+   */
+  onEnterScene: {}
+
+  /**
+   * This event gets triggered when the user lefts the scene
```
nearnshaw: when the user *leaves the scene

menduz (approved):

nchamo (approved):
Amazing work!

kuruk-mm:
Done, I implemented @menduz code.

Now one of the new test-scenes owns two parcels and it's working.

And thanks @nchamo for your great help :smile:

PR #2263, opened 2021-04-05
chore: update catalyst client lib

nchamo:
We are updating the dcl-catalyst-client library, and we are adding a "null check" for improved reliability


kernel/packages/shared/catalogs/sagas.ts
```diff
@@ -171,8 +177,10 @@ function* fetchWearablesV2(filters: WearablesRequestFilters) {
         client
       )
       for (const { amount, definition } of ownedWearables) {
-        for (let i = 0; i < amount; i++) {
-          result.push(definition)
+        if (definition) {
```
nchamo, line 174: There are some corner cases where the definition might be null, so we are adding a check here

agusaldasoro (approved):

PR #2264, opened 2021-04-05
feat: add ssao settings

BrianAmadori:
## What? 

* A new `Ultra` profile was added on top of the `High` profile.
* All profiles were modified.
* Add new setting spinbox for the SSAO feature.
* Fixed potential bug related to float point precision that could cause the preset spinbox to skip levels and go to `custom` randomly.
* Upped the "Match 720p" resolution a bit to be more decent.


unity-client/Assets/Scripts/MainScripts/DCL/Controllers/HUD/SettingsPanelHUD/Scripts/ControlsModule/SliderControlModel.cs
```diff
@@ -15,9 +15,6 @@ public class SliderControlModel : SettingsControlModel
         [Tooltip("Max allowed value for the slider.")]
         public float sliderMaxValue;
 
-        [Tooltip("True if the slider values will be integers.")]
-        public bool sliderWholeNumbers;
```
sandrade-dcl, line 19: A consequence of remove this param is that the sliders that had using it, now will show their values as decimal, like en these examples:

image

We should fix it.

sandrade-dcl (changes requested):

pravusjif (approved):
I'm approving the PR taking into account that Santi's reported visual bug + objects culling setting will be fixed (as the culling is really needed for performance, affecting mostly the unneeded shadows rendering as you already know).

SSAO setting worked flawlessly for me.

sandrade-dcl (approved):

PR #2271, opened 2021-04-08
fix: disable world camera on navmap open

BrianAmadori:
World camera was active when the navmap was open. Disabling it should improve the navmap performance by a mile.


sandrade-dcl (approved):
Reviewed and tested. Good little improvement! :)

pravusjif (approved):
Beautiful

PR #2274, opened 2021-04-08
fix: better error handling for wearable requests

BrianAmadori:
## Why? 
Right now, when any owned wearable request fails on initial load, we get a perpetual loading screen. This PR makes the loading fail right away.

## What? 
- If any owned wearable request fails, we throw a fatal error with the endpoint info
- For other wearable failures we just output a warning with the endpoint info

### DIFF
--- a/kernel/packages/shared/catalogs/sagas.ts
+++ b/kernel/packages/shared/catalogs/sagas.ts
@@ -38,6 +38,8 @@ import { FeatureFlags } from 'shared/meta/types'
 import { CatalystClient, OwnedWearablesWithDefinition } from 'dcl-catalyst-client'
 import { parseUrn } from '@dcl/urn-resolver'
 import { getCatalystServer, getFetchContentServer } from 'shared/dao/selectors'
+import { ReportFatalError } from 'shared/loading/ReportFatalError'
+import { CONTENT_SERVER_DOWN } from 'shared/loading/types'
 
 declare const globalThis: Window & RendererInterfaces & StoreContainer
 export const BASE_AVATARS_COLLECTION_ID = 'urn:decentraland:off-chain:base-avatars'
@@ -178,9 +180,19 @@ function* fetchWearablesV2(filters: WearablesRequestFilters) {
         }
       }
     }
+
+    if (result.length === 0) {
+      ReportFatalError(CONTENT_SERVER_DOWN)
+      throw new Error(`[FATAL]: Can't fetch owned wearables!\nCatalyst:${catalystUrl}`)
+    }
+
   } else {
     const wearables = yield call(fetchWearablesByFilters, filters, client)
     result.push(...wearables)
+
+    if (result.length === 0) {
+      defaultLogger.error(`[Caution]: Couldn't resolve non-owned wearables\nCatalyst:${catalystUrl}`)
+    }
   }
 
   const v1Wearables = yield call(mapV2WearablesIntoV1, result)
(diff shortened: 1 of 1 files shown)


kernel/packages/shared/catalogs/sagas.ts
```diff
@@ -178,9 +180,19 @@ function* fetchWearablesV2(filters: WearablesRequestFilters) {
         }
       }
     }
+
+    if (result.length === 0) {
+      ReportFatalError(CONTENT_SERVER_DOWN)
+      throw new Error(`[FATAL]: Can't fetch owned wearables!\nCatalyst:${catalystUrl}`)
```
nchamo, line 186: I don't think we should fail at any point. We should let Unity decide what to do if something fails. Specially when there is no need to ask for owned wearables when the client starts (check https://github.com/decentraland/unity-renderer/issues/214). 

Adding more logs/warnings sounds great, but throwing a fatal error should be up to Unity. Because maybe there is a way to fail gracefully

nchamo (changes requested):
We shouldn't be throwing fatal errors in cases where we can recover gracefully

kernel/packages/shared/catalogs/sagas.ts
BrianAmadori: The thing is that "owned wearables" include the wearables I'm wearing, and those wearables are required for the loading start. If we don't fail here the loading will get stuck forever.

Until now, Unity has no way to make the fatal error screen appear (we have no pipeline for that), and furthermore, when this fails the message that should indicate Unity that the wearable couldn't be resolved never gets there.

I propose this as a temporary solution until we find the way to make Unity throw the fatal error, as right now the failing flow at this stage is completely broken. The fatal error is preferred to the forever loading screen from an UX standpoint.

nchamo: But the thing is that "owned wearables" should not be required for the loading to start. That's something that Unity is doing and can be changed. Unity can ask for wearables that the user is currently wearing by ids, and that wouldn't fail if thegraph is down. 

The "owned wearables" are only needed when the avatar editor is opened, or when you click on another avatar (and you want to see their owned wearables). That is what https://github.com/decentraland/unity-renderer/issues/214 is all about. 

That's why I don't want to show a fatal error, because the "owned wearables" request to Kernel shouldn't even be a part of the loading process. It isn't necessary, and it will force the explorer to shut down if thegraph is down.

sandrade-dcl: @nchamo a thing that I don't completely understand is: if thegraph is down and my avatar has some "owned" wearable, why the fetch by id would not fail? Doesn't the "fecth by id" flow go through thegraph?

nchamo: No, it doesn't. If you fetch by id, you get the wearable's definition, which is stored on the content server. The blockchain (thegraph) is only queried for information about ownership, not about definitions.

sandrade-dcl: Ok, thanks for the clarification @nchamo. Then effectively to defer the "owned wearables" request from the UserProfile (at the beggining) to the AvatarEditor would fix this specific "thegraph" problem.

Other thing to discuss, maybe out of this context, is that we should start to handle different error situations (I guess from both sides Kernel and Unity) in order to avoid to make the initial loading freezed and give the user some real feedback of wath is happening.

So @BrianAmadori what should we do with this PR? Do you think it is worth firstly prioritize the task https://github.com/decentraland/unity-renderer/issues/214 and plan the implementation of other kind of error handlings for later?

BrianAmadori: In any case, this PR should add a temporary solution improving the feedback in the future cases that for some reason the wearable server returns empty wearable arrays. I don't see the reason for not doing this. @nchamo 

We can merge this for this release and then improve the flow later.

sandrade-dcl: I though the problem that we were discussing in this PR was the "thegraph" one. But obviously if we are currently having problems with the wearables server, then the problem is different! In this case I agree with @BrianAmadori in this temporal solution.

nchamo: Having 0 owned wearables is not only valid, is the most common case. This PR, as it is, will make the explorer only load for people who own wearables. 

It's Unity who decides which wearables are necessary and which aren't. If you want a temporary solution, let's add to the `BrowserInterface` a way to show a fatal error, and you can call that instead of freezing. 

If you don't want to do that either, please check what wearables Unity considers as required and copy that here

sandrade-dcl: That's true, an user can receive 0 owned wearables and it shouldn't be an error.

BrianAmadori: Ok, good catch (we should have started from there). I'm going to rethink the approach.

PR #2275, opened 2021-04-08
fix: support prod collections in WITH_COLLECTIONS

nchamo:
As of now, the `WITH_COLLECTIONS` query param only worked with ropsten and mumbai collections. We are now adding the possibility to also use mainnet/matic collections. This will help in testing our code with the wearables our users use

### DIFF
--- a/kernel/packages/shared/catalogs/sagas.ts
+++ b/kernel/packages/shared/catalogs/sagas.ts
@@ -162,10 +162,15 @@ function* fetchWearablesV2(filters: WearablesRequestFilters) {
   if (filters.ownedByUser) {
     if (WITH_FIXED_COLLECTIONS) {
       const collectionIds = WITH_FIXED_COLLECTIONS.split(',')
-      const orgClient = new CatalystClient('peer.decentraland.org', 'EXPLORER')
+      const orgCatalyst = 'https://peer.decentraland.org'
+      const orgClient = new CatalystClient(orgCatalyst, 'EXPLORER')
       const zoneWearables = yield client.fetchWearables({ collectionIds })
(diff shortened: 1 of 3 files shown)


kernel/packages/shared/catalogs/sagas.ts
```diff
@@ -162,8 +162,15 @@ function* fetchWearablesV2(filters: WearablesRequestFilters) {
   if (filters.ownedByUser) {
     if (WITH_FIXED_COLLECTIONS) {
       const collectionIds = WITH_FIXED_COLLECTIONS.split(',')
```
agusaldasoro, line 166: Can we add a comment here as a disclaimer showing that this must be used only in zone for testing purposes?

kernel/packages/shared/catalogs/sagas.ts
```diff
@@ -162,8 +162,15 @@ function* fetchWearablesV2(filters: WearablesRequestFilters) {
...
-      const wearables = yield call(fetchWearablesByFilters, { collectionIds }, client)
-      result.push(...wearables)
+      const orgCatalyst = 'https://peer.decentraland.org'
```
agusaldasoro: Can we use the Catalyst Client library to give any catalyst from prod instead of hardcoding it here?

agusaldasoro (approved):
Great work!

kernel/packages/shared/catalogs/sagas.ts
nchamo: Done!

kernel/packages/shared/catalogs/sagas.ts
nchamo: Done!

PR #2279, opened 2021-04-12
feat: biw add save/load scene state

D4rWiNSS:
# What? 
This PR adds a save/load system for the scenes that are not published.

It also introduces compatibility with the current builder, so you can edit the scenes in both editors.

however,  take into account that right now it only saves the state when you exit from builder in world mode ( in the explorer, in the builder it will work as usual). This will change in the future and Unity will handle the save system

# Why? 
The biw currently don't have a save/load feature, if you don't publish your changes to the content server you will lose your progress 

This implements a way to save your scene in the builder server so you can always have a draft of your land available

This branch also introduces compatibility with the builder, so you can edit a scene in the current builder and in the builder in world and the progress will be saved in both parts

# Things to have into account

For this part to work we need to do a couple of things:

1. Deploy a new version of the builder-api ( on ropsten at least)
2. Refactor Unity part to handle correctly the updates to the kernel 

Right now Unity is not sending some messages in batch to the kernel and it can cause some inconsistencies during the save.

**Right now it only saves when you exit from the builder in world explorer, in the next PR Unity will handle the save with more consistency**


unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadableShape.cs
```diff
@@ -151,7 +151,14 @@ protected virtual void AttachShape(IDCLEntity entity)
             ContentProvider provider = null;
 
             if (!string.IsNullOrEmpty(model.assetId))
+            {
                 provider = AssetCatalogBridge.i.GetContentProviderForAssetIdInSceneObjectCatalog(model.assetId);
+                if (string.IsNullOrEmpty(model.src))
+                {
+                    SceneObject sceneObject = AssetCatalogBridge.i.GetSceneObjectById(model.assetId);
+                    model.src = sceneObject.model;
+                }
```
D4rWiNSS, line 160: This is done in order to take into account the builder compatibility, since the builder already have the assets it doesn't send the src field, so we set it in case that we don't receive it

kernel/packages/scene-system/stateful-scene/BuilderStatefulActor.ts
```diff
@@ -0,0 +1,51 @@
+﻿import { CatalystClient } from 'dcl-catalyst-client'
+import { EntityType } from 'dcl-catalyst-commons'
+import { SceneStateDefinition } from './SceneStateDefinition'
+import { ILand } from 'shared/types'
+import { SceneStateStorageController } from 'shared/apis/SceneStateStorageController/SceneStateStorageController'
+import { deserializeSceneState } from './SceneStateDefinitionSerializer'
+
+declare var window: any
```
nchamo: I think you are not using this here

kernel/packages/scene-system/stateful-scene/SceneStateActor.ts
```diff
@@ -0,0 +1,33 @@
+import { IEngineAPI } from 'shared/apis/EngineAPI'
+import { SceneStateDefinition } from './SceneStateDefinition'
+import { Component, ComponentData, ComponentId, EntityId, StateContainer, StatefulActor } from './types'
+
+export class SceneStateActor extends StatefulActor {
```
nchamo: What's the purpose of the `SceneStateActor`. It looks like it's just a container for `SceneStateDefinition`

kernel/packages/scene-system/stateful.scene.system.ts
```diff
@@ -17,47 +19,69 @@ class StatefulWebWorkerScene extends Script {
...
     // Send the initial state ot the renderer
-    this.sceneState.sendStateTo(this.renderer)
-    this.renderer.sendInitFinished()
+    this.sceneActor.sendStateTo(this.rendererActor)
+
+    this.rendererActor.sendInitFinished()
     this.log('Sent initial load')
 
+    // Listen to scene state events
+    this.ListenToEvents(sceneId)
```
nchamo:
```suggestion
    this.listenToEvents(sceneId)
```

kernel/packages/scene-system/stateful.scene.system.ts
```diff
@@ -17,47 +19,69 @@ class StatefulWebWorkerScene extends Script {
...
+  }
+
+  private ListenToEvents(sceneId: string): void {
```
nchamo:
```suggestion
  private listenToEvents(sceneId: string): void {
```

kernel/packages/scene-system/stateful-scene/types.ts
```diff
@@ -17,28 +19,28 @@ export interface StateContainer {
...
+}
+
 /**
  * An actor that contains the state of the scene, but it can also generate updates to it
  */
 export abstract class StatefulActor implements StateContainer {
+  protected readonly eventSubscriber: EventSubscriber
+
+  constructor(protected readonly engine: IEngineAPI) {
+    this.eventSubscriber = new EventSubscriber(this.engine)
```
nchamo: Why are you adding this to the "parent" class. Isn't the RendererActor the only one who needs this? I'm not seeing it being used anywhere else

kernel/packages/scene-system/stateful-scene/BuilderStatefulActor.ts
```diff
@@ -0,0 +1,51 @@
...
+    protected readonly realmDomain: string,
+    private readonly sceneStorage: SceneStateStorageController
+  ) {}
+
+  async getInititalSceneState(): Promise<SceneStateDefinition> {
+    const sceneState = await this.getContentLandDefinition()
+    return sceneState ? sceneState : new SceneStateDefinition()
+  }
+
+  async getContentLandDefinition(): Promise<SceneStateDefinition | undefined> {
```
nchamo: Should this be private?

kernel/packages/scene-system/stateful-scene/BuilderStatefulActor.ts
```diff
@@ -0,0 +1,51 @@
...
+    const builderProjectByCoordinates = await this.sceneStorage.getProjectManifestByCoordinates(
+      this.land.sceneJsonData.scene.base
+    )
+    if (builderProjectByCoordinates) {
+      return deserializeSceneState(builderProjectByCoordinates)
+    }
+
+    //if there is no project associated to the land, we search the last builder project deployed in the land
+    const catalyst = this.getContentClient()
+    const entity = await catalyst.fetchEntityById(EntityType.SCENE, this.land.sceneId)
```
nchamo: There is no need to fetch the entity from the content server again. The metadata is already in `this.land.sceneJsonData`. Sorry I didn't realize before

kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
```diff
@@ -6,26 +6,105 @@ import { Authenticator } from 'dcl-crypto'
...
+import { SceneStateDefinition } from 'scene-system/stateful-scene/SceneStateDefinition'
+import { ILand } from 'shared/types'
+import { ExplorerIdentity } from 'shared/session/types'
+import { deserializeSceneState, serializeSceneState } from 'scene-system/stateful-scene/SceneStateDefinitionSerializer'
 
 declare const window: any
 
 @registerAPI('SceneStateStorageController')
 export class SceneStateStorageController extends ExposableAPI {
+  private builderApiManager = new BuilderServerAPIManager()
```
nchamo:
```suggestion
  private readonly builderApiManager = new BuilderServerAPIManager()
```

kernel/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
```diff
@@ -0,0 +1,247 @@
...
+  async updateProjectManifest(builderManifest: BuilderManifest, identity: ExplorerIdentity) {
+    try {
+      this.setManifestOnServer(builderManifest, identity)
+    } catch (e) {
+      console.trace(e)
+      return undefined
+    }
+  }
+
+  async createProjectWithCoords(land: ILand, identity: ExplorerIdentity): Promise<BuilderManifest> {
```
nchamo: I think you can pass only the coords to this method, instead of the whole `ILand` object, right?

kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
```diff
@@ -6,26 +6,105 @@ import { Authenticator } from 'dcl-crypto'
...
+    if (newProject) {
+      this.builderManifest = newProject
+      const translatedManifest = fromBuildertoStateDefinitionFormat(this.builderManifest.scene)
+      return serializeSceneState(translatedManifest)
+    }
+    return undefined
+  }
+
+  @exposeMethod
+  async createProjectWithCoords(land: ILand): Promise<SerializedSceneState> {
```
nchamo: I think you can pass only the coords to this method, instead of sending the whole `ILand` object

kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
```diff
@@ -6,26 +6,105 @@ import { Authenticator } from 'dcl-crypto'
...
+    const newProject = await this.builderApiManager.createProjectWithCoords(land, this.getIdentity())
+    this.builderManifest = newProject   
+    return serializeSceneState(new SceneStateDefinition())
```
nchamo: I don't fully get what this method returns. Why are we returning an empty scene definition? Is it necessary? Can't we just know on the other side and create the scene definition from scratch?

kernel/packages/shared/apis/SceneStateStorageController/types.ts
```diff
@@ -20,3 +20,74 @@ export enum CONTENT_PATH {
...
+export type BuilderAsset = {
+  id: string
+  assetPackId: string
+  name: string
+  model: string
+  script: string
+  thumbnail: string
+  tags: string[]
+  category: string
+  contents: Map<string,string>
```
nchamo: Are you sure this is a `Map`? It looks like the `BuilderAsset` is downloaded from the builder API, so I don't think this is actually a `Map`

kernel/packages/shared/apis/SceneStateStorageController/StorableSceneStateTranslation.ts
```diff
@@ -18,6 +21,84 @@ type StorableComponent = {
...
+    //we add the entity to builder format
+    let builderEntity: BuilderEntity = {
+      id: entityId,
+      components: builderComponentsIds
+    }
+    entities[builderEntity.id] = builderEntity
+  }
+
+  // We create the scene and add it to the manifest
+  var sceneState: BuilderScene = {
```
nchamo: Please try to avoid using `var`. Always use `const`, unless you need to modify the variable, and in that case use `let`

kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
```diff
@@ -6,26 +6,105 @@ import { Authenticator } from 'dcl-crypto'
...
+  async saveProjectManifest(serializedSceneState: SerializedSceneState) {
+    //Deserialize the scene state
+    const sceneState: SceneStateDefinition = deserializeSceneState(serializedSceneState)
+
+    //Convert the scene state to builder scheme format
+    let builderManifest = toBuilderFromStateDefinitionFormat(sceneState, this.builderManifest)
+
+    //We get all the assetIds from the gltfShapes so we can fetch the corresponded asset
+    let idArray: string[] = []
+    Object.entries(builderManifest.scene.components).forEach((component) => {
```
nchamo: It looks like you are not using the keys, so if you did `Object.values` instead of `Object.entries` you would only get what you need, and avoid the `[1]`

kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
```diff
@@ -6,26 +6,105 @@ import { Authenticator } from 'dcl-crypto'
...
+      if (component[1].type === 'GLTFShape') {
+        let found = false
+        Object.entries(builderManifest.scene.assets).forEach((assets) => {
```
nchamo: Same here, but instead of `Object.entries` you could use `Object.keys`

kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
```diff
@@ -6,26 +6,105 @@ import { Authenticator } from 'dcl-crypto'
...
+        }
+      }
+    })
+
+    //We fetch all the assets that the scene contains since builder needs the assets
+    builderManifest.scene.assets = await this.builderApiManager.getAssets(idArray)
+
+    //This is a special case. The builder needs the ground separated from the rest of the components so we search for it.
+    //Unity handles this, so only 1 entitty will contain the "ground" category. We can safely assume that we can search it and assign
+    Object.entries(builderManifest.scene.assets).forEach((asset) => {
```
nchamo: In cases where you do need both the key and value, you can so something like:
```
Object.entries(builderManifest.scene.assets).forEach(([assetId, asset]) => {
...
```
That way, it will be clearer in the code later

_Note: I'm not sure if the key and value are actually asset id and asset. It was just an example_

kernel/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
```diff
@@ -0,0 +1,247 @@
...
+  createHeaders(idToken: string) {
+    if (!idToken) return {}
+    const headers: Record<string, string> = {
+      'Content-Type': 'application/json',
+      Authorization: `Bearer ${idToken}`
+    }
+    return headers
+  }
+
+  authorize = (identity: ExplorerIdentity, method: string = 'get', path: string = '') => {
```
nchamo:
```suggestion
  authorize(identity: ExplorerIdentity, method: string = 'get', path: string = '') {
```
Both ways are valid, but we should try to make it consistent.

kernel/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
```diff
@@ -0,0 +1,247 @@
...
+      method: 'PUT',
+      body: body
+    }
+
+    const response = await fetch(urlToFecth, params)
+    const data = await response.json()
+    return data
+  }
+
+  getBaseUrl(): string {
```
nchamo: I think (correct me if I'm wrong) but this and some other methods here could be private. If so, please make them private

kernel/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
```diff
@@ -0,0 +1,247 @@
...
+      const urlToFecth = `${this.getBaseUrl()}${queryParams}`
+
+      let params: RequestInit = {
+        headers: this.authorize(identity, 'get', '/projects/' + queryParams)
+      }
+
+      const response = await fetch(urlToFecth, params)
+      const data = await response.json()
+
+      var value = JSON.parse(JSON.stringify(data))
```
nchamo: This is weird. Why is this necessary?

kernel/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
```diff
@@ -0,0 +1,247 @@
...
+      try {
+        // Fetch unknown assets
+        const response = await fetch(`${BASE_ASSET_URL}?${queryParams}`)
+        const { data }: { data: BuilderAsset[] } = await response.json()
+        data.map((builderAsset) => builderAsset).forEach((asset) => this.assets.set(asset.id, asset))
+      } catch (e) {
+        console.trace(e)
+      }
+    }
+    const test: Record<string, BuilderAsset> = {}
```
nchamo: The name `test` is weird. Is this intended?

kernel/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
```diff
@@ -0,0 +1,247 @@
...
+      const urlToFecth = `${this.getBaseUrl()}${queryParams}`
+
+      let params: RequestInit = {
+        headers: this.authorize(identity, 'get', '/projects/' + queryParams)
+      }
+
+      const response = await fetch(urlToFecth, params)
+      const data = await response.json()
+
+      var value = JSON.parse(JSON.stringify(data))
```
nchamo: This is weird. Why is this necessary?

kernel/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
```diff
@@ -0,0 +1,247 @@
...
+      if (value['data'] === 'false') {
```
nchamo:
```suggestion
      if (value.data === 'false') {
```

kernel/packages/scene-system/stateful.scene.system.ts
```diff
@@ -17,47 +19,69 @@ class StatefulWebWorkerScene extends Script {
...
     // Listen to storage requests
     this.eventSubscriber.on('stateEvent', ({ data }) => {
       if (data.type === 'StoreSceneState') {
```
nchamo: Since we now have a new event type `SaveSceneState`, should we rename this to `PublishSceneState`? And maybe also change the method name `storeState`. 

I think it would be clearer that way

kernel/packages/scene-system/stateful.scene.system.ts
```diff
@@ -17,47 +19,69 @@ class StatefulWebWorkerScene extends Script {
...
         this.sceneStateStorage
-          .storeState(sceneId, serializeSceneState(this.sceneState))
+          .storeState(sceneId, serializeSceneState(this.sceneActor.getState()))
           .catch((error) => this.error(`Failed to store the scene's state`, error))
       }
     })
+
+    // Listen to save scene requests
+    this.eventSubscriber.on('stateEvent', ({ data }) => {
+      if (data.type === 'SaveSceneState') {
```
nchamo, line 76: I understand that this is called when the user exists BiW mode, right? But how will it behave in the future? Should every change be updated in the builder?

kernel/packages/scene-system/stateful-scene/BuilderStatefulActor.ts
```diff
@@ -0,0 +1,51 @@
+﻿import { CatalystClient } from 'dcl-catalyst-client'
+import { EntityType } from 'dcl-catalyst-commons'
+import { SceneStateDefinition } from './SceneStateDefinition'
+import { ILand } from 'shared/types'
+import { SceneStateStorageController } from 'shared/apis/SceneStateStorageController/SceneStateStorageController'
+import { deserializeSceneState } from './SceneStateDefinitionSerializer'
+
+declare var window: any
+
+export class BuilderStatefulActor {
```
nchamo, line 6: Why are we calling this a `StatefulActor` if it doesn't extend `StatefulActor`? It probably makes sense depending on how we expect it to evolve, but I wanted to understand your plans for it

nchamo (commented):
Great work!

Added some comments

kernel/packages/scene-system/stateful-scene/SceneStateActor.ts
D4rWiNSS: At the beginning of the project, it has a purpose but since this has evolved, now it doesn't have a clear purpose, I just let it as it can be helpful in the future, however, if you think we should eliminate it, I can eliminate it without problems!

kernel/packages/scene-system/stateful.scene.system.ts
D4rWiNSS: Yes right now it is called only when the user exists BiW mode, however, in the future, this will change.

The idea is that unity will handle the saving of the scene every X seconds or when important events occurs (like exit biw, enter preview mode, ...etc)

The problem right now is that unity report the changes to kernel each X seconds, not always, so we can end with a different state in the builder API, in the kernel, and in Unity so if we move the responsibility to Unity, we can be sure that kernel and builder API will have always the same state, also at important events we can force Unity to report the changes and save. This way we can be sure that at important events, Unity, kernel and builder API have the same state

Let me know if this wasn't clear

kernel/packages/scene-system/stateful-scene/BuilderStatefulActor.ts
D4rWiNSS: The idea behind this actor is to handle the state of the scene based on the server state.

Right now since it only asks the builder server, it is more descriptive that way, however, in the future when the decentraliced server system comes in, it should be renamed to something like `ServerStatefulActor` and it should manage all the interaction with the server. So for me, it is an actor that handles the communication with the server. 

I don't know if this makes sense to you, please let me know!

kernel/packages/scene-system/stateful-scene/SceneStateActor.ts
nchamo: Yeah, I think we should go back to it just being a `SceneStateDefinition` instead of an actor. We need to try to keep things as simple as possible, since the problem itself it already complex enough

kernel/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
```diff
@@ -0,0 +1,245 @@
...
+    try {
+      this.setManifestOnServer(builderManifest, identity)
+    } catch (e) {
+      console.trace(e)
+      return undefined
+    }
+  }
+
+  async createProjectWithCoords(coordinates: string, identity: ExplorerIdentity): Promise<BuilderManifest> {
+    var builderManifest = this.createEmptyDefaultBuilderScene(coordinates, identity.rawAddress)
```
nchamo:
```suggestion
    const builderManifest = this.createEmptyDefaultBuilderScene(coordinates, identity.rawAddress)
```

kernel/packages/shared/apis/SceneStateStorageController/StorableSceneStateTranslation.ts
```diff
@@ -18,6 +21,84 @@ type StorableComponent = {
...
+    limits: builderManifest.scene.limits,
+    ground: builderManifest.scene.ground
+  }
+
+  builderManifest.scene = sceneState
+  return builderManifest
+}
+
+export function fromBuildertoStateDefinitionFormat(scene: BuilderScene): SceneStateDefinition {
+  var sceneState = new SceneStateDefinition()
```
nchamo:
```suggestion
  const sceneState = new SceneStateDefinition()
```

kernel/packages/shared/apis/SceneStateStorageController/StorableSceneStateTranslation.ts
```diff
@@ -18,6 +21,84 @@ type StorableComponent = {
...
+
+  const entitiesMap = new Map(Object.entries(scene.entities))
+  const componentMap = new Map(Object.entries(scene.components))
+
+  for (let entity of entitiesMap.values()) {
+    let components: Component[] = []
+    for (let componentId of entity.components.values()) {
+      for (let builderComponent of componentMap.values()) {
```
nchamo: You already have a map (`componentMap`). Can't you search by id instead of iterating through all of them and checking if they have the same if?

kernel/packages/shared/apis/SceneStateStorageController/StorableSceneStateTranslation.ts
```diff
@@ -18,6 +21,84 @@ type StorableComponent = {
...
+  }
+
+  builderManifest.scene = sceneState
+  return builderManifest
+}
+
+export function fromBuildertoStateDefinitionFormat(scene: BuilderScene): SceneStateDefinition {
+  var sceneState = new SceneStateDefinition()
+
+  const entitiesMap = new Map(Object.entries(scene.entities))
```
nchamo: I find it kind of weird that you convert the object into a `Map` just to get the values. Can't you use `Object.values(scene.entities)`?

kernel/packages/scene-system/stateful.scene.system.ts
nchamo: So to make it clear, the source of truth will be Unity from now on, right? Sounds good to me, just want to be 100% clear on this

kernel/packages/scene-system/stateful.scene.system.ts
```diff
@@ -17,47 +19,69 @@ class StatefulWebWorkerScene extends Script {
...
+
+    this.rendererActor.sendInitFinished()
     this.log('Sent initial load')
 
+    // Listen to scene state events
+    this.listenToEvents(sceneId)
+  }
+
+  private listenToEvents(sceneId: string): void {
     // Listen to storage requests
```
nchamo: Could you update this comment so that it is something publish related?

kernel/packages/scene-system/stateful-scene/BuilderStatefulActor.ts
```diff
@@ -0,0 +1,40 @@
+﻿import { SceneStateDefinition } from './SceneStateDefinition'
+import { ILand } from 'shared/types'
+import { deserializeSceneState } from './SceneStateDefinitionSerializer'
+import { ISceneStateStorageController } from 'shared/apis/SceneStateStorageController/ISceneStateStorageController'
+
+export class BuilderStatefulActor {
+  constructor(
+    protected readonly land: ILand,
+    protected readonly realmDomain: string,
```
nchamo: I think you don't need this anymore

pravusjif (approved):

nchamo (approved):
Great work!

kernel/packages/scene-system/stateful.scene.system.ts
D4rWiNSS: Kernel will continue to contain the current definition of the scene, but Unity will have now the responsibility to save the project

PR #2281, opened 2021-04-12
fix: transform position overflow crash

### LINKED ISSUE #248: Infinity Engine scene crashes with the RC

BrianAmadori:
This PR aims to fix https://github.com/decentraland/unity-renderer/issues/248

Test link: https://play.decentraland.zone/branch/fix/transform-position-overflow-crash/index.html?ENV=org&position=-71%2C-36


unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Utils.cs
```diff
@@ -653,5 +653,19 @@ public static bool CompareFloats( float a, float b, float precision = 0.1f )
...
+        /// </summary>
+        /// <param name="renderer"></param>
+        /// <returns>The bounds value if the value is correct, or a mocked bounds object with clamped values if its too far away.</returns>
+        public static Bounds GetSafeBounds( this Renderer renderer )
+        {
+            if ( Vector3.Distance(renderer.transform.position, Vector3.zero) > 1000000 )
+                return new Bounds( Vector3.one * 1000000, Vector3.one * 0.1f );
+
+            return renderer.bounds;
+        }
```
pbosio: question:
Why are we using `Vector3.Distance` here instead of `Abs(renderer.transform.position.magnitude)`?
EDIT: I think `Abs` will not be needed in that situation but could be better if we use `sqrMagnitude`
`if ( Mathf.Abs(renderer.transform.position.sqrMagnitude) > limit*limit )`?
suggestion: 
Can we use a local constant for the limit to avoid the same number to be repeated? `const float limit = 1000000;`

pbosio (approved):

pravusjif (approved):
I agree with Pato's suggestion to use `Mathf.Abs(renderer.transform.position.magnitude)` or even much more performant `Mathf.Abs(renderer.transform.position.sqrMagnitude)` would be better.

unity-client/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Utils.cs
pbosio: mmm.. after giving it another though, if we are trying to avoid an overflow, maybe using `sqrMagnitude` it's not the best approach 🤔 😅

BrianAmadori: It's a tiny overflow, not an actual float overflow, I think your suggestion is correct. Ty!.

BrianAmadori: Feedback applied.

Issue #1812, opened 2020-12-14
Changing volume of an AudioSource causes it to restart from beginning


cfravel:
Create a scene with a longish mp3 (i suggest music) for testing AudioSource
Add a "button" or other clickable entity, or a UI, that changes its volume up or down.
// expected: sound continues playing from the current locaition, but audio changes
// actual:  the sound restarts from the beginning, at thenew volume.

This is not, new, it has been true probably the whole time we've had AudioClip/AudioSource.

This really screws up mp3-based scenes where the volume needs to change during the sound, so I hope it can be prioritized.

An example is the DJ Console in the Conference Center  Cafe and on the DistrictX Tower.   The volume control isn't really usable by the DJ to fade a song in/out.

pbosio:
Fixed: should be live on next release and available now on `decentraland-ecs@next `

PR #2285, opened 2021-04-12
fix: AudioSource.playOnce() restart the audio when is called

### LINKED ISSUE #15: AudioSource.playOnce() not working as expected
`AudioSource.playOnce()` should play the whole sound file from start to finish, but if the sound was already playing, it will just continue from where it was.

This problem is evidenced very clearly in this scene:

https://github.com/decentraland-scenes/shooting-range

In this shooting range, the sound from shooting has a fairly long echo. If I shoot in rapid succession, I don't hear anything new, I just continue hearing the tail of the sound of the first shot. I should be hearing a distinct shot sound starting from the start each time I shoot. That's what `playOnce()` suggests.

In this same scene, I tried adding a `playing = false` to force the sound to reset, but that doesn't work either. So if the sound is already playing, I don't have any workaround or way to hear the sound again from the start until it finishes playing.

Since it's a single AudioSource, I don't expect keep hearing the ending of the first playing of the sound in another channel, but I do expect the same channel to reset back to the start.

kuruk-mm:
# What?

This bug was introduced in #1914 because there was a bug (#1812) that when we changed the volume or pitch from the SDK, the audio restarted. So @pbosio fix it, but we lost the behavior to restart sounds with playOnce().

(more details about the bug -> decentraland/unity-renderer#15)

We discussed with @pbosio and @BrianAmadori the solution to add a timestamp when we execute playOnce(), and we only restart the audio when we see a different timestamp. That is what this PR does.

I added one test-scene where we have a system changing the volume constantly of a music, and other box to reproduce multiple times the same sounds restarting on every playOnce().

Video (turn on the sound):

https://user-images.githubusercontent.com/12563266/114466083-5b225400-9bbe-11eb-95e6-717901fdef35.mp4

# Why?

Fix decentraland/unity-renderer#15


unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs
```diff
@@ -162,11 +164,12 @@ private void ApplyLoadedAudioClip(DCLAudioClip clip)
             }
 
             Model model = (Model) this.model;
-            if (audioSource.enabled && model.playing && !audioSource.isPlaying)
+            if (audioSource.enabled && model.playing && playedAtTimestamp != model.playedAtTimestamp)
             {
                 //To remove a pesky and quite unlikely warning when the audiosource is out of scenebounds
                 audioSource.Play();
             }
```
pbosio: It looks like (I didn't test it) if I do the following:
```
audio.playOnce()
audio.playing = false
audio.playing = true
```
the audio won't play any more unless `playOnce` is called called again?

pbosio:
@sandrade-dcl 
I've just saw this: https://github.com/decentraland/explorer/blob/7331d9e19e03f89a5f307e2118df908022bbd198/unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs#L76-L77 😅
What line should stay and what line should be removed? 🤣

kuruk-mm:
> @sandrade-dcl
> I've just saw this:
> 
> https://github.com/decentraland/explorer/blob/7331d9e19e03f89a5f307e2118df908022bbd198/unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs#L76-L77
> 
> 
> What line should stay and what line should be removed?

Haha, I guess the second line is correct because it overwrites the first one.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs
kuruk-mm: I reread what Brian said, and he thought about this case. I force-pushed the change.
What do you think now?

sandrade-dcl:
> @sandrade-dcl
> I've just saw this:
> 
> https://github.com/decentraland/explorer/blob/7331d9e19e03f89a5f307e2118df908022bbd198/unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs#L76-L77
> 
> 😅
> What line should stay and what line should be removed? 🤣

Hmmm... I'm reviewing the changes history and this second line was introduced by @Suduck in 4a684422912ff723b94b4de3c3ff204846edaeb9. I guess he could confirm, but yeah it seems to be the second line the good one.

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs
```diff
@@ -162,11 +164,15 @@ private void ApplyLoadedAudioClip(DCLAudioClip clip)
...
 
             Model model = (Model) this.model;
-            if (audioSource.enabled && model.playing && !audioSource.isPlaying)
+            bool shouldPlay = playedAtTimestamp != model.playedAtTimestamp ||
+                              (model.playing == true && audioSource.isPlaying == false);
```
sandrade-dcl:
```suggestion
            bool shouldPlay = playedAtTimestamp != model.playedAtTimestamp ||
                              (model.playing && !audioSource.isPlaying);
```

sandrade-dcl (approved):
Good job!

D4rWiNSS:
> > @sandrade-dcl
> > I've just saw this:
> > https://github.com/decentraland/explorer/blob/7331d9e19e03f89a5f307e2118df908022bbd198/unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs#L76-L77
> > 
> > 😅
> > What line should stay and what line should be removed? 🤣
> 
> Hmmm... I'm reviewing the changes history and this second line was introduced by @Suduck in [4a68442](https://github.com/decentraland/explorer/commit/4a684422912ff723b94b4de3c3ff204846edaeb9). I guess he could confirm, but yeah it seems to be the second line the good one.

This was introduced with the changes in the models of components. It seems like an auto-merge from visual studio setting both lines or me doing it without noticing it.

In any case...

image

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs
```diff
@@ -162,11 +164,15 @@ private void ApplyLoadedAudioClip(DCLAudioClip clip)
...
+
+            if (audioSource.enabled && model.playing && shouldPlay)
             {
                 //To remove a pesky and quite unlikely warning when the audiosource is out of scenebounds
                 audioSource.Play();
             }
```
D4rWiNSS, line 173: Our coude convention for one liner if is like that ->
```
if (audioSource.enabled && model.playing && shouldPlay)
    audioSource.Play();
```

D4rWiNSS (approved):
Good job! I only left a small comment!

D4rWiNSS:
> > @sandrade-dcl
> > I've just saw this:
> > https://github.com/decentraland/explorer/blob/7331d9e19e03f89a5f307e2118df908022bbd198/unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs#L76-L77
> > 
> > 😅
> > What line should stay and what line should be removed? 🤣
> 
> Hmmm... I'm reviewing the changes history and this second line was introduced by @Suduck in [4a68442](https://github.com/decentraland/explorer/commit/4a684422912ff723b94b4de3c3ff204846edaeb9). I guess he could confirm, but yeah it seems to be the second line the good one.

Looking again at this, it should be the first one right @sandrade-dcl ? It seems like it should also play if it is a portable experience with a global scene event if you are not in the scene. I can be wrong though, I just wanting to be sure

sandrade-dcl:
> > > @sandrade-dcl
> > > I've just saw this:
> > > https://github.com/decentraland/explorer/blob/7331d9e19e03f89a5f307e2118df908022bbd198/unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs#L76-L77
> > > 
> > > 😅
> > > What line should stay and what line should be removed? 🤣
> > 
> > 
> > Hmmm... I'm reviewing the changes history and this second line was introduced by @Suduck in [4a68442](https://github.com/decentraland/explorer/commit/4a684422912ff723b94b4de3c3ff204846edaeb9). I guess he could confirm, but yeah it seems to be the second line the good one.
> 
> Looking again at this, it should be the first one right @sandrade-dcl ? It seems like it should also play if it is a portable experience with a global scene event if you are not in the scene. I can be wrong though, I just wanting to be sure

Hmmm you're right! I didn't notice the `globalScene.isPortableExperience` check. So definitely the good line is the first. Please @kuruk-mm take it into account 👍

unity-client/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs
kuruk-mm: > Good job! I only left a small comment!

I think is more readable because of the comment. If I remove the comment, I could put like you said. Should I do that? :thinking:

D4rWiNSS: Nah, left it that way then!

pbosio (approved):
LGTM!
Could we get a thumbsup from @nchamo on this? since we are making modifications on a SDK component?

menduz (approved):

PR #2291, opened 2021-04-14
feat: add domain info to current realm reported to renderer

pbosio:


pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/DataStore/Variables/RealmsInfo/CurrentRealmVariable.cs
```diff
@@ -20,13 +20,19 @@ public class CurrentRealmModel
     {
         public string layer;
         public string serverName;
+        public string domain = string.Empty;
 
         public bool Equals(CurrentRealmModel other)
```
BrianAmadori, line 25: Why do we need all those `Equals` overrides?. We can't do anything about this to have only one of them? (suggestion for the future).

BrianAmadori (approved):
LGTM

unity-client/Assets/Scripts/MainScripts/DCL/DataStore/Variables/RealmsInfo/CurrentRealmVariable.cs
pbosio: I don't like it either, but, I found out that on renderer we mostly handle realms like `{layer}-{serverName}` (friends, players nearby from comms, etc) and we don't care nor we have the info for the `domain`, for that reason I decided that it was better to have that override

D4rWiNSS (approved):

PR #2294, opened 2021-04-15
feat: add deployments fetch to catalysts from renderer

pbosio:


unity-client/Assets/Scripts/MainScripts/DCL/ServiceProviders/Catalyst/Catalyst.cs
```diff
@@ -0,0 +1,115 @@
...
+public interface ICatalyst : IDisposable
+{
+    Promise<SceneDeploymentPayload> GetDeployedScenes(string[] parcels, bool onlyCurrentlyPointed = true, string sortBy = null, string sortOrder = null);
+    Promise<string> GetDeployments(DeploymentOptions deploymentOptions);
+    Promise<string> Get(string url);
+}
+
+public class Catalyst : ICatalyst
+{
+    private const int CACHE_TIME_MSECS = 5 * 60 * 1000;
```
D4rWiNSS: miliseconds for cache can be an overkill, are you thinking in a special case where you will need the milliseconds there? instead of seconds/minutes?

D4rWiNSS (dismissed):
Small request!

unity-client/Assets/Scripts/MainScripts/DCL/ServiceProviders/Catalyst/Catalyst.cs
pbosio: You remind me I have to check something related to that :)
It's using milliseconds cause the task delay it's expecting milliseconds. But the caching here is set for 5 minutes

pbosio:
Setting this PR back to draft after talking to @nchamo it would be better to use entities api instead of deployments

sandrade-dcl (approved):
Very clean code, congrats! Only as suggestion, I would add a minimum of documentation for the `Catalyst` class that would describe their main responsibilities.

pbosio:
> Very clean code, congrats! Only as suggestion, I would add a minimum of documentation for the `Catalyst` class that would describe their main responsibilities.

Agree. added at commit https://github.com/decentraland/explorer/pull/2294/commits/f836b2f95d99aff1939ed8df424dbb558a9b8856
Thanks!

unity-client/Assets/Scripts/MainScripts/DCL/ServiceProviders/Catalyst/Catalyst.cs
```diff
@@ -0,0 +1,218 @@
...
+
+        // NOTE: remove from cache after CACHE_TIME time passed
+        CoroutineStarter.Start(RemoveCache(url, CACHE_TIME));
+    }
+
+    private IEnumerator RemoveCache(string url, float delay)
+    {
+        yield return WaitForSecondsCache.Get(delay);
+        cache?.Remove(url);
+    }
```
D4rWiNSS, line 217: Would it be more perfomance to just check time at the moment of do the call to check if the cache has expire? This way we wouldn't have a courutine running

D4rWiNSS (approved):

unity-client/Assets/Scripts/MainScripts/DCL/ServiceProviders/Catalyst/Catalyst.cs
pbosio: I thought about that, but my main concern was that the result json from the catalyst can be huge and I didn't want it to stay stored at the dictionary permanently.

PR #2303, opened 2021-04-19
chore: update wearables AB Converter with new endpoint

pravusjif:
* Updated wearables AB Converter to fetch from the new endpoint (previous endpoint will be deprecated soon)
* Created new structure to parse new endpoint JSON response (we can no longer use WearableItem class directly as the JSON structure has changed)
* Updated wearables AB Converter logic to fetch the different "pages" of wearables (the new endpoint doesn't retrieve all the wearables like the previous one)


pbosio (approved):

unity-client/Assets/ABConverter/Client.cs
```diff
@@ -61,18 +61,128 @@ public Settings(ContentServerUtils.ApiTLD tld = ContentServerUtils.ApiTLD.ORG)
...
+                
+                foreach (var wearableData in deployments)
+                {
+                    // Populate new WearableItem with fetched data
+                    WearableItem wearable = new WearableItem()
+                    {
+                        id = wearableData.metadata.id,
+                        category = wearableData.metadata.data.category,
+                        tags = wearableData.metadata.data.tags,
+                        baseUrl = "https://peer.decentraland.org/content/contents/",
```
D4rWiNSS: Can we move this to a constant?

unity-client/Assets/ABConverter/Client.cs
```diff
@@ -370,7 +480,7 @@ public static void DumpAllNonBodiesWearables()
             EnsureEnvironment();
 
             // For debugging purposes we can intercept this item list with LinQ for specific wearables
-            List<WearableItem> avatarItemList = GetAvatarMappingList("https://wearable-api.decentraland.org/v2/collections")
+            List<WearableItem> avatarItemList = GetAvatarMappingList("https://peer.decentraland.org/content/deployments?entityType=wearable&onlyCurrentlyPointed=true")
```
D4rWiNSS: can we move this to constants?

D4rWiNSS (approved):
LGTM! Just small suggestions

unity-client/Assets/ABConverter/Client.cs
pravusjif: good idea, done.

unity-client/Assets/ABConverter/Client.cs
pravusjif: good idea, done.

sandrade-dcl (approved):

PR #2304, opened 2021-04-20
fix: detect bogus texture

AjimenezDCL:
Sometimes Unity hides CORS issues if the response succeeded. The consequence is that ugly red question mark texture.

There's no way to detect this but to compare the texture with the, so called, bogus texture.

Got the idea from here: https://wiki.unity3d.com/index.php/TextureBogusExtensions


pbosio (commented):
Could we make `IsQuestionMarkPNG` public and static so instead of checking every texture against that we do it manually in each particular scenarios where we don't want the question mark?
I would prefer leaving the question mark rather than doing all of this checks for every texture :P

AjimenezDCL:
Performance wise this check can be ignored. There's an early return based on the texture size (the question mark is an 8x8 texture, so it would be really strange to retrieve a texture of that size).

Do you think we will have an use-case of wanting the question mark to appear?

pbosio:
> Performance wise this check can be ignored. There's an early return based on the texture size (the question mark is an 8x8 texture, so it would be really strange to retrieve a texture of that size).
> 
> Do you think we will have an use-case of wanting the question mark to appear?

If we know the height and width of the question mark texture, could we check for that first? before casting the texture to a texture2d and encoding it?

AjimenezDCL:
You are absolutly right, I got the code and didnt check it much further. Nice catch

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Texture/AssetPromise_Texture.cs
```diff
@@ -130,20 +126,30 @@ protected override bool AddToLibrary()
...
         public override object GetId()
         {
             // We only use the id-with-settings when storing/reading from the library
             return idWithDefaultTexSettings;
         }
 
-        public bool UsesDefaultWrapAndFilterMode()
+        public bool UsesDefaultWrapAndFilterMode() { return wrapMode == DEFAULT_WRAP_MODE && filterMode == DEFAULT_FILTER_MODE; }
+
+        static byte[] questionMarkPNG = { 137, 80, 78, 71, 13, 10, 26, 10, 0, 0, 0, 13, 73, 72, 68, 82, 0, 0, 0, 8, 0, 0, 0, 8, 8, 2, 0, 0, 0, 75, 109, 41, 220, 0, 0, 0, 65, 73, 68, 65, 84, 8, 29, 85, 142, 81, 10, 0, 48, 8, 66, 107, 236, 254, 87, 110, 106, 35, 172, 143, 74, 243, 65, 89, 85, 129, 202, 100, 239, 146, 115, 184, 183, 11, 109, 33, 29, 126, 114, 141, 75, 213, 65, 44, 131, 70, 24, 97, 46, 50, 34, 72, 25, 39, 181, 9, 251, 205, 14, 10, 78, 123, 43, 35, 17, 17, 228, 109, 164, 219, 0, 0, 0, 0, 73, 69, 78, 68, 174, 66, 96, 130, };
```
pbosio: just this suggestion: could we make `questionMarkPNG` readonly and move it to the top of the class?

pbosio (approved):
Great! 
I'll just leave you a almost "Obsessive-compulsive disorder" suggestion 😅

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Texture/AssetPromise_Texture.cs
```diff
@@ -130,20 +126,30 @@ protected override bool AddToLibrary()
...
+        internal bool IsQuestionMarkPNG(Texture tex)
         {
-            return wrapMode == DEFAULT_WRAP_MODE && filterMode == DEFAULT_FILTER_MODE;
+            if (!tex)
+                return true;
+
+            if (tex.width * tex.height != questionMarkPNG.Length )
```
menduz: Should this consider the 4 channels?
`tex.width * tex.height * 4 != questionMarkPNG.Length`

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Texture/AssetPromise_Texture.cs
```diff
@@ -130,20 +126,30 @@ protected override bool AddToLibrary()
...
+                return false;
+
+            byte[] png1 = (tex as Texture2D).EncodeToPNG();
```
menduz, line 148: Not sure about unity, but encoding textures back to PNG requires querying the GPU memory which is a pipeline blocking operation. I'd try to avoid this as much as possible. Do we have the underlying number of updates to a texture or something else in the texture loader we can hook into?

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Texture/AssetPromise_Texture.cs
AjimenezDCL: My calculations were completly off, to be honest not even taking into account the 4 channels would work.

Somehow the 8x8 image ends up in an byte array of 122 entries. Not sure what the PNG encoding is doing in the background.

I'm going to add a hardcoded check for the texture's width and height. In any case, if Unity decides to change that bogus image, the test will fail, so not the end of the world.

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Texture/AssetPromise_Texture.cs
AjimenezDCL: I think you are right here, therefore the early return. As commented in your previous concern, I'm gonna solve it comparing against the texture's width and height.

So we would end up doing this check only against other 8x8 textures and that should be affordable.

pravusjif: Not sure how much performance we would gain (it could be meassured with [`System.Diagnostics.Stopwatch`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch)) but if we want to avoid the PNG encoding altogether, we may be able to compare using the Tex2D pixels as we do [here](https://github.com/decentraland/explorer/blob/e11f86b07b114ae230ef01eb2d17dd7726b16e4e/unity-client/Assets/Scripts/Tests/VisualTests/VisualTestHelpers.cs#L208) for our visual tests.

pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Texture/AssetPromise_Texture.cs
AjimenezDCL: First we would need to have the texture itself contained in the project and we could run into false-negative issues due to floating point precission errors.

That's doable as well but I don think we are gaining that much.

AjimenezDCL:
Merging this. If you think any of the open discussion require an actionable, let me know and we can open a ticket about it and solve it there

PR #2307, opened 2021-04-20
fix: Reduced analytics calls and removed queue

sicarul:
The current analytics tracking is quite broken, a lot of unnecesary messages are sent, and the queueing solution makes it so we have hundreds if not thousands of messages queued for segment in a single session. A very serious implication of the current status is that for example if you have 150 events in the queue and the explorer fails, the "generic_error" event has to wait until those 150 events are sent before being sent, which most probably won't happen because the user will close or refresh the browser.

This PR fixes this issue by:

- Removing the queueing (the function has been renamed to reflect that as well)
- Removing the following events:
  - **USER_DATA**: Was sending the full user profile of each new user that spawned nearby, unnecesary, this would scale exponentially in a big event like a concert.
  - **USER_REMOVED**: Similar to the previous one, but when a user disappears from sight or disconnects
  - **USER_EXPRESSION**: Each time someone uses an emote, maybe we can add a new event when the current player does it, but the current event triggers if anyone nearby uses an emote.
  - **User Position**: Was reporting each time the user moves (at least 50 in 3d euclidean distance) or 10 seconds have passed) - this seems excessive and unnecesary, as we already have the "Move To Parcel" event. To compensate for the data loss, i've added the exact position to the "Move to Parcel event".
  - **raw perf report**: This was the raw data for the performance report, an array of 1000 numbers sent around once per second, this is not sustainable nor that useful. To compensate for deleting this report, i've added percentiles 1, 5, 10, 20 and 80 to the performance report.
  - **hiccup report**: The 3 metrics generated in this report are sent at the same time as the performance report, i've added these 3 metrics to the performance report instead.
- Changing the **Comms Status V2** to fire once per minute instead of each 30 seconds


kernel/packages/shared/session/getPerformanceInfo.ts
```diff
@@ -15,27 +15,20 @@ export function getPerformanceInfo(data: { samples: string; fpsIsCapped: boolean
...
+    p20: sorted[Math.ceil(length * 0.20)],
+    p50: sorted[Math.ceil(length * 0.50)],
     p75: sorted[Math.ceil(length * 0.75)],
-    p90: sorted[Math.ceil(length * 0.9)],
+    p80: sorted[Math.ceil(length * 0.80)],
+    p90: sorted[Math.ceil(length * 0.90)],
     p95: sorted[Math.ceil(length * 0.95)],
     p99: sorted[Math.ceil(length * 0.99)],
     max: sorted[length - 1],
-    capped: data.fpsIsCapped
```
menduz, line 24: We are using this bit

menduz (changes requested):

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -138,16 +138,9 @@ export class BrowserInterface {
     if (newWindow != null) newWindow.opener = null
   }
 
-  public PerformanceHiccupReport(data: { hiccupsInThousandFrames: number; hiccupsTime: number; totalTime: number }) {
-    queueTrackingEvent('hiccup report', data)
-  }
-
-  public PerformanceReport(data: { samples: string; fpsIsCapped: boolean }) {
+  public PerformanceReport(data: { samples: string; fpsIsCapped: boolean, hiccupsInThousandFrames: number; hiccupsTime: number; totalTime: number }) {
     const perfReport = getPerformanceInfo(data)
```
BrianAmadori, line 142:
```suggestion
    const perfReport = getRawPerformanceInfo(data)
```

If we are going to keep one of the events, please let's keep the raw getter and remove the `getPerformanceInfo` method altogether. Kernel shouldn't be calculating percentiles, that should be done from the query on whatever service we use to consume the data (i.e. metabase or grafana).

kernel/packages/shared/session/getPerformanceInfo.ts
menduz: Sorry for the short review, I'm in mobile. @pravusjif can you confirm we are using this flag to measure two different rendering profiles?

BrianAmadori (changes requested):
Ty for the PR! Left a request.

kernel/packages/shared/session/getPerformanceInfo.ts
BrianAmadori: Yes, we do

sicarul: @menduz  I have not removed the flag, it appears that way because i added a comma and more data to this event

kernel/packages/unity-interface/BrowserInterface.ts
sicarul: @BrianAmadori are you sure? The raw event is certainly less CPU usage, but it uses much more network bandwith instead (Sends an array with 1000 numbers), i'm fine with either.

menduz: In the case of sending only the raw, the data ingestion in that case would be at least 1000x more expensive. The only thing we need from histograms are percentiles, total samples and sum of all samples. Everything else can be calculated in the reporting tools. Raw data is not convenient or more useful than those bits of information.

sicarul: I've been testing this function locally and it takes about 0.05ms to run in my (admitedly fast intel i9) notebook, and this would run only once every 1000 samples, do we really need to optimize this on the CPU side instead of bandwitdth? Thanks!

BrianAmadori: The network usage points feel valid to me, those are strong concerns. 

I'd like to reconsider this approach in the case we need to scale the data filtering, we have to be careful of client side computations as per-frame performance it's very critical for us. In any case, take my LGTM.

BrianAmadori (approved):

kernel/packages/shared/session/getPerformanceInfo.ts
sicarul: @menduz can you confirm this appears ok from your end? I have not removed the flag
Thanks!

menduz: Totally my bad, I got confused for the small screen

menduz (approved):

PR #2309, opened 2021-04-21
feat: add animation state layers to play multiple animations at diffe…

### LINKED ISSUE #107: Any running animations need to be stopped before playing a new one
If I play an animation and then I want to play another animation, or the same animation again, I need to first explicitly stop the last played animation. It's a very common newbie pitfall, they aren't able to play an animation a second time, and the reason is that they need to stop the animation explicitly before they call it again.

This doesn't work

```ts
sit.play()

Input.instance.subscribe('BUTTON_DOWN', ActionButton.POINTER, true, (e) => {
  sit.play()
}
```

When dealing with a model that has several animations that may have ben called before, the workarounds needed for this can get really nasty.
See this example:
https://github.com/decentraland-scenes/Block-dog

Here you need to explicitly stop all these animations every time before you want to play another one
    sit.playing = false
    stand.playing = false
    walk.playing = false
    drink.playing = false
    idle.playing = false
    stand.playing = false

This issue has been around ever since we migrated away from Babylon. Back when HardlyDifficult did the escape room tutorial (April 2019), he avoided mentioning the workaround for this bc it was a bug we were supposedly going to fix by the time the videos were released.  ....sure.

kuruk-mm:
...erent layers and add a new function on animator to stop all animation states

# What?
Fix decentraland/unity-renderer#107

Add layers to `AnimationState`, when you execute `play` you will pause all animation states from the same layer.

Add new `stop` function on `Animator` to stop all animations

Add new `reset` parameter in `play` function to decide if you want to reset the current clip animation.

New test scene:

https://user-images.githubusercontent.com/12563266/115482447-11bcaf00-a225-11eb-8c89-c59717a6516a.mp4

**Extra bugfix!:** `Weight` of an `AnimationState` was not being send to Unity when we called it by `setParams`.

# Why?
Details: decentraland/unity-renderer#107

### DIFF
--- a/kernel/packages/decentraland-ecs/src/decentraland/Components.ts
+++ b/kernel/packages/decentraland-ecs/src/decentraland/Components.ts
@@ -536,21 +536,45 @@ export class Animator extends Shape {
   }
 
   /**
-   * Adds an AnimationState to the animation lists.
+   * Resets and pauses the animation state, if the clip is null it will stop all animations on this animator
    */
-  stop() {
-    for (let i = 0; i < this.states.length; i++) {
-      const clip = this.states[i]
-      clip.stop()
+  stop(clip?: AnimationState) {
+    if (clip) {
+      clip.playing = false
+      clip.shouldReset = true
+    } else {
+      for (let i = 0; i < this.states.length; i++) {
+        const animationState = this.states[i]
+        this.stop(animationState)
+      }
     }
   }
 
-  // @internal
-  pauseLayer(layer: number) {
+  /**
+   * Starts the animation
+   */
+  play(clip: AnimationState, reset: boolean = false) {
     for (let i = 0; i < this.states.length; i++) {
-      const clip = this.states[i]
-      if (clip.layer === layer) {
-        clip.pause()
+      const animationState = this.states[i]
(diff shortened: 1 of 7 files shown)


kernel/packages/decentraland-ecs/src/decentraland/Components.ts
```diff
@@ -513,6 +513,16 @@ export class Animator extends Shape {
     clip.onChange(() => {
       this.dirty = true
     })
+
+    // Access private member to hide internal playCallback implementation to the user
+    clip['playCallback'] = (layer: number) => {
```
menduz: Make the method public and annotate it with `// @internal` to strip its declaration from the generated bundle. It avoids type unsafety

menduz (changes requested):

kernel/packages/decentraland-ecs/src/decentraland/Components.ts
kuruk-mm: Done. I changed this. Now `AnimationState` has his own owner `Animator` to call that function directly. We decide that for future changes to `Animator`.

Thanks for the info about `// @internal` I didn't know that.

menduz (approved):

pravusjif (approved):

PR #2312, opened 2021-04-21
refactor: send wearables as v2 to Unity

nchamo:
Before this change, Kernel was mapping wearables to v1 before sending them to Unity. Now, we are always mapping them to v2.

The idea is to always have wearable ids be urns, and we will also translate them into v2, and store them like that on the catalog.

_Note: this needs to be tested and merged together with https://github.com/decentraland/unity-renderer/issues/251_

TestLink: https://explorer.decentraland.zone/branch/refactor/send-wearables-as-v2-to-unity/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:refactor/send-wearables-as-v2-to-unity


kernel/packages/shared/catalogs/sagas.ts
```diff
@@ -306,12 +294,14 @@ function* fetchWearablesV1(filters: WearablesRequestFilters) {
...
-  let response: Wearable[]
+  let response: PartialWearableV2[]
   if (filters.wearableIds) {
     // Filtering by ids
     response = filters.wearableIds
       .map((wearableId) =>
-        wearableId === 'dcl://base-avatars/SchoolShoes' ? 'dcl://base-avatars/Moccasin' : wearableId
+        wearableId === 'urn:decentraland:off-chain:base-avatars:SchoolShoes'
+          ? 'urn:decentraland:off-chain:base-avatars:Moccasin'
+          : wearableId
```
BrianAmadori, line 319: What's this hack for? It's still relevant?

nchamo: Yes, this is still necessary. The wearable-api has both `dcl://base-avatars/Moccasin` and `dcl://base-avatars/moccasin`. So we converted `Moccasin` to `SchoolShoes` during the migration to the catalysts. But if we need to use the v1 catalog, we need to translate it back to `Moccasin` for it work

pablitar (approved):
LGTM!

PR #2316, opened 2021-04-23
feat: add sdk event to subscribe to camera mode change and we can get…

### LINKED ISSUE #275: SDK to know what camera mode the player is in
Interaction in 1st vs 3rd person is drastically different, but the scene can’t adapt to these different conditions, as there’s no information about what mode a player is using.

More details in the RFC (containing other requests)
https://docs.google.com/document/d/1YisfNs5xq12gfkFnYmcu0adUFJcKS3chVwss-nklsOo/edit#

kuruk-mm:
… the value of it too

# What? 
Fix decentraland/unity-renderer#275

https://user-images.githubusercontent.com/12563266/115803183-c2a68380-a3b6-11eb-9d9d-1507f1691205.mp4

**New SDK code:**
```
onCameraModeChanged.add(({ cameraMode }) => {
  log("onCameraModeChanged:", cameraMode)
})

log("Current CameraMode:", Camera.instance.cameraMode)
```

**Test scene:**
Turns red the box on first person camera, and blue on third person camera.

@menduz I changed `SceneRuntime.subscribe` because if we call that function multiple times it subscribe more than once in each event.
Let me know if it correct.

### DIFF
--- a/kernel/packages/scene-system/sdk/SceneRuntime.ts
+++ b/kernel/packages/scene-system/sdk/SceneRuntime.ts
@@ -82,6 +82,11 @@ export abstract class SceneRuntime extends Script {
 
   eventSubscriber!: EventSubscriber
 
+  // this dictionary contains the list of subscriptions.
+  // the boolean value indicates if the client is actively
+  // listenting to that event
+  subscribedEvents: { [event: string]: boolean } = {}
+
   onUpdateFunctions: Array<(dt: number) => void> = []
   onStartFunctions: Array<Function> = []
   onEventFunctions: Array<(event: any) => void> = []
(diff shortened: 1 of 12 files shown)


kuruk-mm:
I'm getting a very strange error on the CI.

```
/tmp/explorer/kernel/scripts/buildEcsTypes.js:21
        throw new Error(`The file ${dtsFile} contains imports:\n${content}`);
        ^

Error: The file /tmp/explorer/kernel/packages/decentraland-ecs/types/dcl/index.d.ts contains imports:
```

It seems it is a problem with the `buildEcsTypes.ts` script. When I execute `make watch`, the first time I got the error, and the second one everything goes well.

Going deeper I found that the code generated in:`/tmp/explorer/kernel/packages/decentraland-ecs/types/dcl/index.d.ts` has this:
```
/**
 * This event is triggered when you change your camera between 1st and 3rd person
 * @public
 */
declare const onCameraModeChanged: Observable<{
    cameraMode: import("./Types").CameraMode;
}>;
```

And looks that it is an `import` like the error said.

I don't know exactly how to fix this.

kernel/packages/scene-system/sdk/SceneRuntime.ts
```diff
@@ -82,6 +82,11 @@ export abstract class SceneRuntime extends Script {
 
   eventSubscriber!: EventSubscriber
 
+  // this dictionary contains the list of subscriptions.
+  // the boolean value indicates if the client is actively
+  // listenting to that event
+  subscribedEvents: { [event: string]: boolean } = {}
```
menduz: Use of Map<string, boolean> may make more sense

menduz (approved):
Looks good, I suggested a minor change.

For future events, I'd think about a way to skip kernel and send events directly from unity to the sdk. Same mechanism bur avoiding observables in kernel.

BrianAmadori (approved):

PR #2319, opened 2021-04-23
fix: AlphaTexture apply fixes

BrianAmadori:
AlphaTexture apply was causing issues with some textures that expected luminance -> alpha

We were doing alphaTexture.r -> alpha

Also, it seems that `AlphaTexture` was being taken into account for alpha test cutoff when it shouldn't. (a bit weird, but we should keep the bw compatibility).

Should address https://github.com/decentraland/unity-renderer/issues/283

Tested on: 
- lights that are deployed in 150, -100
- sammich scene deployed at 49, -46


AjimenezDCL (approved):

kuruk-mm (approved):
It's so much better now :)

pravusjif (approved):

PR #2323, opened 2021-04-26
fix: onpointerevent colliders instantiation

pravusjif:
### WHY
While debugging some interaction problems with OnPointerDown/OnClick components in one of the casinos, I observed these rare behaviours happening with the OnPointerEvent MeshColliders: 

* Sometimes we have OnPointerEvent child colliders created in some meshes but not registered in `CollidersManager.i.colliderInfo`, so the PointerEventsController can't relate the Mesh Collider with its UUIDComponent.

* Some meshes pools have their original gameobject with its OnPointerEvent child colliders as well, this shouldn't be happening.

* Sometimes we get 2 OnPointerEvent child colliders for the same mesh, this bugs the interaction because sometimes the stray collider is targeted, and that collider is not bound to any UUIDComponent

Apparently those symptoms where appearing because in this case some shape properties (like visibility or collisions enabled) are being updated before the shape has finished loading, thus triggering the `OnShapeUpdated` event early.

#### ~~First quick-fix aproach applied (Old)~~

* ~~Removed condition that escaped re-initializing the PointerEvent colliders of the last updated shape, since I suspect the entity's `OnShapeUpdated` is being called several times for the same shape.~~

* ~~Implemented update in OnPointerEventColliders to avoid creating PointerEvent colliders for a renderer that already has any, recycling that one and registering it in `CollidersManager.i.colliderInfo` if it's not already there.~~

#### Final fix aproach applied

* Removed condition that escaped re-initializing the PointerEvent colliders of the last updated shape

* Added `isLoaded` conditional in `LoadableShape.ApplyChanges()` to avoid calling the `OnShapeUpdated()` event before the shape has finished loading

#### New casino testing (the user needs to be whitelisted): https://play.decentraland.zone/branch/fix/onPointerEventCollidersInstantiation/index.html?CATALYST=peer-testing.decentraland.org&ENV=org&position=-95%2C109&realm=testing-amber

### DIFF
--- a/unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadableShape.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadableShape.cs
@@ -139,7 +139,7 @@ namespace DCL.Components
                 if (updateCollisions)
                     ConfigureColliders(entity);
 
-                entity.OnShapeUpdated?.Invoke(entity);
+                TriggerOnShapeUpdated(entity);
             }
 
             previousModel = model;
@@ -255,7 +255,7 @@ namespace DCL.Components
 
             ConfigureColliders(entity);
 
-            entity.OnShapeUpdated?.Invoke(entity);
+            TriggerOnShapeUpdated(entity);
 
             OnFinishCallbacks?.Invoke(this);
             OnFinishCallbacks = null;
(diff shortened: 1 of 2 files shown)


unity-client/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadableShape.cs
```diff
@@ -284,5 +284,12 @@ public override void CallWhenReady(Action<ISharedComponent> callback)
                 OnFinishCallbacks += callback;
             }
         }
+
+        private void TriggerOnShapeUpdated(IDCLEntity entity)
```
BrianAmadori:
```suggestion
        private void RaiseOnShapeUpdated(IDCLEntity entity)
```

BrianAmadori (approved):

PR #2326, opened 2021-04-26
refactor: new way to send events from unity to the sdk

### LINKED ISSUE #293: Design a way to skip kernel and send events directly from Unity to the SDK
Same mechanism from the actual one but avoiding observables in kernel

kuruk-mm:
# What? 
Fix decentraland/unity-renderer#293
Related to #2316

Now we can send the event without going throght SceneWorkerSystem

# Why? 
To avoid repeating code for each event, now we can send an event from Unity to SDK directly.

We discussed with @menduz to remove all "`on`" prefix in `IEvents` to avoid redundancy to avoid code like:
```events.on('onCameraModeChanged')```

### DIFF
--- a/kernel/packages/decentraland-ecs/src/decentraland/Events.ts
+++ b/kernel/packages/decentraland-ecs/src/decentraland/Events.ts
@@ -48,7 +48,7 @@ function createSubscriber(eventName: keyof IEvents) {
  * This event is triggered when you change your camera between 1st and 3rd person
  * @public
  */
-export const onCameraModeChanged = new Observable<IEvents['onCameraModeChanged']>(createSubscriber('onCameraModeChanged'))
+export const onCameraModeChanged = new Observable<IEvents['cameraModeChanged']>(createSubscriber('cameraModeChanged'))
 
 /**
  * These events are triggered after your character enters the scene.
@@ -82,8 +82,8 @@ export function _initEventObservables(dcl: DecentralandInterface) {
           onLeaveScene.notifyObservers(event.data as IEvents['onLeaveScene'])
           return
         }
-        case 'onCameraModeChanged': {
-          onCameraModeChanged.notifyObservers(event.data as IEvents['onCameraModeChanged'])
+        case 'cameraModeChanged': {
+          onCameraModeChanged.notifyObservers(event.data as IEvents['cameraModeChanged'])
           return
         }
       }
--- a/unity-client/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
+++ b/unity-client/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
@@ -105,12 +105,6 @@ namespace DCL.Interface
             public string eventType;
             public T payload;
         }
-        [System.Serializable]
-        public class AllScenesEvent<T>
-        {
-            public string eventType;
-            public T payload;
-        }
 
         [System.Serializable]
         public class UUIDEvent<TPayload>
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by menduz** [comment]
Sadly we can't change the camera event name if we released at least one version of the ECS in latest. Nor we can't change any of the events names without breaking backwards compatibility. It should go back to the "on" prefix. I'd add a comment to the IEvents struct to _not_ prefix the events with "on" in the case we add new ones

**Comment by kuruk-mm** [comment]
> Sadly we can't change the camera event name if we released at least one version of the ECS in latest. Nor we can't change any of the events names without breaking backwards compatibility. It should go back to the "on" prefix. I'd add a comment to the IEvents struct to _not_ prefix the events with "on" in the case we add new ones

`onCameraModeChanged`/`cameraModeChanged` is not in production, it was merged 3 days ago #2316. So we have time to change it.

And I already added the IEvent struct comment that you mention:
```
 * Note: Don't use `on` prefix for IEvents to avoid redundancy with `event.on("onEventName")` syntax.
```
let me know if you want something different :smile:

**Comment by menduz** [comment]
Did this work in the wild? I can't see in the code where all the scenes that are loaded in the map get added to the `loadedSceneWorkers` map

Edit. Just found it

File: unity-client/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs [hunk 1]
```diff
@@ -625,6 +631,14 @@ public static void SendMessage<T>(string type, T message)
         private static RequestWearablesPayload requestWearablesPayload = new RequestWearablesPayload();
         private static SearchENSOwnerPayload searchEnsOwnerPayload = new SearchENSOwnerPayload();
 
+        public static void SendAllScenesEvent<T>(string eventType, T payload)
+        {
+            AllScenesEvent<T> allScenesEvent = new AllScenesEvent<T>();
```
menduz: This should be a private static to avoid allocations in every message

menduz (approved):

menduz:
I was talking about the `AllScenesEvent<T> allScenesEvent = new AllScenesEvent<T>();`

That should be private static, to avoid allocations on every method call @kuruk-mm

kernel/packages/decentraland-ecs/src/decentraland/Events.ts
```diff
@@ -48,7 +48,7 @@ function createSubscriber(eventName: keyof IEvents) {
  * This event is triggered when you change your camera between 1st and 3rd person
  * @public
  */
-export const onCameraModeChanged = new Observable<IEvents['onCameraModeChanged']>(createSubscriber('onCameraModeChanged'))
+export const onCameraModeChanged = new Observable<IEvents['cameraModeChanged']>(createSubscriber('cameraModeChanged'))
```
BrianAmadori, line 51:
```suggestion
export const cameraModeChanged = new Observable<IEvents['cameraModeChanged']>(createSubscriber('cameraModeChanged'))
```
If we are going to change the event name, we shouldn't do it for this name as well?

BrianAmadori: I see that the enter and leave scenes have the `on` moniker too. Maybe we should rename them too?

kuruk-mm: ~~Yes, and there are more IEvents with `on`~~
~~I talked with @menduz and that code it's already in production, we cannot change it.~~

~~`cameraModeChange` is not in production yet.~~

Edit: I understood wrong. The events exposed to the SDK must have `on` prefix. We need to avoid the `on` in IEvent interface.

kuruk-mm: There is a pending issue about this: decentraland/unity-renderer#295

kuruk-mm:
I drafted this PR, it's blocked until we finish the unity-renderer migration.

PR #2329, opened 2021-04-27
chore: add onpointerevent colliders creation tests

pravusjif:
* Added mechanism to avoid re-creating OPEColliders on shapes that dont change their mesh
* Added missing tests


unity-client/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/OnPointerEventColliders.cs
```diff
@@ -73,6 +76,33 @@ Collider CreateCollider(Renderer renderer)
             return meshCollider;
         }
 
+        private bool AlreadyCreatedColliders(Renderer[] rendererList)
```
AjimenezDCL: I think something like `AreCollidersCreated` would fit better here.

unity-client/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/OnPointerEventColliders.cs
```diff
@@ -73,6 +76,33 @@ Collider CreateCollider(Renderer renderer)
...
+                        break;
+                    }
+                }
+
+                if (!foundChildCollider)
+                    return false;
+            }
+            
+            return true;
+        }
```
AjimenezDCL, line 109:
```suggestion
            for (int i = 0; i < rendererList.Length; i++)
            {
                for (int j = 0; j < colliders.Length; j++)
                {
                    if (colliders[j].transform.parent == rendererList[i].transform)
                        return true;
                }
            }
            
            return false;
        }
```

I think this is equivalent and we dont have to rely on extra variables.

unity-client/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/Tests/UUIDComponentShould.cs
```diff
@@ -72,5 +75,77 @@ public IEnumerator BeDestroyedCorrectlyWhenReceivingComponentDestroyMessage()
...
+            };
+            var onPointerDownComponent = TestHelpers.EntityComponentCreate<OnPointerDown, OnPointerDown.Model>(scene, entity,
+                model, CLASS_ID_COMPONENT.UUID_CALLBACK);
+            
+            // 2. Attach a shape
+            var shapeModel = new LoadableShape<LoadWrapper_GLTF, LoadableShape.Model>.Model();
+            shapeModel.src = Utils.GetTestsAssetsPath() + "/GLB/Lantern/Lantern.glb";
+            var componentId = TestHelpers.CreateAndSetShape(scene, entityId, DCL.Models.CLASS_ID.GLTF_SHAPE, JsonConvert.SerializeObject(shapeModel));
+            
+            // 4. Change a shape component property while it loads
```
AjimenezDCL: Where is our step 3? :p

AjimenezDCL (approved):
Approving with small suggestions!

sandrade-dcl (approved):
LGTM! (Be careful because the unity tests are currently failing)

unity-client/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/OnPointerEventColliders.cs
pravusjif: Actually I think it's not exactly the same, since in this new form we would return true on the first renderer that has a collider matching it as its parent, but we should check **every renderer and every collider**.

Probably your approach would work but because I think it would be strange to have 1 renderer that is the same as the previous colliders creation and other renderers that are not. But I'd keep it as I pushed it originally just to cover all the renderer and colliders.

If we want to avoid using that bool I could check if the collider's `j` reached the end of the collection without breaking the traversing.

Also this check could be done with LinQ too but I think it would be much less performant.

pravusjif:
Migrated PR to new repo: https://github.com/decentraland/unity-renderer/pull/350

PR #2330, opened 2021-04-28
fix: Block Explore and Tutorial buttons + Hide scene blockers while in BiW editor

sandrade-dcl:
### What this PR includes?
- [x] **Block Explore and Tutorial buttons**: Players can activate the "Explore & Play" and "Tutorial" functionality while editing a land, which can result in unsaved or unpublished changes. This funciton should be deactivated while editing a land, greying out the icon.
- [x] **Hide scene blockers**: Players shouldn't see any green blocker while ther're editing a land.

### Testing
1. Go to: https://explorer.decentraland.zone/branch/fix/Block-Explore-and-Tutorial-buttons-from-the-taskbar-while-in-BiW/index.html?ENV=zone&ENABLE_BUILDER_IN_WORLD&position=100,100&realm=fenrir-amber
2. Activate the Builder In World editor pressing the "**K**" key.
3. See the new loading screen.


pravusjif (approved):

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
```diff
@@ -454,6 +455,12 @@ public void EnterEditMode()
             HUDController.i.builderInWorldMainHud.RefreshCatalogAssetPack();
         }
 
+        if (HUDController.i.taskbarHud != null)
+        {
+            HUDController.i.taskbarHud.SetExploreInteractable(false);
+            HUDController.i.taskbarHud.SetMoreTutorialInteractable(false);
+        }
+
```
D4rWiNSS, line 463: A little worried about the approach, if we add a new button, we should remember to block it here or we could use it.

Can we have a function like "SetBuilderInWorldMode" and all buttons that are not explicitly marked as "work with builder" be desactivated?

D4rWiNSS (changes requested):
Good Job! I have found a bug 

After teleporting with /goto skybox is not set back correctly

image

unity-client/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
D4rWiNSS: for example, the quest button, or the portable experience button

sandrade-dcl: Yeah, it is already planned. We'll implement in a separated PR 👍

D4rWiNSS (approved):

PR #2308, opened 2021-04-20
chore: improve renderer package download consistency

kuruk-mm:
# What?
Fix decentraland/unity-renderer#253

`make watch` / `make update-renderer`: it will download the current branch build renderer package (or from master if there is no branch-build package published yet).
`make update-renderer-master` (removing old make update-renderer-branch): it will download and overwrite the current build files with the ones from the master branch build renderer package (you need to have `make watch` executed to take effect otherwise will be overwritten)

# Why?
To have more consistency on how we download the unity renderer packages.
And I separated it in a new script to have more flexibility with multiple conditionals.

PR #2331, opened 2021-04-28
chore: add gitattribute to treat shell scripts as lf (line feed) a…

kuruk-mm:
…nd avoid conflicts in windows os with crlf (carriage return + line feed)

# What? 
After #2308 we started executing `updateRenderer.sh` when we use `make watch`.

Git in Windows were modifying the end line from `lf` (line feed) to `crlf` (carriage return + line feed) and that was breaking the script.

So I added in .gitattributes to considerer all `.sh` as `lf`.

More info about `lf` vs `crlf`: https://stackoverflow.com/a/1552770

Edit: When I merge I will rename the commit message from `refactor` to `chore`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -70,3 +70,5 @@
 #*.RTF   diff=astextplain
 TestResources/Audio/Train.wav filter=lfs diff=lfs merge=lfs -text
 *.jspre linguist-language=JavaScript
+# Shell scripts require LF
+*.sh text eol=lf


kuruk-mm:
I tested and this works for the future. If you already pulled the changes from #2308 maybe you need to manually change from CRLF->LF.

In Visual Studio Code we can change it easily in the bottom right bar:
image

CRLF -> LF

Other options is going back in git before #2308 and then goes forwards.
Example:
```
git fetch origin
git checkout 545821505e4507f4c7d64d7eb4745f244a2cde6f # Go to a commit before the updateRenderer was added
git checkout refactor/sh-lf
```

sandrade-dcl (changes requested):
I have just test it and it seems it continues giving the same error:

image

D4rWiNSS (approved):

sandrade-dcl (approved):
Ok, in my case, after executing the commands:
```
git fetch origin
git checkout 545821505e4507f4c7d64d7eb4745f244a2cde6f # Go to a commit before the updateRenderer was added
git checkout refactor/sh-lf
```

It works! Thanks Mateo!

pravusjif (approved):

PR #2333, opened 2021-04-29
feat: biw ux improvements

D4rWiNSS:
# What? 

1. TopActionButtons are disabled when you don't have selected entities
2. Reset button now resets position, rotation and scale
3. Tooltips has been updated to reflect the shortcuts changes
4. Entity list view now have a red color in the icons that match the style of the editor


sandrade-dcl (approved):
The code looks fine! But I cannot test the build because you have some tests failing. I'll do when it is fixed 👍

AjimenezDCL (approved):
Looks good!

PR #2342, opened 2021-05-04
feat: add builder projects panel HUD (kernel side)

pbosio:
Test link: https://explorer.decentraland.zone/branch/feat/builder-projects-panel-hud/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:feat/add-hud-builderprojectpanel&ENABLE_BUILDER_IN_WORLD
Sibling PR (renderer side): https://github.com/decentraland/unity-renderer/pull/400


menduz (approved):

menduz:
Approved, what would happen if we call `.ConfigureHUDElement(x, obj)` with an `x` that unity does not recognize? If it doesn't fail, I'd say to merge this anytime you want. Otherwise we wait for the renderer artifacts

pbosio:
> Approved, what would happen if we call `.ConfigureHUDElement(x, obj)` with an `x` that unity does not recognize? If it doesn't fail, I'd say to merge this anytime you want. Otherwise we wait for the renderer artifacts

I would be just yet another error on the console :P
I will lobby for an approval on the renderer side first, it's just a 3 files PR

PR #2343, opened 2021-05-04
chore: remove unity client folder

BrianAmadori:
This PR:

- Removes the unity folder
- Modifies the CI to remove all the unity stuff and the `package` step of the CI. Now the unity-renderer build taken should be always the `1.X.X` npm package.

Caveats:

The kernel release CI may stop supporting new unity builds because we don't publish the NPM package with the `release` tag anymore, as the previous release CI setup relied on the publishing-npm-packages-with-current-branch-name-as-tag feature. 

This should be addressed in another PR.


D4rWiNSS (approved):
We need to remove the unity test and build unity from CI in order to make this work.

I didn't go over the 7956 files but I guess it is ok?

menduz:
Ready when you ready @BrianAmadori

menduz (approved):

PR #2347, opened 2021-05-08
feat: add sdk event when the scene started/is ready

### LINKED ISSUE #276: SDK to receive an event when scene finishes loading
Loading of scenes is also quite broken. It would be of huge help to have an event that alerts when all the models in a scene finished loading. Today all you can do is guess. For example if animations start playing as soon as their model is loaded, then they all end up out of sync with each other. The only way to sync them is to stop and start them again at an arbitrary time after the loading hopefully finished.

More info in the RFC
https://docs.google.com/document/d/1YisfNs5xq12gfkFnYmcu0adUFJcKS3chVwss-nklsOo/edit#

kuruk-mm:
# What? 
Resolves decentraland/unity-renderer#276

Add observable event to the sdk, that it triggers when the scene started / is ready.

# Test
I added a test-scene to see the result.

Currently the only way to test is cloning this branch and executing: `make test-scenes` and then `make watch`

### DIFF
--- a/kernel/packages/decentraland-ecs/src/decentraland/Events.ts
+++ b/kernel/packages/decentraland-ecs/src/decentraland/Events.ts
@@ -75,10 +75,10 @@ export const onLeaveSceneObservable = new Observable<IEvents['onLeaveScene']>(cr
 export const onLeaveScene = onLeaveSceneObservable
 
 /**
- * This event is triggered once the scene should start.
+ * This event is triggered after all the resources of the scene were loaded (models, textures, etc...)
  * @public
  */
-export const onSceneStartObservable = new Observable<IEvents['sceneStart']>(createSubscriber('sceneStart'))
+export const onSceneReadyObservable = new Observable<IEvents['sceneStart']>(createSubscriber('sceneStart'))
 
 /**
  * @internal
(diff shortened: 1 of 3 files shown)


kernel/packages/decentraland-ecs/src/decentraland/Events.ts
```diff
@@ -74,6 +74,12 @@ export const onLeaveSceneObservable = new Observable<IEvents['onLeaveScene']>(cr
 /* @deprecated Use onLeaveSceneObservable instead. */
 export const onLeaveScene = onLeaveSceneObservable
 
+/**
+ * This event is triggered once the scene should start.
```
menduz: Can we have a more narrow description? Isn't the scene supposed to have started to at the stage this `const` is being evaluated? It may be confusing.

It is not clear to me what is this event actually signaling.

menduz (changes requested):

kernel/packages/decentraland-ecs/src/decentraland/Events.ts
kuruk-mm: I will changed the comment to something more described.

This happens when the Scene is ready and it loaded its models.

Maybe change the event name to:

`onSceneReadyObservable`?

One example in the world for thid event is when a scene is loading we see green limiters.  When those limiters disappear (the scene finish loading) it would execute this event.

menduz: Ohh I see, "onSceneReadyObservable" sounds good enough for the use case, maybe something like "Event triggered after all the resources of the scene were loaded (models, textures, etc...)" may help the description

menduz (approved):

nchamo (approved):
Great work!

PR #2355, opened 2021-05-11
feat: LOS in explorer settings

pravusjif:
We enable LOS dynamic change from the explorer settings.

Testing link: 
https://play.decentraland.zone/branch/feat/LOSInExplorerSettings/index.html?ENV=org&position=-86%2C-135&realm=baldr-blue&renderer=urn%3Adecentraland%3Aoff-chain%3Arenderer-artifacts%3Afeat%2FLOSInExplorerSettings

Screen Shot 2021-05-13 at 12 12 51


menduz (approved):

kernel/packages/decentraland-loader/lifecycle/controllers/parcel.ts
```diff
@@ -35,24 +37,30 @@ export class ParcelLifeCycleController extends EventEmitter {
...
     const sightedParcelsSet = new Set<string>()
 
-    const newlySightedParcels = sightedParcels.filter(parcel => {
+    const newlySightedParcels = sightedParcels.filter((parcel) => {
       sightedParcelsSet.add(parcel)
       return this.parcelSighted(parcel)
     })
 
-    const secureParcels = new Set(parcelsInScope(this.config.lineOfSightRadius + this.config.secureRadius, position))
+    const secureParcels = new Set(parcelsInScope(this.config.lineOfSightRadius + secureRadius, this.currentPosition))
```
BrianAmadori, line 56: We might want to review the `secureRadius` parameter later. Right now the secure/LOS ratio is 1:1, but with this change the LOS can be 10 and the `secureRadius` would remain at 4.

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -339,6 +339,12 @@ export class BrowserInterface {
     setDelightedSurveyEnabled(data.enabled)
   }
 
+  public SetScenesLoadRadius(data: { newRadius: number }) {
```
BrianAmadori, line 342: Just in case: we must make sure the `lineOfSightRadius` is properly initialized at the client start according to the saved settings. 

Also, please confirm that we don't have any race condition issues between our settings initialization and the `Lifecycle.initialize` dispatch that comes with `startUnitySceneWorkers()` call on Unity's initialization from kernel side.

BrianAmadori: Another detail: I see that this PR is not addressing how the setting is going to interact with our current LOS url param. The LOS param should be deprecated? Or maybe we have to make sure that if the LOS param exists, the setting is forced at least at the explorer initialization stage.

kernel/packages/decentraland-loader/lifecycle/controllers/parcel.ts
pravusjif: Yes, I thought of the same but forgot to change that. I will keep the secureRadius fed into the function to be always 1:1 with the LOS, when reporting position.

kernel/packages/unity-interface/BrowserInterface.ts
pravusjif: Great points, the saved settings are working correctly in local but not in the build, I'm debugging that issue.

Regarding the LOS URL param, I think that URL param should have priority over the panel setting (unless the setting is changed in runtime after starting the application)

kernel/packages/decentraland-loader/lifecycle/controllers/parcel.ts
pravusjif: Done

kernel/packages/unity-interface/BrowserInterface.ts
pravusjif: Done.
* Now the saved value is used when initializing the explorer
* The LOS URL param is prioritized when initializing the explorer

kernel/packages/decentraland-loader/lifecycle/controllers/parcel.ts
```diff
@@ -53,6 +54,7 @@ export class ParcelLifeCycleController extends EventEmitter {
     })
 
     const secureParcels = new Set(parcelsInScope(this.config.lineOfSightRadius + secureRadius, this.currentPosition))
+    console.log('PRAVS - Kernel - Update loaded parcels with LOS...' + this.config.lineOfSightRadius)
```
BrianAmadori:
```suggestion
```

BrianAmadori (approved):
Great job!

PR #2360, opened 2021-05-13
feat: add player expression, position change and rotation change even…

### LINKED ISSUE #300: Add player events for the SDK
- [x] Event when player position or rotation changes
- [ ] ~~Event when player changes wearables~~ (derived to: #435)
- [x] Event when player does an emote

kuruk-mm:
…ts to the sdk

# What? 
Resolves decentraland/unity-renderer#300

New SDK events when we...
... play emotes/expressions ~~(multiplayer event!)~~
... change our position
... change our camera rotation

https://user-images.githubusercontent.com/12563266/118167917-9f8f5280-b3fd-11eb-962c-216684490085.mp4

the test-scenes does:
- The text depends on what `emote` you execute ~~(multiplayer!!)~~ (using `onPlayerExpressionObservable`)
- The text always `lookAt` your position (using `onPositionChangedObservable`)
- The `box shape` replicates your camera rotation (using `onRotationChangedObservable`)

**Incidence:**
I found a problem with the `Expressions` where the each expression event from the local user was being executed with an invalid `uuid`, more info -> decentraland/unity-renderer#300 

# Why? 
Add features to the SDK

### DIFF
--- a/kernel/packages/decentraland-ecs/src/decentraland/Events.ts
+++ b/kernel/packages/decentraland-ecs/src/decentraland/Events.ts
@@ -1,7 +1,9 @@
 import { EventConstructor } from '../ecs/EventManager'
 import { Observable } from '../ecs/Observable'
+import { ReadOnlyQuaternion, ReadOnlyVector3 } from './math'
 import { DecentralandInterface, IEvents, RaycastResponsePayload, CameraMode } from './Types'
 export { CameraMode }
+export { ReadOnlyVector3, ReadOnlyQuaternion }
 
 /**
  * @public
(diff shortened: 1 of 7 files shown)


nearnshaw:
How far do other players have to be to get events from their emotes?

There's no way to explicitly know the position of another player, so if I'm getting data of another player doing an emote maybe they are in another scene, or even super far away, and I would probably only care if they're near

kuruk-mm:
> How far do other players have to be to get events from their emotes?
> 
> There's no way to explicitly know the position of another player, so if I'm getting data of another player doing an emote maybe they are in another scene, or even super far away, and I would probably only care if they're near

It will notify every scene that is loaded/loading in your range of vision.
If you want to get the event just when you're inside the scene you need to combine it with `onEnterSceneObservable` and `onLeaveSceneObservable`

I removed the `multiplayer` behavior in `onPlayerExpressionObservable` like we discussed. It only works for yourself.

menduz (approved):
The expressionId event is sound, it is great that now we are sending a consisntent uuid every event. 

I'd remove the position change and rotation events. Since @nearnshaw has a very valid point on the consistency of the fields. We may generate new events with consistent naming for those cases, or merge them in another PR

PR #2364, opened 2021-05-14
feat: Selecting quests url based on TLD + making the quests saga failure tolerant

pablitar:
We will soon have two quests URLs: One for production and one for development. Explorer should select which one based on its TLD.

The server can also be adjusted through the QUESTS_SERVER_URL url parameter.

### DIFF
--- a/kernel/packages/config/index.ts
+++ b/kernel/packages/config/index.ts
@@ -56,8 +56,10 @@ export namespace parcelLimits {
   export const minParcelX = -150
   export const minParcelZ = -150
 
-  export const descriptiveValidWorldRanges = getWorld().validWorldRanges
-    .map((range) => `(X from ${range.xMin} to ${range.xMax}, and Y from ${range.yMin} to ${range.yMax})`)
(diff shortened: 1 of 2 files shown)


kernel/packages/config/index.ts
```diff
@@ -289,6 +289,8 @@ export function getServerConfigurations() {
   const metaFeatureFlagsBaseUrl = `https://feature-flags.decentraland.${notToday}/explorer.json`
   const ASSET_BUNDLES_DOMAIN = qs.ASSET_BUNDLES_DOMAIN || `content-assets-as-bundle.decentraland.${TLDDefault}`
 
+  const QUESTS_SERVER_URL = qs.QUESTS_SERVER_URL ?? `https://quests-api.decentraland.${notToday}`
```
nchamo: I think that `notToday` would be either `org` or `zone`, not `io`. Is that ok? Would it still work?

nchamo (approved):

kernel/packages/config/index.ts
pablitar: Oops...

PR #2369, opened 2021-05-18
feat: generic analytics endpoint for Unity

AjimenezDCL:
Solves this: https://app.zenhub.com/workspaces/unity-6047bad476c3c0001942ee91/issues/decentraland/unity-renderer/299

Goes in pair with this `unity-renderer` PR: https://github.com/decentraland/unity-renderer/pull/458

Provide an endpoint for Unity to track generic events.

### DIFF
--- a/kernel/packages/unity-interface/BrowserInterface.ts
+++ b/kernel/packages/unity-interface/BrowserInterface.ts
@@ -89,11 +89,6 @@ type SystemInfoPayload = {
   systemMemorySize: number
 }
 
-type GenericAnalyticPayload = {
-  eventName: string
-  data: any
-}
-
 export class BrowserInterface {
   private lastBalanceOfMana: number = -1
 
@@ -179,16 +174,12 @@ export class BrowserInterface {
     trackEvent('system info report', data)
   }
 
-  public GenericAnalytic(payload: GenericAnalyticPayload) {
-    trackEvent(payload.eventName, payload.data)
-  }
-
   public PreloadFinished(data: { sceneId: string }) {
     // stub. there is no code about this in unity side yet
   }
 
-  public Track(data: { name: string; properties: { key: string; value: string }[] | null }) {
-    const properties: Record<string, string> = {}
+  public Track(data: { name: string; properties: { key: string; value: any }[] | null }) {
+    const properties: Record<string, any> = {}
     if (data.properties) {
       for (const property of data.properties) {
         properties[property.key] = property.value
--- a/kernel/packages/unity-interface/BrowserInterface.ts
+++ b/kernel/packages/unity-interface/BrowserInterface.ts
@@ -89,6 +89,11 @@ type SystemInfoPayload = {
   systemMemorySize: number
 }
 
+type GenericAnalyticPayload = {
+  eventName: string
+  data: any
+}
+
 export class BrowserInterface {
   private lastBalanceOfMana: number = -1
 
@@ -172,6 +177,10 @@ export class BrowserInterface {
 
   public SystemInfoReport(data: SystemInfoPayload) {
     trackEvent('system info report', data)
+  }
+
+  public GenericAnalytic(payload: GenericAnalyticPayload) {
+    trackEvent(payload.eventName, payload.data)
   }
 
   public PreloadFinished(data: { sceneId: string }) {


pbosio (approved):

menduz (changes requested):
In what cases the properties are  not strings? Please limit the use case to strings. We should able to handle all of the tracking cases with a `Map<string, string>` as properties

pravusjif (changes requested):
Apparently, somehow, the change triggers a bug in which the world never loads at specific coords for example at: 
```
https://play.decentraland.zone/feat/generics_metrics_from_unity/index.html?ENV=org&position=-11%2C-10&realm=baldr-amber&renderer=urn%3Adecentraland%3Aoff-chain%3Arenderer-artifacts%3Afeat%2Fgeneric_metrics_controller
```

Screen Shot 2021-05-18 at 13 52 43

While it doesn't happen with master
```
https://play.decentraland.zone/index.html?ENV=org&position=-11%2C-10&realm=baldr-amber&renderer=urn%3Adecentraland%3Aoff-chain%3Arenderer-artifacts%3Afeat%2Fgeneric_metrics_controller
```

menduz:
Pravus, you are using an incorrect URL

```
https://play.decentraland.zone/branch/feat/generics_metrics_from_unity/index.html?ENV=org&position=-11%2C-10&realm=baldr-amber&renderer=urn%3Adecentraland%3Aoff-chain%3Arenderer-artifacts%3Afeat%2Fgeneric_metrics_controller
```

menduz:
This change is not necessary tho, I suggest to close this PR and to reduce the scope of what can be tracked to keep a map<string,string> as it works today. Adding `any` is an antipattern that soften our typechecking, this is a clear opportunity where we must seek strong API contracts instead of loose definitions of "any" types that can break without notice.

BrianAmadori:
@menduz The `<string, string>` vs `<string, any>` discussion with the type-safe criteria is meaningless. I mean that we are not going to be type-safe with any of those alternatives when we take into account that the payload is just passed over to kernel to send it as JSON. Also, we aren't going to use that value after sending it, so its a one way road.

With this in mind, today we agreed to go for the `<string, string>` way but we are even considering just send a `string` with the dictionary as JSON. In any case we are not going to get type safeness with any implementation of the analytics message.

AjimenezDCL:
@BrianAmadori @menduz 
Just spoke with @sicarul and sending serialized jsons to Segment is a big no. So... If we want an endpoint to track arbitrary data for events I cannot think of any other way than using untyped data.

menduz:
This PR would send JSONs to segment, instead, If we don't merge this, we are enforcing only sending labels and values.

Sorry folks, I think there is so much missing context from unity side about monitoring and events to continue discussing this. We should close this PR and decide to adapt every use case to a plain map of labels and values. That is what scales better for any monitoring platform.

It would imply that the only things that we can attach to the event is a dictionary of strings

`SendEvent(eventName: string, labels: Record<string, string>)`

That is compatible with both segment, storing in plain databases, prometheus and it is also friendly with our alarms system.

The direct effect in the code would be to change the usage from

```ts
// BAD
SendEvent("click_in_quests_button", {
  quest: {
    questId: "some-id"
  },
  position: {
    x: 1,
    y: 2
  }
})

// GOOD
SendEvent("click_in_quests_button", {
  quest_id: "some-id",
  position_x: "1",
  position_y: "2",
})
```

sicarul:
From my PoV, both cases laid out by @menduz above work fine with Segment, however it is simpler to debug and to follow up when using the second format, because the names in the properties will 100% match the names in the DWH (e.g. if you have location:{map:{x:1}} segment will create something like location_map_x), the only disadvantages i see for using the string/string map are:

1) No native types since all are strings (numbers and booleans will have to be strings)
2) Will need to serialize arrays

I think both of these are reasonable compromises in favor of simplicity of the analytics subsystem, therefore i suggest we follow his suggestion, from my "data receiver" PoV.

Thanks!

AjimenezDCL:
Sure thing guys. 

We were trying to keep the `objects` with the assumption that they would be beneficial for the data analysis. All based on the most recent analytic implementation of the `SystemInfo` report where the payload sent to segment is such as:

```
type SystemInfoPayload = 
{
  graphicsDeviceName: string
  graphicsDeviceVersion: string
  graphicsMemorySize: number
  processorType: string
  processorCount: number
  systemMemorySize: number
}
```

If you rather have everything flattened out, that's completly fine.

Thank you all for the discussion, I will close the PR and proceed as agreed.

PR #2370, opened 2021-05-18
fix: map profiles to urn everywhere

nchamo:
Now that unity supports URNs, we need to map all legacy ids to URNs. The problem is that we forgot to map ids in profiles in two different scenarios:
* Profiles that were stored in local storage.
* Profiles received by comms

We are now fixing that.


pablitar (approved):

pablitar:
As Tusam would say, "puede fallar".

PR #2372, opened 2021-05-18
feat: Added API to expose explorer configuration to scenes

pablitar:
Scenes currently don't really have much information about the client that is executing them. The idea is to start building an API that the scenes can use to query that information, and to start adding data to it as it is needed.

The proposed contract is that "configuration" values can be present or not, so scenes have to handle the case in which the configuration is missing or doesn't exists anymore.

For the moment, this will be used to have smart defaults for the quests library, keeping in mind that we should synchronize what the scenes see and what the UI (which is handled by kernel + unity) shows.

### DIFF
--- a/kernel/packages/decentraland-ecs/types/@decentraland/EnvironmentAPI/index.d.ts
+++ b/kernel/packages/decentraland-ecs/types/@decentraland/EnvironmentAPI/index.d.ts
@@ -8,7 +8,7 @@ declare module '@decentraland/EnvironmentAPI' {
 
   export type ExplorerData = {
     clientUri: string
-    configurations: Record<string, any>
+    configurations: Record<string, string | number | boolean>
   }
 
   /**
@@ -25,5 +25,5 @@ declare module '@decentraland/EnvironmentAPI' {
   /**
    * Returns explorer configuration and environment information
    */
-  export function getExplorerData(): Promise<ExplorerData>
+  export function getExplorerConfiguration(): Promise<ExplorerData>
 }
--- a/kernel/packages/decentraland-ecs/types/@decentraland/EnvironmentAPI/index.d.ts
+++ b/kernel/packages/decentraland-ecs/types/@decentraland/EnvironmentAPI/index.d.ts
@@ -6,7 +6,7 @@ declare module '@decentraland/EnvironmentAPI' {
     displayName: string
   }
 
-  export type ExplorerData = {
+  export type ExplorerConfiguration = {
     clientUri: string
     configurations: Record<string, string | number | boolean>
   }
@@ -25,5 +25,5 @@ declare module '@decentraland/EnvironmentAPI' {
   /**
    * Returns explorer configuration and environment information
    */
-  export function getExplorerConfiguration(): Promise<ExplorerData>
+  export function getExplorerConfiguration(): Promise<ExplorerConfiguration>
 }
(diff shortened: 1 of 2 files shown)


kernel/packages/decentraland-ecs/types/@decentraland/EnvironmentAPI/index.d.ts
```diff
@@ -5,6 +5,13 @@ declare module '@decentraland/EnvironmentAPI' {
     serverName: string
     displayName: string
   }
+
+  export type ExplorerData = {
+    clientUrl: string
+    buildNumber: number
+    configurations: Record<string, any>
```
menduz: Could it more specific with the Record<string, any> to limit the use case to Record<string, string>? ala process.env

that way we ensure type safety for the present usages and a safeguard for the future

kernel/packages/shared/apis/EnvironmentAPI.ts
```diff
@@ -53,4 +60,18 @@ export class EnvironmentAPI extends ExposableAPI {
       displayName: `${serverName}-${layer}`
     }
   }
+
+  /**
+   * Returns explorer configuration and environment information
+   */
+  @exposeMethod
+  async getExplorerData(): Promise<ExplorerData> {
```
menduz: "Data" is a very broad name, just like "Id" what do you think about `getExplorerConfiguration` or something even more specific?

menduz (changes requested):

kernel/packages/decentraland-ecs/types/@decentraland/EnvironmentAPI/index.d.ts
pablitar: I'll make a compromise: Strings, number or booleans.

kernel/packages/shared/apis/EnvironmentAPI.ts
pablitar: Changed.

kernel/packages/decentraland-ecs/types/@decentraland/EnvironmentAPI/index.d.ts
```diff
@@ -5,6 +5,12 @@ declare module '@decentraland/EnvironmentAPI' {
     serverName: string
     displayName: string
   }
+
+  export type ExplorerData = {
```
nchamo:
```suggestion
  export type ExplorerConfiguration = {
```

For consistency

kernel/packages/decentraland-ecs/types/@decentraland/EnvironmentAPI/index.d.ts
```diff
@@ -14,4 +20,10 @@ declare module '@decentraland/EnvironmentAPI' {
    * Returns whether the scene is running in preview mode or not
    */
   export function isPreviewMode(): Promise<boolean>
+
+
+  /**
+   * Returns explorer configuration and environment information
+   */
+  export function getExplorerConfiguration(): Promise<ExplorerData>
```
nchamo:
```suggestion
  export function getExplorerConfiguration(): Promise< ExplorerConfiguration>
```

nchamo (approved):

kernel/packages/decentraland-ecs/types/@decentraland/EnvironmentAPI/index.d.ts
menduz: good enough

menduz (approved):
Please consider Chamo's suggestions before merge

PR #2374, opened 2021-05-18
feat: performance meter tool

pravusjif:
client side: https://github.com/decentraland/unity-renderer/pull/460

Testing:
1. access: https://play.decentraland.zone/branch/feat/PerformanceMeterTool/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:feat/PerformanceMeterTool&ENV=org
2. open the browser console and execute `clientDebug.RunPerformanceMeterTool(10000);` to run the performance meter tool for 10 seconds
3. wait for the complete data report to be printed in the console


kernel/packages/unity-interface/UnityInterface.ts
```diff
@@ -539,4 +531,25 @@ export class UnityInterface {
...
+    this.unityInterface = unityInterface
+  }
+
+  public DumpScenesLoadInfo() {
+    this.unityInterface.SendMessageToUnity('Main', 'DumpScenesLoadInfo')
+  }
+
+  public DumpRendererLockersInfo() {
+    this.unityInterface.SendMessageToUnity('Main', 'DumpRendererLockersInfo')
+  }
```
menduz, line 547: Why using a class instead of exported top level funtions using the `let unityInterface`?

pravusjif: I think exposing a `clientDebug` object is easier for discovering the exposed debug functions in there just by remembering that object's name, typing it and `.` and seeing the console methods suggestions, instead of having to remember evey debug function we have for the client (right now that would be `DumpScenesLoadInfo()`, `DumpRendererLockersInfo()` and now `RunPerformanceMeterTool(durationInMilliseconds: number)`)

menduz (approved):

PR #2375, opened 2021-05-18
fix: stuck in loading while in signup flow

pbosio:
New users are often stuck in an infinite loading screen while creating their profile.
This is because activate the renderer was only called when scenes at player's position finished loading.
In the signup flow scenario, the scene finished loading but kernel deactivate the renderer and didn't activate it again.
So if scene finished loading before user finished the signup then the renderer is never activated.

Steps to reproduce the issue:
1- open browser, incognito, cache clean and with metamask disabled (cause it's easier to test as a guest)
2- go to play.decentraland.zone
3- mute the tab (you can thank me later)
4- enter as guest
5- you now see avatar editor, don't click anything
6- open a new tab/window on your browser and go to your favorite crypto-news website and catch up with the latest news (you need to wait like a minute or more to make sure that scenes finish loading on background)
7- go back to avatar editor, press done
8- fill up your avatar name ("TheRealPravus" is recommended but not mandatory)
9- accept TOS
10- if everything went wrong then you are stuck in an eternal loading screen

### DIFF
--- a/kernel/packages/entryPoints/website.ts
+++ b/kernel/packages/entryPoints/website.ts
@@ -42,6 +42,7 @@ import { kernelConfigForRenderer } from '../unity-interface/kernelConfigForRende
 import Html from 'shared/Html'
 import { filterInvalidNameCharacters, isBadWord } from 'shared/profiles/utils/names'
 import { startRealmsReportToRenderer } from 'unity-interface/realmsForRenderer'
+import { isWaitingTutorial } from 'shared/loading/selectors'
 
 const logger = createLogger('website.ts: ')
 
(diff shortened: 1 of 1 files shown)


kernel/packages/entryPoints/website.ts
```diff
@@ -141,23 +142,29 @@ namespace webApp {
...
         }
 
         EnsureProfile(identity.address)
           .then((profile) => {
             i.ConfigureEmailPrompt(profile.tutorialStep)
             i.ConfigureTutorial(profile.tutorialStep, tutorialConfig)
             i.ConfigureHUDElement(HUDElementID.GRAPHIC_CARD_WARNING, { active: true, visible: true })
+
+            if (isWaitingTutorial(globalThis.globalStore.getState())) {
+              teleportObservable.notifyObservers(worldToGrid(lastPlayerPosition))
```
nchamo: Is this how the renderer is activated?

pbosio: Yes, if scene is already loaded it will activate the renderer there, if scene is not loaded then it will wait until loading finishes

nchamo: Could we move this (and the same call below) to a new function named `activateRenderer` or something like that? Because I'm sure I'll forget what this does in a few months

nchamo (approved):
Thanks for the comment!

PR #2382, opened 2021-05-20
fix race conditions

menduz:
# What? 
...

# Why? 
...


website/src/components/auth/EthLogin.tsx
```diff
@@ -38,14 +37,9 @@ export const EthLogin: React.FC<EthLoginProps> = (props) => {
...
             </button>
-            {!hasWallet && (
-              <button
-                className="eth-login-guest-button"
-                onClick={handlePlayAsGuest}
-              >
-                Enter as Guest
-              </button>
-            )}
+            <button className="eth-login-guest-button" onClick={handlePlayAsGuest}>
```
nchamo, line 40: Just want to understand this change. We are now allowing users to login as guests, even when they have a wallet, right?

nchamo (approved):

website/src/components/auth/EthLogin.tsx
menduz: Yep

menduz: I needed it to test, and there is a conversation in the chat for it with @RominaMarchetti. Romi, do you signoff?

RominaMarchetti (approved):

PR #2386, opened 2021-05-24
fix: restore accidentally deleted static files

menduz:
# What? 
@pablitar is there anything else to check?


pablitar (approved):
@menduz I hope not.

PR #2394, opened 2021-05-27
fix: enormous transform values physics breakdown

pravusjif:
**WHY**
Right now the physics engine sync suffers a breakdown when loading the Genesis Plaza from far away (when the clouds start moving), this is due to receiving super large values for some entities position and scale, the details and evidence can be seen at the [corresponding issue](https://app.zenhub.com/workspaces/unity-6047bad476c3c0001942ee91/issues/decentraland/unity-renderer/494).

**WHAT**
Added pos and scale capping to avoid enormous values breaking the physics engine in runtime

**TESTING**
Repro steps:
1. Enter the world at `-11, -11`
2. Wait until the genesys plaza ends  its loading and then a bit more for the clouds system to start running
3. You will notice that some objects in the road at `-11, -10` are not colliding with the character (this is what the PR fixes)
4. Move to the Genesis Plaza NFTShapes gallery and you will see the picture frames are not there. (not fixed by this PR since that is consecuence of the scene code crashing)

* Production (bugged): https://play.decentraland.org/?position=-11,-11&LOS=4&realm=fenrir-amber

* This branch (fixed): https://play.decentraland.zone/branch/fix/EnormousTransformValuesPhysicsBreakdown/index.html?ENV=org&position=-11,-11&LOS=4&realm=fenrir-amber

### DIFF
--- a/kernel/packages/scene-system/sdk/Utils.ts
+++ b/kernel/packages/scene-system/sdk/Utils.ts
@@ -1,6 +1,8 @@
 import { CLASS_ID, Transform } from "decentraland-ecs/src"
 import { PB_Transform, PB_Vector3, PB_Quaternion } from "../../shared/proto/engineinterface_pb"
+import { Vector3 } from "decentraland-ecs"
 
+const VECTOR3_MEMBER_CAP = 1000000 // Value measured when genesis plaza glitch triggered a physics engine breakdown
 const pbTransform: PB_Transform = new PB_Transform()
 const pbPosition: PB_Vector3 = new PB_Vector3()
 const pbRotation: PB_Quaternion = new PB_Quaternion()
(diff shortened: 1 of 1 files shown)


menduz:
I suggest you to move the responsibility to the message handler from unity, since it is closer to the bug. 

In this case, there is nothing broke to fix here. Also in unity there exist the possibility to add a test for this to prevent accidental deletion.

pravusjif:
> I suggest you to move the responsibility to the message handler from unity, since it is closer to the bug.
> 
> In this case, there is nothing broke to fix here. Also in unity there exist the possibility to add a test for this to prevent accidental deletion.

We decided to tackle this capping kernel-side to avoid performance impacts in the client, I already have the capping + test in [this client-side PR](https://github.com/decentraland/unity-renderer/pull/506) I just closed.

I just ran some performance tests in the client (with `System.Diagnostics.Stopwatch`) to have more info on the performance impact of capping the transform value vectors in there, here are a couple of snapshots measuring the time spent on updating transform values (pos + rot + scale) using roughly the same sampling time and the same loaded coords (the kernel patch was not running in these tests, only with and without the patch in the client):

no-client-patch

client-patch

I think the impact is almost inexistent, @BrianAmadori do you think we could have the capping in the client?

BrianAmadori:
I think that the issues with enormous values in the transform are pretty global and every renderer implemented would have them. At the moment I'd prefer to save the milliseconds and leave this as it is, we don't have a single ms to spare here.

menduz (approved):

PR #2401, opened 2021-06-01
feat: signup screen unity

AjimenezDCL:
https://app.zenhub.com/workspaces/unity-6047bad476c3c0001942ee91/issues/decentraland/unity-renderer/408

Remove signup from website and enable an endpoint for Unity to report the passport information.


menduz:
Do we have an ADR explaining the change and how it will behave afterwards? I think we should align the implementation before moving forward, only to make sure we are paying attention to all corner cases and to evaluate PRs with a framework. 

Because **I really love the idea of removing 1000 lines of code**, but I want to make sure that my approval makes sense in the whole context of the change.

AjimenezDCL:
> Do we have an ADR explaining the change and how it will behave afterwards? I think we should align the implementation before moving forward, only to make sure we are paying attention to all corner cases and to evaluate PRs with a framework.
> 
> Because **I really love the idea of removing 1000 lines of code**, but I want to make sure that my approval makes sense in the whole context of the change.

It's still a WIP (I should have set the PR as draft, my bad), This kernel is working locally and I want to check the behaviour in the build.

But sure thing, I will make an ADR explaining the flow.

kuruk-mm (approved):

menduz (approved):

PR #2402, opened 2021-06-01
Feat: Use the new catalyst health check to verify a server is up before connecting

jmoguilevsky:
# What? 
Filters out non healthy catalyst peers, to prevent connecting to an unhealthy one

# Why? 
...

### DIFF
--- a/kernel/packages/shared/dao/index.ts
+++ b/kernel/packages/shared/dao/index.ts
@@ -22,6 +22,8 @@ import { realmToString } from './utils/realmToString'
 import { PIN_CATALYST } from 'config'
 const qs: any = require('query-string')
 
+const DEFAULT_TIMEOUT = 5000
+
 const v = 50
 const score = ({ usersCount, maxUsers = 50 }: Layer) => {
   if (usersCount === 0) {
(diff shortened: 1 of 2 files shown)


menduz:
Are we waiting for all catalysts to whether succeed or fail? What happens if we have a gateway timeout of 50sec in _one_ of the catalysts? Will it hang for 50sec?

jmoguilevsky:
do you think it would be better to do a promise race? 
The thing is that it needs to run the algorithm to choose a catalyst to compare between servers

pablitar:
> do you think it would be better to do a promise race?
> The thing is that it needs to run the algorithm to choose a catalyst to compare between servers

We had this issue before when selecting which catalyst to connect to. We ended up using a timeout: https://github.com/decentraland/explorer/blob/master/kernel/packages/shared/dao/index.ts#L42

I'm not sure if we need to do that horrible stuff to have a timeout, but I'd see if we could have a timeout for this requests too. If a catalyst don't respond, at most you'd be blocked 5 or 10 seconds.

jmoguilevsky:
Just added an abort controller to timeout at 5s, what do you think about that?
I think 5s is a good measure for a server that we don't want to connect to, either way, the score algorithm would've scored it very low

pablitar (approved):
LGTM!

PR #2405, opened 2021-06-02
feat: [BiW Editor] New publishing flow for set name, description and screenshot of the scene (Kernel Side)

sandrade-dcl:
## What this PR includes?

- Kernel side for this implementation: https://github.com/decentraland/unity-renderer/pull/537


kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
```diff
@@ -91,9 +91,26 @@ export class SceneStateStorageController extends ExposableAPI implements ISceneS
   }
 
   @exposeMethod
-  async publishSceneState(sceneId: string, sceneState: SerializedSceneState): Promise<DeploymentResult> {
+  async publishSceneState(sceneId: string, sceneName: string, sceneDescription: string, sceneState: SerializedSceneState): Promise<DeploymentResult> {
```
menduz: I want to clarify that changes in signatures like this one are only enabled before we make the public release, after that moment, we should create a new function accepting new arguments, to not break backwards compatibility

menduz: to clarify just in case, we are good now, but watch out for the future changes

sandrade-dcl: Yeah, you're right. At the moment, as we are in a "pre-launch" stage, we are still designing certains parts. Thanks for the clarification!

menduz (approved):

nchamo (approved):

PR #2406, opened 2021-06-02
fix: error reporting overhaul

BrianAmadori:
# What? 

This PR implements most of the requirements of the [error reporting RFC](https://docs.google.com/document/d/1eCHxBpvnFc-fvEXf540y7unUr0mXhQ6TmidkkUnTn2E/edit#heading=h.o921fa8bd20p)

In a nutshell:
- An additional `context` field is added for easy filtering
- Manual throws (saga errors, unity's error handler, etc) are uploaded to Rollbar as critical errors.
- Unity errors include a BIG payload with detailed client state as JSON (this weights around 25KB or more). For more details on the payload contents check the RFC.
- Refactored all our throw instances to have some form of `context` data.
- Hardware capability errors (like no webgl, etc) aren't reported to rollbar, but they are tracked on analytics.
- `comms_error_x` analytics event changed to `error_comms_retry` and the number of retries is included in the error payload.

# How to test

Go to console and jot the following:
``` 
clientDebug.TestErrorReport(`your critical error of choice`, `renderer#errorHandler`)
```

Now, you'll see the error appear in Rollbar dashboard.

To visualize the full unity payload, you have to use the "Explore as RQL query" on the ocurrences tab:

image

Then change the visualization mode and voila:

image

Technical details:

- now, `ReportFatalError` will only report the error to Rollbar and accept Error objects as parameters, so rollbar has the stack and other useful data
- now, `BringDownClientAndShowError` is exported and is in charge of send the tracking event for the popup, as the tracking data can be useful. `ReportFatalError` doesn't send tracking events anymore. 
- For compatibility errors, we just call `BringDownClientAndShowError`. We don't report compatibility errors to Rollbar.
- `ReportFatalError` now report the errors as critical.
- introduced the `ErrorContext` namespace so the errors can be classified in Rollbar's dashboard.
- add placeholder to inject renderer's and comms' payloads.
- removed the reporting for renderer messages queue as this is not useful at all
- modified `generic_error` payload to match Rollbar payload


kernel/packages/shared/loading/ReportFatalError.ts
```diff
@@ -54,47 +61,83 @@ export function bringDownClientAndShowError(event: ExecutionLifecycleEvent) {
...
+    ReportFatalError(error, context)
   }
-  trackEvent('generic_error', {
-    message: event,
-    errorInfo
+}
+
+export function ReportFatalError(error: Error, context: ErrorContextTypes, payload: any = null) {
+  const finalPayload = GetErrorPayload(context, payload)
+  trackEvent('error_generic', {
```
menduz: Is the event `error_generic` fatal enough?

kernel/packages/shared/loading/ReportFatalError.ts
```diff
@@ -54,47 +61,83 @@ export function bringDownClientAndShowError(event: ExecutionLifecycleEvent) {
...
+    message: error.message,
```
sicarul, line 116: Is it also possible to send the type? (error.type) in addition to the message?

kernel/packages/shared/loading/ReportFatalError.ts
```diff
@@ -54,47 +61,83 @@ export function bringDownClientAndShowError(event: ExecutionLifecycleEvent) {
...
+    payload: finalPayload
```
sicarul: The payload can be ~25kb sometimes? Might be appropiate to shorten it when sending it to segment? (e.g. only send first 1000 characters of the JSON.stringify to Segment, and the complete payload to rollbar?) The main objective for segment data is to drive analytics, and even though we can handle it, we might want to connect a particular service that doesn't handle these big payloads (Segment can send data to many destinations)

kernel/packages/shared/loading/ReportFatalError.ts
sicarul: That's a good question, do we have any "non-fatal" errors going through this function? If not, might be worth renaming to "error_fatal" or "fatal_error".

BrianAmadori: ofc not, didn't want to change any tracking event name to not break any existing queries. But I can change it to `error_critical`. In fact I can also change the method name to `ReportCriticalError` to encourage ubiquity. wdyt?
cc: @sicarul

sicarul: We do not have any good queries for errors, so don't worry about backwards compatibility. The existing queries are broken AFAIK anyway.

menduz (approved):

kernel/packages/shared/loading/ReportFatalError.ts
menduz: I'll defer to @sicarul to decide, you already have my approve

pablitar (approved):
LGTM! Great work!

kernel/packages/shared/loading/ReportFatalError.ts
sicarul: @BrianAmadori did you see this comment?

kernel/packages/shared/loading/ReportFatalError.ts
sicarul: @BrianAmadori did you also see this comment?

sicarul (changes requested):

kernel/packages/shared/loading/ReportFatalError.ts
sicarul: Talked offline - Will add context (Type doesn't exist anymore)

kernel/packages/shared/loading/ReportFatalError.ts
sicarul: Talked offline - Let's remove the payload, part of it only is not useful.

sicarul (approved):
LGTM

PR #487, opened 2020-02-06
feat: /goto next (with placeholder parcels)

eordano:

PR #2407, opened 2021-06-03
refactor: ENABLE_WEB3=false enters always as a guest

kuruk-mm:
# What?
Related to #487

When we splicitly set `ENABLE_WEB3=false` in the browser, we disable the sign in process, and we enter directly as a guest.

# Why?
To have a way to skip Web3 interaction for tests and desktop (temporarly)


kuruk-mm:
Wrong approach

PR #2412, opened 2021-06-08
feat: support unpublished wearables

nchamo:
We are now adding support to test wearables that have not been published yet. These wearables live on the builder server, and have not been deployed on a content server yet.

Collections of this sort have a uuid instead of a urn, so the idea is to use the `WITH_COLLECTIONS` parameters with the uuid.

For example:
```play.decentraland.zone?WITH_COLLECTIONS=9d2e8db1-fbed-473d-8280-108293d4163a```


pablitar (approved):
LGTM! Great work!!

PR #2426, opened 2021-06-15
feat: biw custom assets pack

D4rWiNSS:
This PR gives the renderer a way to ask for the builder in world catalog headers to complete the catalog request


kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -552,6 +559,20 @@ export class BrowserInterface {
...
 
+  public RequestBIWCatalogHeader() {
+    const store: Store<RootState> = globalThis['globalStore']
+    const identity = getCurrentIdentity(store.getState())
+    if (!identity) {
+      let emptyHeader: Record<string, string> = {}
+      unityInterface.SendBuilderCatalogHeaders(emptyHeader)
+    } else {
+      const headers = BuilderServerAPIManager.authorize(identity, 'get', '/assetpacks')
+      headers['Content-Type'] = 'application/json'
```
menduz: What is this for? Content-Type header is not valid for GET requests

D4rWiNSS: It was a copy & paste fail, thanks for the heads up! Fixed!

sandrade-dcl (approved):
LGTM! But I have a doubt (only for curiossity): why are we making 2 calls (1 requesting the headers and 1 requesting the custom assets using these headers)? Couldn't we do both things mounting only 1 request directly in kernel? I guess you want to take advantage of our request system (with retries) and that's why you do it in this way, right?

D4rWiNSS:
> LGTM! But I have a doubt (only for curiossity): why are we making 2 calls (1 requesting the headers and 1 requesting the custom assets using these headers)? Couldn't we do both things mounting only 1 request directly in kernel? I guess you want to take advantage of our request system (with retries) and that's why you do it in this way, right?

There are a couple of factor for this decision, first of all, we need the private key of the wallet that is stored in kernel and we can't pass the private key since that can compromise the security so we are generating the headers in kernel and passing them to unity so unity can make the proper call.

In order to delay the initialization of builder in world to when the user is using it, I have 2 messages, 1 for request the headers, and 1 to send the headers to the renderer.

Since we have a more robust web request system in renderer and we are moving all the logic to the renderer, I though this is the more robust way to implement it

menduz:
Next time we had to send headers to renderer, which imo is a good approach, I would think about a generic solution to don't repeat ourselves and more importantly not repeat code

PR #2427, opened 2021-06-15
feat: message to authenticate from the renderer as a guest or using w…

kuruk-mm:
…allet connect

Main issue: decentraland/unity-renderer#487
Depends on decentraland/unity-renderer#611

# What? 
Add a way to authenticate from the renderer as a guest or using wallet_connect

# Why? 
Desktop client needs to decide itself when will authenticate


kernel/packages/shared/session/index.ts
```diff
@@ -119,6 +120,18 @@ export async function userAuthentified(): Promise<void> {
   })
 }
 
+export function authenticateWhenItsReady(providerType: ProviderType | null) {
```
pablitar, line 123: Maybe you want to check if we are already at the sign_in state?
```typescript
export function authenticateWhenItsReady(providerType: ProviderType | null) {
  const store: Store<RootState> = globalThis.globalStore
  if (store.getState().session.loginStage === LoginStage.SIGN_IN) {
    globalThis.globalStore.dispatch(authenticate(providerType))
  }
  const unsubscribe = store.subscribe(() => {
    const loginStage = store.getState().session.loginStage
    if (loginStage === LoginStage.SIGN_IN) {
      unsubscribe()
      globalThis.globalStore.dispatch(authenticate(providerType))
    }
  })
}
```

I'm not sure about the possible flows, so this may not be necessary. But I'm asking since this may be a gotcha.

kuruk-mm: Yeah, you're right, and I would like to ignore the message if we already signed in.

I pushed the changes, what do you think now?

And the flow it's to authenticate from unity-renderer to give the ability to desktop to sign in as a guest (and in the future with wallet connect).

I forgot to link the issue to this PR:
decentraland/unity-renderer#487

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -278,9 +289,19 @@ export class BrowserInterface {
...
+      providerType = ProviderType.WALLET_CONNECT
+    }
+
+    authenticateWhenItsReady(providerType)
+  }
+
   public SendPassport(passport: {name: string, email: string}) {
     globalThis.globalStore.dispatch(signupForm(passport.name, passport.email))
     globalThis.globalStore.dispatch(signUp())
+
```
kuruk-mm:
```suggestion
```

kernel/packages/shared/session/index.ts
pablitar: Looks good! 

Just as an extra mile, you may want to put the accesses to the store in a selector, since that is considered best practice for accessing state in redux and similar: https://github.com/decentraland/explorer/blob/master/kernel/packages/shared/session/selectors.ts

But feel free to ignore this since we are not respecting that convention in a bunch of places already hahaha.

pablitar (approved):
LGTM!

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -286,6 +292,15 @@ export class BrowserInterface {
...
   }
 
+  public SendAuthentication(data: { rendererAuthenticationType: RendererAuthenticationType }) {
+    let providerType: ProviderType | null = null
+    if (data.rendererAuthenticationType === RendererAuthenticationType.WALLET_CONNECT) {
+      providerType = ProviderType.WALLET_CONNECT
+    }
+
+    authenticateWhenItsReady(providerType)
+  }
```
pbosio: why not sending a string as payload instead of trying to match enums of integers and maintaining one enum on kernel and other on the renderer? it would be probably easier to extend in the future
`ProviderType.WALLET_CONNECT -> "wallet_connect"`

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -287,6 +288,16 @@ export class BrowserInterface {
...
 
+  public SendAuthentication(data: { rendererAuthenticationType: string }) {
+    let providerType: ProviderType | null = null
+
+    if (data.rendererAuthenticationType === 'wallet_connect') {
+      providerType = ProviderType.WALLET_CONNECT
+    }
+
+    authenticateWhenItsReady(providerType)
+  }
```
pbosio: I think you could do something like this
but we should ask a typescript lord 
@pablitar ? 🙂

```suggestion
  public SendAuthentication(data: { rendererAuthenticationType: string }) {
    const providerType: ProviderType | null = Object.values(ProviderType).includes(data.rendererAuthenticationType as ProviderType) ? data.rendererAuthenticationType as ProviderType : null

    authenticateWhenItsReady(providerType)
  }
```

pbosio (approved):

kernel/packages/unity-interface/BrowserInterface.ts
pablitar: Looks good! Sorry I was late hahaha.

PR #2428, opened 2021-06-15
feat: guest button

2fd:
# What? 

Add `Play as a Guest` button inside the login modal

Screen Shot 2021-06-18 at 15 57 32


RominaMarchetti:
@2fd the Guest button should say "Play as Guest", not "enter" and here is the figma link with the design for that popup adaptation https://www.figma.com/file/feevEf3dpoEMDLLcm5deSs/SignInV1?node-id=1291%3A54

pravusjif (changes requested):
Also the old "Play as Guest" button that appears below the "Play" button should be removed right?

https://user-images.githubusercontent.com/1031741/122266667-68a2d580-ceb0-11eb-9e31-1a2f65948106.mov

2fd:
That wasn't specified in the request, @RominaMarchetti WDYT?

menduz:
Mmm I think it would be a good idea to have both and measure how much people click both buttons before deciding to remove one

pravusjif (approved):

RominaMarchetti (approved):

PR #2429, opened 2021-06-15
fix: wait for identity to be loaded when going to the builder

nchamo:
Unity is asking for the owned wearables during start up, so it might happen that the identity hasn't been resolved yet. So we are now waiting for it to be resolved before going to the builder API


pablitar (approved):
Best two lines of code ever :heart:

PR #2430, opened 2021-06-15
fix: BiW create builder project from scene's state definition

pbosio:
When opening a scene to edit using builder in world the project could have been deleted from builder dapp.
In that scenario we load the scene with it state definition
We generate a new builder project and we upload it to the builder server


kernel/packages/shared/apis/SceneStateStorageController/utils.ts
```diff
@@ -0,0 +1,33 @@
+import { CLASS_ID } from 'decentraland-ecs/src'
+
+/**
+ * We are converting from numeric ids to a more human readable format. It might make sense to change this in the future,
+ * but until this feature is stable enough, it's better to store it in a way that it is easy to debug.
+ */
```
menduz, line 6: :so-nice:

nchamo (approved):

PR #2432, opened 2021-06-17
feat: [BiW Editor] BiW should update Project metadata and thumbnail (Kernel side)

sandrade-dcl:
## What this PR includes?

- Kernel side for this implementation: https://github.com/decentraland/unity-renderer/pull/628

### DIFF
--- a/kernel/packages/scene-system/stateful.scene.system.ts
+++ b/kernel/packages/scene-system/stateful.scene.system.ts
@@ -80,6 +80,16 @@ class StatefulWebWorkerScene extends Script {
           .catch((error) => this.error(`Failed to save the scene's manifest`, error))
       }
     })
+
+    // Listen to save scene info requests
+    this.eventSubscriber.on('stateEvent', ({ data }) => {
+      const { type, payload } = data
+      if (type === 'SaveSceneInfo') {
+        this.sceneStateStorage
+          .saveSceneInfo(serializeSceneState(this.sceneDefinition), payload.title, payload.description, payload.screenshot)
+          .catch((error) => this.error(`Failed to save the scene's info`, error))
+      }
+    })
   }
 
   private error(context: string, error: Error) {
--- a/kernel/packages/scene-system/stateful.scene.system.ts
+++ b/kernel/packages/scene-system/stateful.scene.system.ts
@@ -81,12 +81,12 @@ class StatefulWebWorkerScene extends Script {
       }
     })
 
-    // Listen to save scene info requests
+    // Listen to save project info requests
     this.eventSubscriber.on('stateEvent', ({ data }) => {
       const { type, payload } = data
-      if (type === 'SaveSceneInfo') {
+      if (type === 'SaveProjectInfo') {
         this.sceneStateStorage
-          .saveSceneInfo(serializeSceneState(this.sceneDefinition), payload.title, payload.description, payload.screenshot)
+          .saveProjectInfo(serializeSceneState(this.sceneDefinition), payload.title, payload.description, payload.screenshot)
           .catch((error) => this.error(`Failed to save the scene's info`, error))
       }
     })
--- a/kernel/packages/shared/apis/SceneStateStorageController/ISceneStateStorageController.ts
+++ b/kernel/packages/shared/apis/SceneStateStorageController/ISceneStateStorageController.ts
@@ -8,5 +8,5 @@ export interface ISceneStateStorageController {
   getProjectManifestByCoordinates(land: string): Promise<SerializedSceneState | undefined>
   createProjectWithCoords(coordinates: string): Promise<boolean>
   createProjectFromStateDefinition(): Promise<SerializedSceneState | undefined>
-  saveSceneInfo(sceneState: SerializedSceneState, sceneName: string, sceneDescription: string, sceneScreenshot: string): Promise<boolean>
+  saveProjectInfo(sceneState: SerializedSceneState, projectName: string, projectDescription: string, projectScreenshot: string): Promise<boolean>
 }
--- a/kernel/packages/shared/apis/SceneStateStorageController/ISceneStateStorageController.ts
+++ b/kernel/packages/shared/apis/SceneStateStorageController/ISceneStateStorageController.ts
@@ -8,4 +8,5 @@ export interface ISceneStateStorageController {
   getProjectManifestByCoordinates(land: string): Promise<SerializedSceneState | undefined>
   createProjectWithCoords(coordinates: string): Promise<boolean>
   createProjectFromStateDefinition(): Promise<SerializedSceneState | undefined>
+  saveSceneInfo(sceneState: SerializedSceneState, sceneName: string, sceneDescription: string, sceneScreenshot: string): Promise<boolean>
 }
--- a/kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
+++ b/kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
@@ -49,7 +49,7 @@ export class SceneStateStorageController extends ExposableAPI implements ISceneS
 
     if (!manifest) return undefined
 
-    globalThis.unityInterface.SendBuilderProjectInfo(manifest.project.title, manifest.project.description)
+    globalThis.unityInterface.SendBuilderProjectInfo(manifest.project.title, manifest.project.description, false)
     this.builderManifest = manifest
     const definition = fromBuildertoStateDefinitionFormat(manifest.scene)
     return serializeSceneState(definition)
@@ -59,7 +59,7 @@ export class SceneStateStorageController extends ExposableAPI implements ISceneS
   async getProjectManifestByCoordinates(land: string): Promise<SerializedSceneState | undefined> {
     const newProject = await this.builderApiManager.getBuilderManifestFromLandCoordinates(land, this.getIdentity())
     if (newProject) {
-      globalThis.unityInterface.SendBuilderProjectInfo(newProject.project.title, newProject.project.description)
+      globalThis.unityInterface.SendBuilderProjectInfo(newProject.project.title, newProject.project.description, false)
       this.builderManifest = newProject
       const translatedManifest = fromBuildertoStateDefinitionFormat(this.builderManifest.scene)
       return serializeSceneState(translatedManifest)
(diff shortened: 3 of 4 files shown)


kernel/packages/shared/apis/SceneStateStorageController/ISceneStateStorageController.ts
```diff
@@ -8,4 +8,5 @@ export interface ISceneStateStorageController {
   getProjectManifestByCoordinates(land: string): Promise<SerializedSceneState | undefined>
   createProjectWithCoords(coordinates: string): Promise<boolean>
   createProjectFromStateDefinition(): Promise<SerializedSceneState | undefined>
+  saveSceneInfo(sceneState: SerializedSceneState, sceneName: string, sceneDescription: string, sceneScreenshot: string): Promise<boolean>
```
pbosio: we already have the `saveSceneState` method. why not using that instead of creating a new one?

kernel/packages/scene-system/stateful.scene.system.ts
```diff
@@ -80,6 +80,16 @@ class StatefulWebWorkerScene extends Script {
...
+
+    // Listen to save scene info requests
+    this.eventSubscriber.on('stateEvent', ({ data }) => {
+      const { type, payload } = data
+      if (type === 'SaveSceneInfo') {
+        this.sceneStateStorage
+          .saveSceneInfo(serializeSceneState(this.sceneDefinition), payload.title, payload.description, payload.screenshot)
+          .catch((error) => this.error(`Failed to save the scene's info`, error))
+      }
+    })
```
pbosio: we already have the `SaveSceneState` event type. why not using that instead of creating a new one?

kernel/packages/shared/apis/SceneStateStorageController/ISceneStateStorageController.ts
sandrade-dcl: `saveSceneState` and `saveProjectInfo` do different things and are called in different situations:
- `saveSceneState`: updates the entities states of the project manifest and is called each time a change is done in the BiW editor.
- `saveProjectInfo`: updates the basic info of the project (name, description and screenshot) and is called only in 3 situation: when we create a new project, when we exit BiW editor and there have been changes or when we publish.

kernel/packages/scene-system/stateful.scene.system.ts
sandrade-dcl: https://github.com/decentraland/explorer/pull/2432#discussion_r653809226

kernel/packages/shared/apis/SceneStateStorageController/ISceneStateStorageController.ts
pbosio: as far as I can see. we will be trying to save the project twice to the builder-server in the case of publishing.
and we might find some unexpected results depending in which event comes first cause one will overwrite the other.
maybe we can send those payloads you need as optional and inject them to the project manifest if they exist instead of creating new events and methods? I don't know. What's Chamo's opinion?

kernel/packages/scene-system/stateful.scene.system.ts
D4rWiNSS: Maybe we can change it to "SaveProjectInfo"  since we are saving the info of the project? This way it will be less confusing

kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
```diff
@@ -100,6 +100,26 @@ export class SceneStateStorageController extends ExposableAPI implements ISceneS
...
+    sceneState: SerializedSceneState,
+    sceneName: string,
+    sceneDescription: string,
+    sceneScreenshot: string
+  ): Promise<boolean> {
+    let result: boolean
+    try {
+      const thumbnailBlob: Blob = base64ToBlob(sceneScreenshot, 'image/png')
+      await this.updateProjectDetails(sceneState, sceneName, sceneDescription, thumbnailBlob)
+      result = true
```
D4rWiNSS: I remember that we had an issue with messages from kernel to unity with base64 because they were a too long messages, we should have it into account

kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
```diff
@@ -350,6 +353,28 @@ export class SceneStateStorageController extends ExposableAPI implements ISceneS
...
+
+    // Update the project info
+    builderManifest.project.title = sceneName
+    builderManifest.project.description = sceneDescription
+
+    // Update the manifest
+    await this.builderApiManager.updateProjectManifest(builderManifest, this.getIdentity())
+
+    // Update the thumbnail
+    await this.builderApiManager.updateProjectThumbnail(builderManifest.project.id, thumbnailBlob, this.getIdentity())
```
D4rWiNSS, line 376: This should be done in the same line, we should update the manifest and only call to the API 1 time. The builder API will save the project and manifest together, so we are making 2 calls for the same purpose

kernel/packages/shared/apis/SceneStateStorageController/ISceneStateStorageController.ts
D4rWiNSS: If you refer to 

`await this.builderApiManager.updateProjectManifest(builderManifest, this.getIdentity())`  

and 

`await this.builderApiManager.updateProjectThumbnail(builderManifest.project.id, thumbnailBlob, this.getIdentity())<!--EndFragment-->`

This should be done in the same call or at least use the same project and manifest since as stated by pbosio, it can cause some info lose in certain scenarios

D4rWiNSS (changes requested):

kernel/packages/shared/apis/SceneStateStorageController/ISceneStateStorageController.ts
sandrade-dcl: Personally I don't like the idea of reusing a same method with different info injected and, depending on if some of this info is null or not, follow a flow or other inside the method. For me each method should have only one responsibility. Are we having different methods using a common function (in this case `updateProjectDetails()`) along its flow? Yes, but I don't think it is a problem.

About what you say of the `publishSceneState`, it is not saving the project twice, it is saving the project in builder on the one hand and deploying it in server on the other. It is something that I already had implemented before this PR and was approved by Nico.

Anyway let ask @nchamo and see what he think. Thanks for the feedback Pato!

sandrade-dcl: @Suduck why it could cause info lose? We are doing the calls in this way:
```
await this.builderApiManager.updateProjectManifest(builderManifest, this.getIdentity())
await this.builderApiManager.updateProjectThumbnail(builderManifest.project.id, thumbnailBlob, this.getIdentity())
```

The second call is wating for the first one is finished.

kernel/packages/scene-system/stateful.scene.system.ts
sandrade-dcl: Yes, it seems a more descriptive name 👍

kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
sandrade-dcl: I don't know what issue you refeer, but I have not had any problem at the moment updating the thumbnail through a base64. In fact it was the method that Nico suggested me to save images.

kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
sandrade-dcl: It is done in this way because it is what the Builder API currently offers (`projects/{projectId}/manifest` and `projects/{projectId}/media`). Do you think it is worth changing the API to unify these 2 calls?

kernel/packages/scene-system/stateful.scene.system.ts
D4rWiNSS: Rememeber to change it in the unity part too!

kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
D4rWiNSS: My bad, I thought that it was updating the manifest on one hand and the project on the other. NVM!

kernel/packages/scene-system/stateful.scene.system.ts
sandrade-dcl: `SaveProjectInfo` renaming (Kernel side) done in https://github.com/decentraland/explorer/pull/2432/commits/7b545b17ce9ad8bcadccad730134e2036f869555.

kernel/packages/shared/apis/SceneStateStorageController/ISceneStateStorageController.ts
D4rWiNSS: Yep, my bad I thought it was different

D4rWiNSS (approved):

kernel/packages/scene-system/stateful.scene.system.ts
sandrade-dcl: `SaveProjectInfo` renaming (Unity side) done in: https://github.com/decentraland/unity-renderer/pull/628/commits/858e0ff990730807d69ab4d294cac6a666adf3c6

kernel/packages/shared/apis/SceneStateStorageController/ISceneStateStorageController.ts
nchamo: I want to check if I understood correctly. 
Both `saveSceneState` and the new `saveProjectInfo` update the project on to Builder, right?

nchamo: I want to check if I understood correctly. 
Both `saveSceneState` and the new `saveProjectInfo` update the project on to Builder, right?

nchamo: I want to check if I understood correctly. 

Both `saveSceneState` and the new `saveProjectInfo` update the project on to Builder, right? And when they do it, they both send the scene's state to the Builder, right?

Now, my question is: is there an action (for example publishing) that would call both `saveSceneState` and the new `saveProjectInfo` together? Or are they always called at different times?

sandrade-dcl: @nchamo exactly, both `saveSceneState` and the new `saveProjectInfo` update the project on to Builder. Currently there not exist any action that is calling both functions together.

The publish action update the project on to Builder + deploy the scene.

nchamo: Then I see no chance of a race condition that would cause problems. So it's a matter of how you think would be the best way to code it. I'll leave that for you to discuss and decide, since this whole code is something you've been working on for the last months 😊 

I'll take a look at the rest of the PR to check if I see something weird

sandrade-dcl: Ok, thanks Nico!

nchamo (approved):

PR #2433, opened 2021-06-18
report all kinds of errors

menduz:
# What? 
...

# Why? 
...

### DIFF
--- a/kernel/packages/shared/analytics.ts
+++ b/kernel/packages/shared/analytics.ts
@@ -81,7 +81,7 @@ export function identifyEmail(email: string, userId?: string) {
   }
 }
 
-export function trackEvent(eventName: string, eventData: any) {
+export function trackEvent(eventName: string, eventData: Record<string, any>) {
   const data = { ...eventData, time: new Date().toISOString(), sessionId, version: (window as any)['VERSION'] }
 
   if (DEBUG_ANALYTICS) {
@@ -91,6 +91,7 @@ export function trackEvent(eventName: string, eventData: any) {
   if (!window.analytics) {
     return
   }
+
   window.analytics.track(eventName, data, { integrations: { Klaviyo: false } })
 }
 
@@ -118,13 +119,16 @@ function hookObservables() {
   const gridPosition = Vector2.Zero()
 
   positionObservable.add(({ position }) => {
-
     // Update seconds variable and check if new parcel
     if (performance.now() - lastTime > 1000) {
       worldToGrid(position, gridPosition)
       const currentPosition = `${gridPosition.x | 0},${gridPosition.y | 0}`
       if (previousPosition !== currentPosition) {
-        trackEvent('Move to Parcel', { newParcel: currentPosition, oldParcel: previousPosition, exactPosition: { x: position.x, y: position.y, z: position.z } })
+        trackEvent('Move to Parcel', {
+          newParcel: currentPosition,
+          oldParcel: previousPosition,
+          exactPosition: { x: position.x, y: position.y, z: position.z }
+        })
         previousPosition = currentPosition
       }
       lastTime = performance.now()
--- a/kernel/packages/shared/loading/ReportFatalError.ts
+++ b/kernel/packages/shared/loading/ReportFatalError.ts
@@ -110,7 +110,7 @@ export async function ReportFatalErrorWithUnityPayloadAsync(error: Error, contex
 }
 
 export function ReportFatalError(error: Error, context: ErrorContextTypes, payload: Record<string, any> = {}) {
-  const eventData = {
+  const rollbarEventData = {
     // attach every field from the payload to the event
     ...(payload || {}),
     // set the context property
@@ -121,8 +121,14 @@ export function ReportFatalError(error: Error, context: ErrorContextTypes, paylo
     stack: getStack(error).slice(0, 10000)
   }
 
-  trackEvent('error_fatal', eventData)
-  ReportRollbarError(error, eventData)
+  // segment requires less information than rollbar
+  trackEvent('error_fatal', {
+    context: rollbarEventData.context,
+    message: rollbarEventData.message,
+    stack: rollbarEventData.stack,
+    saga_stack: payload['sagaStack']
+  })
+  ReportRollbarError(error, rollbarEventData)
 }
 
 function getStack(error?: any) {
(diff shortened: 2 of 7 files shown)


kernel/packages/unity-interface/loader.ts
```diff
@@ -58,7 +61,12 @@ async function injectRenderer(baseUrl: string, version: string): Promise<LoadRen
         productVersion: '0.1'
       }
 
-      return originalCreateUnityInstance(canvas, config, onProgress)
+      return originalCreateUnityInstance(canvas, config, function (...args) {
+        if (args[0] == 1.0) {
+          trackEvent('unity_started', { version, loading_time: performance.now() - startLoadingTime })
```
BrianAmadori: This event actually tracks the interval for downloading the Unity artifacts, so this should be called `unity_downloading_end` or similar.

Also, we agreed with @sicarul to not calculate the time ourselves but have begin/end events. So we have `unity_downloading_start` as well.

Then we need another event to track the engine initialization time. This is the time between this event and the call to `StartDecentraland()` callback.

kernel/packages/unity-interface/UnityScene.ts
```diff
@@ -24,6 +25,15 @@ export class UnityScene<T> implements ParcelSceneAPI {
 
   constructor(public data: EnvironmentData<T>) {
     this.logger = DEBUG_SCENE_LOG === true ? createLogger(getParcelSceneID(this) + ': ') : createDummyLogger()
+
+    const startLoadingTime = performance.now()
+
+    this.eventDispatcher.once('sceneStart', () => {
```
BrianAmadori, line 31: We should divide this tracking event in two, `scene_start` and `scene_end`. Also let's use snake_case to be more consistent. cc: @sicarul

kernel/packages/shared/loading/ReportFatalError.ts
```diff
@@ -109,27 +109,34 @@ export async function ReportFatalErrorWithUnityPayloadAsync(error: Error, contex
...
+export function ReportFatalError(error: Error, context: ErrorContextTypes, payload: Record<string, any> = {}) {
   const finalPayload = GetErrorPayload(context, payload)
-  trackEvent('error_fatal', {
-    context: context,
-    message: error.message,
-    stack: error.stack?.slice(0, 10000)
-  })
+  const eventData = {
+    // attach every field from the payload to the event
+    ...(finalPayload || {}),
```
BrianAmadori: `finalPayload` may contain the Unity payload that's too big for sending it to segment. We should parametrize this method to send only the payloads on the saga stack case.

BrianAmadori (changes requested):

kernel/packages/unity-interface/loader.ts
sicarul: Yes, please let's split into two events

kernel/packages/unity-interface/loader.ts
```diff
@@ -58,7 +61,12 @@ async function injectRenderer(baseUrl: string, version: string): Promise<LoadRen
         productVersion: '0.1'
       }
 
-      return originalCreateUnityInstance(canvas, config, onProgress)
+      return originalCreateUnityInstance(canvas, config, function (...args) {
+        if (args[0] == 1.0) {
```
sicarul: What's this args thing? How do you know if it'll be 1.0 and what does it mean?

kernel/packages/shared/analytics.ts
```diff
@@ -81,7 +81,7 @@ export function identifyEmail(email: string, userId?: string) {
   }
 }
 
-export function trackEvent(eventName: string, eventData: any) {
+export function trackEvent(eventName: string, eventData: Record<string, any>) {
```
sicarul, line 84: What does this change do?

kernel/packages/unity-interface/UnityScene.ts
menduz: Line 32 has it `scene_start_event` the `sceneStart` is the event sent by unity to the SDK

kernel/packages/shared/loading/ReportFatalError.ts
sicarul: What does this finalpayload thing mean? What is a typical payload? Segment will create a column in the table for each property sent, so i'd like to know what we're going to send here.

sicarul (commented):
@menduz  I see you added `scene_start_event` and `unity_started`, however i don't see them in the deploy being tracked:

image

Am i watching in the wrong place?

sicarul (changes requested):
Please check comments

kernel/packages/unity-interface/loader.ts
menduz: I think even though we _can_ do that, it is easier to have the control of that in our end whenever possible. To avoid race conditions problems like we have right now about changing the sequence order, which leads to break the ETL and metrics

kernel/packages/shared/analytics.ts
menduz: Stronger types, to prevent sending nulls or `typeof eventData == "number"` which leads to unhandled exceptions

kernel/packages/unity-interface/loader.ts
menduz: Addressed

kernel/packages/shared/loading/ReportFatalError.ts
menduz: Addressed in last commit. please check, it is all the `eventData` keys plus all the `payload` keys

kernel/packages/unity-interface/loader.ts
menduz: Pushed both approaches

kernel/packages/shared/loading/ReportFatalError.ts
```diff
@@ -110,18 +110,19 @@ export async function ReportFatalErrorWithUnityPayloadAsync(error: Error, contex
...
+    context,
     // this is on purpose, if error is not an actual Error, it has no message, so we use the ''+error to call a
     // toString, we do that because it may be also null. and (null).toString() is invalid, but ''+null works perfectly
     message: error.message || '' + error,
     stack: getStack(error).slice(0, 10000)
   }
 
   trackEvent('error_fatal', eventData)
-  ReportRollbarError(error, finalPayload)
+  ReportRollbarError(error, eventData)
```
BrianAmadori: We shouldn't send the `eventData` like this because the stack and message fields are redundant, rollbar already brings it from the error object.

BrianAmadori (approved):
A bit concerned about the hardcoded expectation of the `saga_stack` field. 

Aside of that, LGTM codewise.

sicarul (approved):
LGTM

PR #2434, opened 2021-06-19
fix: builder in world. translate rotation applied by builder's deploy

pbosio:
while deploying a scene using builder you can choose to rotate it.
we need to translate those transformations from builder to scene's state definition when editing in builder in world and from scene's state definition to builder when updating the project to the builder-server.

### DIFF
--- a/kernel/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
+++ b/kernel/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
@@ -18,7 +18,8 @@ import { getDefaultTLD } from 'config'
 import { defaultLogger } from '../../logger'
 import { getParcelSceneLimits } from 'atomicHelpers/landHelpers'
 import { CLASS_ID } from 'decentraland-ecs/src'
-import { toHumanReadableType, fromHumanReadableType } from './utils'
+import { toHumanReadableType, fromHumanReadableType, getLayoutFromParcels } from './utils'
+import { SceneSourcePlacement } from 'shared/types'
 
 export const BASE_DOWNLOAD_URL = 'https://builder-api.decentraland.org/v1/storage/contents'
 const BASE_BUILDER_SERVER_URL_ROPSTEN = 'https://builder-api.decentraland.io/v1/'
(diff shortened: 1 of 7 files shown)


kernel/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
```diff
@@ -155,55 +156,92 @@ export class BuilderServerAPIManager {
     return builderManifest
   }
 
-  async createManifestFromSerializedState(
+  async builderManifesFromSerializedState(
```
pablitar:
```suggestion
  async builderManifestFromSerializedState(
```

kernel/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
```diff
@@ -155,55 +156,92 @@ export class BuilderServerAPIManager {
...
     builderProjectId: string,
     baseParcel: string,
     parcels: string[],
     title: string | undefined,
     description: string | undefined,
+    ethAddress: string,
     scene: SerializedSceneState,
-    identity: ExplorerIdentity
-  ): Promise<BuilderManifest | undefined> {
-    try {
```
pablitar, line 168: Why are we removing this try/catch? Won't this make the client crash if something fails in builder in world or something like that?

kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
```diff
@@ -35,13 +35,16 @@ import { ExplorerIdentity } from 'shared/session/types'
...
+import { getLayoutFromParcels } from './utils'
+import { SceneTransformTranslator } from './SceneTransformTranslator'
 
 declare const globalThis: any
 
 export class SceneStateStorageController extends ExposableAPI implements ISceneStateStorageController {
   private readonly builderApiManager = new BuilderServerAPIManager()
   private parcelIdentity = this.options.getAPIInstance(ParcelIdentity)
   private builderManifest!: BuilderManifest
+  private transformTranslator!: SceneTransformTranslator
```
pablitar, line 47: Do we need an instance variable for this? Seems like when it is used, it is used like a local variable

kernel/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
pbosio: Actually the whole function was deleted. The diff here is misleading and looks like just a couple of things where removed :P
We can add a try/catch again just to be safe, but I don't think is necessary. Also the function `builderManifestFromSerializedState` is called inside a try/catch block

pablitar (approved):
LGTM! Great work :D

kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts
pbosio: Most of the time (when the stateful scene to edit is created) a instance of that variable is created.
But we use that reference while saving the scene for example
here: https://github.com/decentraland/explorer/blob/a5d50a174c4c60e3ae951fcb6583ec17108c082a/kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts#L88-L111
and here: https://github.com/decentraland/explorer/blob/a5d50a174c4c60e3ae951fcb6583ec17108c082a/kernel/packages/shared/apis/SceneStateStorageController/SceneStateStorageController.ts#L390-L417
any way, it seems that we could use them as local variable safely in those scenarios as well but I choose not to do that just in case :P

PR #2435, opened 2021-06-21
fix: remove email request prompt

pbosio:
#641 
remove email request prompt hud since we are already asking for user's email during sign-up


kernel/packages/shared/types.ts
```diff
@@ -440,14 +440,13 @@ export enum HUDElementID {
   CONTROLS_HUD = 18,
   EXPLORE_HUD = 19,
   HELP_AND_SUPPORT_HUD = 20,
-  EMAIL_PROMPT = 21,
```
menduz, line 454: Please leave this untouched, we should decouple the kernel change from the renderer change, to do so, we always should make enums append only

menduz (changes requested):
Looking good, please restore the enum and add a "// @deprecated" to the declaration

kernel/packages/shared/types.ts
pbosio: OKs! Good to know! I'll do that.
I think that probably it's not a very good idea to keep using this Enum then
Cause I still need to change the value for the following elements to match unity's

menduz: Is it already merged to master in unity renderer? I rather roll it back, it breaks compatibility with .zone

menduz (approved):
Awesome, thank you 🙌

PR #2438, opened 2021-06-22
Publish the build for decentraland website as an npm package

jmoguilevsky:
# What? 
...

# Why? 
...


.circleci/config.yml
```diff
@@ -223,7 +223,7 @@ jobs:
           paths:
             - website/build
 
-  publish-ecs:
+  publish-ecs-and-website:
```
menduz:
```suggestion
  publish-ecs:
```

website/scripts/publish.sh
```diff
@@ -2,6 +2,4 @@
 
 cp static/package.json build/
 cd build
-echo $(ls ../../kernel/scripts)
-npx tsc --build ../../kernel/scripts/tsconfig.json
-node ../../kernel/scripts/npmPublish.js
+npx -T ../scripts/npmPublish.ts
```
menduz: What does `npx -T` do?

jmoguilevsky: It does nothing on it's own, it was a typo, already sent the fix on another commit

jmoguilevsky: It was `npx ts-node -T ...` so that ts-node doesn't validate typescript errors and just transpiles the code

PR #2443, opened 2021-06-24
feat: add cache invalidation to asset bundles

### LINKED ISSUE #668: Texture issue that happens in zone but not org
Wonderzone appears with the red floor in the RC, the problem persisted even when reloading (in org it never happens).
Wonderzone runs quite slow and with stutters (comparing with ORG).
Comparing with ORG, the terms and services screen has a different font, and the scroll is still wrong (now it doesn't go all the way to the bottom).

BrianAmadori:
## What?

This PR makes the `baseUrl` used for asset bundles parametrizable via our external global configuration file. This is specially useful for cache invalidation. 

The implementation has a fallback that uses our default url when the global config file can't be retrieved, and it preserves the ability to be overriden via an URL param.

As a technical note, at first I tried to use the saga selectors, but saga selectors are cached to only be called again when the store changes. As the URL can be overriden by other factors that are external to the store--and may change on runtime--a simple getter from `config/index` is used instead.

Fixes https://github.com/decentraland/unity-renderer/issues/668


kernel/packages/config/index.ts
```diff
@@ -262,6 +268,20 @@ export function getNetworkFromTLD(tld: string = getTLD()): ETHEREUM_NETWORK | nu
   return null
 }
 
+export function getAssetBundlesBaseUrl(): string {
+  const state = globalThis.globalStore.getState()
+  const result =
+    ASSET_BUNDLES_DOMAIN || state.meta.config.explorer?.assetBundlesFetchUrl || getDefaultAssetBundlesBaseUrl()
+
+  defaultLogger.log(`getAssetBundlesBaseUrl = ${result}`)
```
menduz: is this log really needed? could it be under `if (DEBUG)`

pravusjif (approved):

D4rWiNSS (approved):

kernel/packages/config/index.ts
BrianAmadori: I forgot to remove it, thanks for noticing

PR #2446, opened 2021-06-25
refactor: update local base catalog

nchamo:
We are now updating the local base catalog. That way, we don't have to do the mapping anymore


pablitar (approved):
LGTM!

PR #2452, opened 2021-06-28
refactor: use content server directly

nchamo:
This PR does two things:
1. Deletes the `metaContentService`, which wasn't being used at all
2. The `fetchContentServer` will use the content server directly, instead of using the `/contentv2` lambdas. The use of this lambda is no longer necessary, since now all content (including wearables) can be found on the content server


pbosio (approved):

pablitar (approved):
:heart:

PR #2453, opened 2021-06-29
chore: preload unity while signing

AjimenezDCL:
Fix decentraland/explorer#632

Currently we have a spinner blocking any signin interaction until unity is ready. The metrics show users dropping-off without even seeing the signin buttons due to large loading times. 

This PR addresses the situation by allowing the signin process to be started while the Unity build is loading in the background. The spinner now will be visible if after finishing signin Unity is not ready.

### DIFF
--- a/kernel/packages/shared/ethereum/provider.ts
+++ b/kernel/packages/shared/ethereum/provider.ts
@@ -40,7 +40,7 @@ export function getEthConnector(): EthereumConnector {
 
 export async function requestProvider(type: ProviderType | null) {
   try {
-    const { provider } = await ethConnector.connect(type)
+    const { provider } = await getEthConnector().connect(type)
     requestManager.setProvider(provider)
     providerFuture.resolve({
       successful: !isGuest(),
@@ -56,7 +56,7 @@ export async function requestProvider(type: ProviderType | null) {
 }
 
 export function isGuest(): boolean {
-  return ethConnector.isGuest()
+  return getEthConnector().isGuest()
 }
 
 export function getProviderType() {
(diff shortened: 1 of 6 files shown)


AjimenezDCL:
After the CI passes:

- This branch can be previewed at https://play.decentraland.zone/branch/chore/preload_unity_while_signing/index.html?ENV=org
- The `decentraland-ecs` package can be tested in scenes by running
   ```bash
   npm install "https://play.decentraland.zone/branch/chore/preload_unity_while_signing/decentraland-ecs.tgz"
   ```

RominaMarchetti (commented):
@AjimenezDCL it seems that the spinner was completely removed, and it's a problem considering that after clicking any button visual feedback is not shown anymore, making it look like the request isn't happening. 
In the landing both buttons disappear after clicking any of them while in the wallets modal nothing happens. In every case, we must show the spinner after clicking the buttons the same way it is happening now in prod if possible :)

AjimenezDCL:
@RominaMarchetti 
> It seems that the spinner was completely removed, and it's a problem considering that after clicking any button visual feedback is not shown anymore, making it look like the request isn't happening.
> In the landing both buttons disappear after clicking any of them while in the wallets modal nothing happens. In every case, we must show the spinner after clicking the buttons the same way it is happening now in prod if possible :)

Ahhh I see what you mean, nice catch I overlook that spinner completly. Thanks!

RominaMarchetti (approved):

AjimenezDCL:
Just found a bug. If you don't complete the signup process (leave in the middle of the avatar editor, for example) the next time you try to log-in the world loades with no profile.

D4rWiNSS (approved):
Tested in different situations, waiting, trying to enter as fast as possible, in incognito as guess and normal sign-in, it seems to work in all situations, Good job!

PR #2459, opened 2021-07-02
feat: remote config for synapse

nchamo:
We are now allowing the external configuration of the `synapseUrl` we were using for friends & private chat


pablitar (approved):
LGTM!

PR #2460, opened 2021-07-02
fix: "Browser not supported" error screen missing

### LINKED ISSUE #708: Show warning notification for not recommended browsers
We have decided not to have a "black list" of prohibited browsers and only deny browsers due to technical limitations such as not supporting WebGL2. If an user enters with a browser that is compatible with WebGL2 but not recommended by us (any browser different to Firefox or any not chromium-based browser), we will show him a warning in the landing page indicating it but we won't cut his entrance to the world.

sandrade-dcl:
# What does this PR fix? 
## Fixes https://github.com/decentraland/unity-renderer/pull/708
This error screen is not being showed when trying to run Decentraland with Safari.

The words "Chrome" and "Firefox" are links:
- **Chrome**: https://www.google.com/chrome/?brand=BNSD&gclid=Cj0KCQjw5uWGBhCTARIsAL70sLJsOBpT9ZDBOEtFx49Ya-nc0iCR50klEV8iLNoOdkeVtVly8x9_gXUaAhSdEALw_wcB&gclsrc=aw.ds
- **Firefox**: https://www.mozilla.org/en-US/firefox/new/

### How to test it?
1. Open Safari browser.
2. Go to https://play.decentraland.zone/branch/fix/Browser-not-supported-error-screen-missing/index.html?ENV=org
3. Notice the error screen is showed.
4. Close Safari browser and open Chrome or Firefox.
5. Go to https://play.decentraland.zone/branch/fix/Browser-not-supported-error-screen-missing/index.html?ENV=org
6. Notice the world loads as usual.


kernel/packages/shared/comms/browser.ts
```diff
@@ -0,0 +1,14 @@
+export const compatibleBrowsers: string[] = [
+  "Chrome",
+  "Firefox"
+]
+
+export function isCompatibleBrowser() {
```
menduz, line 6: We can't do allow-lists of browsers, instead I suggest to check the capabilities of the browsers to ensure it will work. I use DCL in brave, and it works perfectly. It also works in latest safari. Edge also works perfectly in guest mode, so as do Opera.

menduz (changes requested):
We can't do allow-lists of browsers, instead I suggest to check the capabilities of the browsers to ensure it will work. I use DCL in brave, and it works perfectly. It also works in latest safari. Edge also works perfectly in guest mode, so as do Opera.

kernel/packages/shared/comms/browser.ts
sandrade-dcl: Yeah, I agree with you. Maybe we should to re-design the requirements for this issue and not to limit so much our supported browsers, because they are more than we thought it would. And furthermore, fixing the problem that we saw yesterday about the paralel loading of the website and the react app (current cause of the problems we are having with the old versions of Safari), I think we could support a lot of browsers.

I'm going to open the discussion with the team before to continue with the issue 👍 

Thanks for the feedback, Agus.

PR #2464, opened 2021-07-05
fix: Stop loading workflow after "Mobile not supported" error screen

### LINKED ISSUE #661: Stop loading workflow after `Mobile not supported` screen
Actionable from: https://github.com/decentraland/unity-renderer/issues/614#issuecomment-866285376

When we enter in Mobile, we see a screen like:

And even with that screen, the loading workflow continues in the background. This generates some confusing errors because of the low specs of Mobile (ex: system out of memory).

**How to reproduce:**
- Connect your phone with USB on your computer
- On your PC with Chrome, enter to `chrome://inspector`
- Open `https://play.decentraland.org/` on your mobile, and open the inspector on your computer for that tab

sandrade-dcl:
# What does this PR fix? 
## Fixes https://github.com/decentraland/unity-renderer/pull/661
When we enter in Mobile, we see a screen like this:

Even with that screen, the loading workflow continues in the background. This generates some confusing errors in our metrics because of the low specs of Mobile (ex: system out of memory).

### Solution
In the `initShared()` function, we have moved the `startSagas()` call after the `isMobile()` one in order to avoid to start all the flow if the user is running form a mobile.

### How to test it?
1. Connect your phone with USB on your computer
2. On your PC with Chrome, enter to `chrome://inspector`
3. Open `https://play.decentraland.zone/branch/fix/Stop-loading-workflow-after-Mobile-not-supported-screen/index.html?ENV=org` on your mobile, and open the inspector on your computer for that tab


D4rWiNSS (approved):
Not tested but code LGTM!

kuruk-mm (approved):
Works perfectly!

PR #2467, opened 2021-07-05
feat: Remove gtag.js (Google Tag Manager) and upgrade Segment.js

sicarul:
We are loading google tag manager (gtag.js) which serves no purpose (We already have google analytics data through segment.js) - I'm also upgrading segment.js to use their new analytics v2.0 feature that reduces the library's weight and improves the performance of tracking data from the browser:

https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#getting-started

Let me know if you have any concerns
(Testing segment.min.js goes from 57kb to 23kb when the server side flag is on, i only enabled it temporarily while testing on .zone to test)


sicarul:
I'm checking how to fix the failed test, vs code adds that semicolon...

menduz (approved):

menduz (approved):

2fd (approved):

PR #2471, opened 2021-07-06
fix: Deactivate errors reporting after client down

### LINKED ISSUE #662: Critical error "renderer#errorHandler" must shut down the app
Actionable from: https://github.com/decentraland/unity-renderer/issues/614#issuecomment-866285376

If we receive an error from the context `renderer#errorHandler`, we must shut down the app to avoid error propagation.

Ex: after a `system out of memory`, we don't stop the app, so we end with a `kernel#saga` error of `memory access out of bounds`.

sandrade-dcl:
# What does this PR fix? 
## Fixes https://github.com/decentraland/unity-renderer/pull/662
Each time the user receives a black screen error, the background flow continues and during this process we are reporting extra errors that are contaminating our metrics. After any of these errors, no more errors should be tracked neither reported to rollbar.

image

**SOLUTION:** Once the `BringDownClientAndShowError(...)` function is called, we stop track events and deactivate `Rollbar`.

### DIFF
--- a/kernel/packages/shared/loading/ReportFatalError.ts
+++ b/kernel/packages/shared/loading/ReportFatalError.ts
@@ -22,6 +22,7 @@ declare const globalThis: StoreContainer
 
 export let aborted = false
 
+// once this function is called, no more errors will be tracked neither reported to rollbar
 export function BringDownClientAndShowError(event: ExecutionLifecycleEvent) {
   if (aborted) {
     return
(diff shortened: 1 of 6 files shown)


kernel/packages/shared/loading/ReportFatalError.ts
```diff
@@ -22,6 +22,7 @@ declare const globalThis: StoreContainer
 
 export let aborted = false
```
BrianAmadori, line 23: We shouldn't export this variable if isn't used elsewhere.

kernel/packages/shared/loading/ReportFatalError.ts
```diff
@@ -43,22 +44,26 @@ export function BringDownClientAndShowError(event: ExecutionLifecycleEvent) {
...
+                : event === AVATAR_LOADING_ERROR
+                  ? 'avatarerror'
+                  : 'fatal'
 
   globalThis.globalStore && globalThis.globalStore.dispatch(fatalError(targetError))
   Html.showErrorModal(targetError)
   aborted = true
+
+  if (window.Rollbar) {
+    window.Rollbar.configure({ enabled: false })
```
BrianAmadori, line 65: If we disable `Rollbar` immediately after logging the error, you are certain that the error web request that got sent earlier in `ReportFatalError` is not aborted?

kernel/packages/shared/loading/ReportFatalError.ts
sandrade-dcl: It was already being used in these 2 places:
- `BrowserInterfaces.ts` -> `ControlEvent(...)` -> when it receives the `ActivateRenderingACK` event, the `aborted` flag is checked to notify observers or not.
- `dcl.ts` -> `startUnitySceneWorkers()` -> when the event `onPositionSettled` is triggered, the `aborted` flag is checked to teleport and activate rendering or not.

kernel/packages/shared/loading/ReportFatalError.ts
sandrade-dcl: Let me check it...

sandrade-dcl: Checked and confirmed: the previous `ReportFatalError` is being sent correctly before deactivate Rollbar. For check it, I have done this:
1. I have forced to sent a `window.Rollbar.critical(new Error('TEST ERROR BEFORE DEACTIVATE ROLLBAR!!'), {})` just BEFORE the deactivation and a `window.Rollbar.critical(new Error('TEST ERROR AFTER DEACTIVATE ROLLBAR!!'), {})` just AFTER.

image

2. I have forced to make a call `BringDownClientAndShowError(...)`.
3. Checking the Network section in the chrome developer tools, we can see that the request related to the first fatal error has been sent but not the second one.

image

BrianAmadori: Amazing!

kernel/packages/shared/loading/ReportFatalError.ts
BrianAmadori: Oops! Didn't find those uses for some reason. Thanks for the explanation!

BrianAmadori (approved):

kuruk-mm (approved):
LGTM!

PR #2477, opened 2021-07-09
fix: Connect to the next best realm if the Explore JumpIn fails (Kernel side)

### LINKED ISSUE #244: Teleporting to a coordinate with an unavailable realm should use the next best available realm
I used the “Jump In” button in our Explore HUD for the Wonder Mine scene (it’s always the most populated scene), and instead of being teleported I got the system message in the chat saying **“Couldn’t find realm loki-amber”**, I reckon the catalyst must’ve been down at the time and now it works fine and this issue is not happening right now.

But I think that teleporting to a specific coords and realm (as what happens using the jump in in the Explore HUD), **if the target realm isn't found or has any detected problem, we should still do the teleportation using the next-best realm**. 

So that if it happens again that the most populated scene is reported as using a realm that is temporarily down, we can still have the explore feature and its jump-on working and get into the scene instead of a broken feature.

sandrade-dcl:
## What does this PR fix?
This PR is the **Kernel side** for the **Fix https://github.com/decentraland/unity-renderer/issues/244**. It has to be merged with the unity side ([here](https://github.com/decentraland/unity-renderer/pull/781)).

**BEFORE** this change, when we jumped in to a scene from the "Explore & Play" window, the system tried to teleport us to a specific realm (choosed by the own Explore feature). When the joinning to this realm failed (because of the server down, a full realm, etc.), the feature was broken and we were not able to do the teleport.

**NOW**, after we try to jump in to a scene from the Explore, if the joinning to the specific realm fails, the system will automatically try to find the next better realm to go (we prioritize the list of available realms by number of users in each realm). Even if it is not be able to find any realm to go, it will finally use our current realm to do the teleport. In this way, the users will always be able to jump to the scene they wanted (although it has to be in another realm).

- **NOTE**: Each retry of finding a new realm will be informed to the user through the chat window.

## How to test the changes?
1. Go to: https://play.decentraland.zone/branch/fix/Connect-to-the-next-best-realm-if-the-jump-in-fails-kernel-side/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/Connect-to-the-next-best-realm-if-the-jump-in-fails-unity-side&ENV=org
2. Open the Explore window.
3. Select different scenes to jump in.

## Our Code Review Standards

https://github.com/decentraland/unity-renderer/blob/master/docs/code-review-standards.md


pravusjif (approved):
Kernel's side looks good to me, I also tried using the jump in apart from the Explore feature and saw no issues.

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -621,6 +631,41 @@ export class BrowserInterface {
...
+    if (nextRealmIndexToTry < jumpInData.nextMostPopulatedRealms.length) {
+      notifyStatusThroughChat(`Couldn't find realm ${realmString}. Trying to connect to the next more populated one...`)
+      await this.TryToJumpToTheNextMostPopulatedRealm(store, nextRealmIndexToTry, jumpInData)
+    } else {
+      notifyStatusThroughChat(`Couldn't find realm ${realmString}. You'll stay in your current realm.`)
+      this.GoTo(jumpInData.gridPosition)
+    }
+  }
+
+  private async TryToJumpToTheNextMostPopulatedRealm(store: Store<RootState>, realmIndex: number, jumpInData: JumpInPayload) {
```
BrianAmadori: I see that this method doesn't do anything that we couldn't do inside Unity.

We can't just expose the `JumpIn` method (with failure callbacks, etc.) and do all the handling from the renderer? Otherwise, this approach would violate our current direction of moving responsibilities to Unity.

BrianAmadori (changes requested):

kernel/packages/unity-interface/BrowserInterface.ts
menduz: ➕

sandrade-dcl: Hi guys, I would like to discuss this suggested approach before start any refactor on this, because I think that in this specific case it is not the best idea to move these responsibilities to Unity.

This `TryToJumpToTheNextMostPopulatedRealm` function is being called by the `JumpIn` function each time a connection with a realm fails and, if this happens, this function is doing a recursive call to the `JumpIn` function again to make a new try with the next best realm. So the current flow is:

### Current approach
1. **[Unity] -> [Kernel]**: Unity calls `JumpIn` function in Kernel with the corresponding realms information.
2. **[Kernel]** Kernel try to connect to the first candidate realm.
2.1. **[Kernel]** If the connection is successfully stablished, it does the jump to the corresponding scene/realm.
2.2. **[Kernel]** If the connection fails, it try with the next candidate realm (we go back to the point 2).

_(Only 1 connection from Unity to Kernel, and Kernel is who have the responsibility of manage the connections)_

With the approach of moving the connection errors handling to Unity (@BrianAmadori correct me if I'm wrong), the flow would be the next:

### Suggested approach
1. **[Unity] -> [Kernel]**: Unity calls `JumpIn` function in Kernel with the first candidate realm information.
2. **[Kernel]** Kernel try to connect to this realm.
2.1. **[Kernel]** If the connection is successfully stablished, it does the jump to this scene/realm.
2.2. **[Kernel] -> [Unity]** If the connection fails, Kernel will send a message to Unity to inform about the connection error.
3. **[Unity]**: Unity receives an error message from Kernel, then it calls again the `JumpIn` function in Kernel with the next candidate realm information (we go back to the point 2).

_(There could be a lot of connections between Unity and Kernel. Kernel would only have the responsibility of connect with a specific realm and Unity would have the responsibility of manage the "retries" system)_

For me, both approachs are valid, but I would personally prefer to avoid so much back-and-forth between Unity and Kernel, and give all the responsibility of "connection retries" to Kernel, which makes the connections with the catalysts.

Anyway, please let review both flows and let me know what do you think. If we finally decide to go with the second approach, we can do the refactor, no problem! :)

BrianAmadori: I think your argument makes sense from a mechanical complexity standpoint. The direction of moving the Kernel responsibilities over to Unity is reasoned from a maintenance costs standpoint.

As we argued many times, we should try to move most of the business logic maintenance responsibilities over to Unity, leaving to Kernel the low-level queries and features that can't be done from Unity. This is NOT done to reduce the roundtrips between Kernel and Unity, but to make maintenance easier (aka less costly).

If the approach of this PR is put on place, then the maintenance of the Explore feature will suddenly get harder because we increase the maintenance cost. I argue that the cost gets higher the more code we have on both components for the same feature. When you need to maintain features that have a Kernel and Unity side running at the same time, TypeScript understanding is needed, kernel<>renderer interoperability knowledge is needed, having to use make watch to test, double PRs, etc.

To rephrase and make the argument as clear as possible: It's true that from a mechanical standpoint we get a lot of roundtrips, but here I think we can trade off the roundtrips to cheapen the maintenance cost--we don't have any performance bottlenecks anyway. 

Also, this would enable the possibility of handling custom jumps in retry logic for other features entirely from Unity too.

sandrade-dcl: Ok! Let follow this approach then. Thank you Brian!

sandrade-dcl: @BrianAmadori I have already applied the agreed refactor: Now, the Kernel side remained as it was but only adding the 2 needed callbacks (`ConnectionToRealmSuccess` and `ConnectionToRealmFailed`) for informing Unity about the result of the JumpIn. And all the logic related to the realms retries system has been moved to the Unity side (https://github.com/decentraland/unity-renderer/pull/781).

Could you please take a look to both PRs? Thanks!

kernel/packages/unity-interface/BrowserInterface.ts
```diff
@@ -500,17 +501,21 @@ export class BrowserInterface {
...
           const cause = e === 'realm-full' ? ' The requested realm is full.' : ''
           notifyStatusThroughChat('Could not join realm.' + cause)
-
+          unityInterface.ConnectionToRealmFailed(data)
           defaultLogger.error('Error joining realm', e)
         }
       )
     } else {
-      notifyStatusThroughChat(`Couldn't find realm ${realmString}`)
+      notifyStatusThroughChat(`Couldn't find realm ${realmString}.`)
```
BrianAmadori, line 517: Note that in the case the retry system has to try multiple times, many `Couldn't find realm` chat messages may appear on the log. This may be undesired behaviour.

BrianAmadori (approved):
LGTM!

kernel/packages/unity-interface/BrowserInterface.ts
BrianAmadori: Done!

PR #2486, opened 2021-07-14
feat: use cdn-uploader for explorer

menduz:
# What? 
Establishes proper content-encoding (gzip/brotli) and mime-types to optimize static CDN deployments.

Screen Shot 2021-07-14 at 18 59 34

# Why? 
Because


menduz:
Merging because it is required for next release with wasm streaming

PR #2497, opened 2021-07-21
fix: Graphics setting "match 1080p" works as "unlimited" every time we start the explorer

### LINKED ISSUE #817: Graphics setting "match 1080p" works as "unlimited" every time we start the explorer
In macOS when we have the "match 1080p" graphics resolution setting saved in the settings panel, every time we start the explorer, it works as "unlimited", thus trying to reach max resolution which is 3072 x 1920 in retina display, greatly decreasing performance.

If I manually change the resolution setting to another setting and come back to 1080, it starts working fine again.

The canvas resolution can be checked in the dev tools, at the Elements tab, looking at the attributes of the `Canvas`, child of the `GameContainer` element:

sandrade-dcl:
# What does this PR change? 
Fix decentraland/unity-renderer/issues/817

In macOS when we have the "match 1080p" graphics resolution setting saved in the settings panel, every time we start the explorer, it works as "unlimited", thus trying to reach max resolution which is 3072 x 1920 in retina display, greatly decreasing performance.

If I manually change the resolution setting to another setting and come back to 1080, it starts working fine again.

The canvas resolution can be checked in the dev tools, at the Elements tab, looking at the attributes of the `Canvas`, child of the `GameContainer` element:

### Test link
https://play.decentraland.zone/branch/fix/Graphics-setting-match-1080p-works-as-unlimited-every-time-we-start-the-explorer/index.html?ENV=org

### DIFF
--- a/kernel/packages/unity-interface/UnityInterface.ts
+++ b/kernel/packages/unity-interface/UnityInterface.ts
@@ -68,10 +68,6 @@ export class UnityInterface {
       return
     }
 
-    if (this.currentHeight === height) {
-      return
-    }
-
     this.currentHeight = height
     resizeCanvas(height)
   }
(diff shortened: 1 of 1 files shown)


pravusjif (approved):
awsome!

PR #2498, opened 2021-07-21
Fix: Prevent avatar ghosts when leaving or changing islands

jmoguilevsky:
In this pr:
- Update the catalyst-peer dependency for the new interface of peerconfig
- Modify the listener for island change to update the peers in the context to prevent ghosts
- Add a listener for the on player leave to prevent ghosts when an avatar moves into another island (the last position recorded was being used for rendering a ghost of the avatar)

# Why? 
...

### DIFF
--- a/kernel/package-lock.json
+++ b/kernel/package-lock.json
@@ -404,9 +404,9 @@
       }
     },
     "@dcl/catalyst-peer": {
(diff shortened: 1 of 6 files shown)


pablitar (approved):
Great work :D

PR #2505, opened 2021-07-22
Fix: Allow for jumps ins and gotos in layerless realms

jmoguilevsky:
# What? 
In this PR:
- The jump In function is updated to use the `realmToString` function which returns the correct URL when the layer is not present
- The goto function uses the status for the island in the case there is no layer present

# Why? 
...


pablitar (approved):
LGTM!

PR #2514, opened 2021-07-28
[Do not merge] Add JWT using PoW

agusaldasoro:
Added code to fetch the cookie with the JWT after implementing Proof of Work every time the realm is changed (this includes start-up).

This PR doesn't include the changes to send the JWT cookie on every request to the Catalyst, as we need to deploy the Catalyst version that handles Cors with origin first.

This PR is blocked to merge until the release of the Catalyst that handles cors with dynamic origin is deployed.

### DIFF
--- a/kernel/packages/shared/dao/index.ts
+++ b/kernel/packages/shared/dao/index.ts
@@ -477,7 +477,6 @@ export function initializeUrlRealmObserver() {
 
 async function getJWT(realm: Realm): Promise<void> {
   const url: string = realm.domain + '/pow-auth/challenge'
-  console.log('Getting Challenge from ', url)
   const challengeResponse = await (await fetch(url, { credentials: 'include' })).json()
   const challenge = challengeResponse.challenge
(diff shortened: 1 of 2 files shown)


kernel/packages/shared/dao/index.ts
```diff
@@ -451,11 +452,12 @@ export function observeRealmChange(
   onRealmChange: (previousRealm: Realm | undefined, currentRealm: Realm) => any
 ) {
   let currentRealm: Realm | undefined = getRealm(store.getState())
-  store.subscribe(() => {
+  store.subscribe(async () => {
     const previousRealm = currentRealm
     currentRealm = getRealm(store.getState())
     if (currentRealm && !deepEqual(previousRealm, currentRealm)) {
       onRealmChange(previousRealm, currentRealm)
+      await getJWT(currentRealm)
```
pablitar: Would it make sense to add this as a subscriptor instead of directly adding it in this method?

PR #2529, opened 2021-08-03
fix: Show warning notification for not recommended browsers (only Kernel side)

### LINKED ISSUE #708: Show warning notification for not recommended browsers

sandrade-dcl:
# What does this PR fix?
Fix https://github.com/decentraland/unity-renderer/pull/708

We have decided not to have a "black list" of prohibited browsers and only deny browsers due to technical limitations such as not supporting WebGL2. If an user enters with a browser that is compatible with WebGL2 but not recommended by us (any browser different to Firefox or any not chromium-based browser), we will show him a warning in the landing page indicating it but we won't cut his entrance to the world.

### DIFF
--- a/kernel/packages/shared/comms/browser.ts
+++ b/kernel/packages/shared/comms/browser.ts
@@ -1,6 +1,21 @@
+export const recommendedBrowsers: string[] = [
+  "Chrome",
+  "Firefox"
+]
+
 export function isWebGLCompatible() {
   // Create canvas element. The canvas is not added to the document itself, so it is never displayed in the browser window.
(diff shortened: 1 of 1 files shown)


sandrade-dcl:
Blocked until the website/kernel refactor is finished and integrated.

PR #2531, opened 2021-08-04
docs: Update "How to test parcels and preview scenes" section

### LINKED ISSUE #551: The flow for loading preview scenes in Unity is broken
These instructions no longer work:
https://github.com/decentraland/explorer/blob/master/docs/how-to-test-parcels-and-preview-scenes.md#unity-editor-debugging-with-dcl-scene-in-preview-mode

The data is never sent to Unity, there are some errors in the browser console:

A discord user was asking about this and when I tried it locally before answering himI saw that the flow is broken.

When we fix this we could also update the URL and remove the `UNITY_ENABLED=true` param

sandrade-dcl:
# What? 
Fix decentraland/unity-renderer/issues/551

Update the "How to test parcels and preview scenes" section.

# Why? 
The processes described in the documentation were out of date.

### DIFF
--- a/docs/how-to-test-parcels-and-preview-scenes.md
+++ b/docs/how-to-test-parcels-and-preview-scenes.md
@@ -37,8 +37,8 @@ NOTE(optional): If you don't need to modify the test-scenes and just build them 
 
 1. Go to the scene folder.
 2. Compile the scene dependencies (run only one of the following options):
-  - *OPTION A*: For working with the @latest Kernel version, run `decentraland-ecs@latest`.
-  - *OPTION B*: For working with the @next Kernel version, run `decentraland-ecs@next`.
+  - **OPTION A**: For working with the @latest Kernel version, run `decentraland-ecs@latest`.
+  - **OPTION B**: For working with the @next Kernel version, run `decentraland-ecs@next`.
 3. Run `dcl start`. A new tab in your browser will be open with the scene in preview mode.
 
 #### Play a scene in Preview Mode using the local Kernel version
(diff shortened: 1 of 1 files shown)


docs/how-to-test-parcels-and-preview-scenes.md
```diff
@@ -20,28 +20,57 @@ Note that the test-scenes building may take a while (10 mins or more).
 
 4. To spawn at the specific coordinates of a test scene append the position query paramter:
 
-    http://localhost:3000/?DEBUG_MODE&position=-100,100
+    http://localhost:3000/?DEBUG_MODE&position=-1,27
```
D4rWiNSS, line 23: Is this change necessary?

D4rWiNSS (approved):
I don't know if this works, but english wise LGTM!

docs/how-to-test-parcels-and-preview-scenes.md
sandrade-dcl: I wanted to change these coords because the `-100,100` was giving me an error (probably because this test scene is broken) and I think it is more proper to give the users a valid example.

menduz: I'm curious about the broken scene. Does it throw an exception? Does it load at all?

BrianAmadori (approved):

docs/how-to-test-parcels-and-preview-scenes.md
sandrade-dcl: The problem seems to be specifically in the test scene `-102,102`. I attach the errors throwed in the browser console:

image
image

menduz: oh ok, that is not good. It seems like scenes can break the renderer. Could you track it as critical? People could not go to nearby locations CC @olavra . I left my approval for the docs, that should be handled in another pr

menduz (approved):

docs/how-to-test-parcels-and-preview-scenes.md
sandrade-dcl: Ok, we're going to create a task for this 👍

